PR #2, opened 2022-10-06
Fix sonarcloud test coverage reporting

Cruikshanks:
When we run our tests locally 94.24% is reported. However, SonarCloud is reporting 97.3%. We never understood this discrepancy in the SROC team. But when we came to WATER we were finally able to track it down to how things were configured in `sonar-project.properties`.

This change updates the file to bring in what we learnt and applied to our other **water-abstraction** projects.


Cruikshanks:
TBH it's still not exactly matching. But it is closer and at least consistent with the other projects 🤷

PR #4, opened 2022-10-10
Add views to system

StuAA78:
https://github.com/DEFRA/water-abstraction-team/issues/54

The first step to moving the service status page from the UI to our new system is to implement views. We do this using https://github.com/DEFRA/hapi-web-boilerplate as a basis to work from.

Once implemented, we add a simple `/hello-world` page, based on the example in `hapi-web-boilerplate`.

Building assets is done in the exact same way as the boilerplate app:
- `npm run build` calls a bash script `bin/build`. This script is intended to run whatever build stages are needed, but at present it only runs `npm run build:css`.
- `npm run build:css` uses `node-sass` to compile the sass stylesheet currently in `client/sass/application.scss` (which simply links to the `govuk-frontend` stylesheet) to standard css in `app/public/build`. Note that we update `.gitignore` so that this folder isn't checked in.
- A `postinstall` script is added to `package.json` so that `npm run build` is automatically run after an `npm ci` or `npm install`.

A few points to bear in mind for potential future changes:
- Is `client/sass/application.scss` where we want our sass stylesheet to live? 
- We've followed the model of the boilerplate app and serve a static file for `/robots.txt` (currently living in `app/public/static/robots.txt`). In the existing UI we have a `no-robots` plugin which serves this up on request without needing a static file to be stored. Do we want to do this instead?
- We probably want to double-check the `ErrorPagesPlugin` to ensure it correctly logs 500 errors; for now we've simply copied the error format in `hapi-web-boilerplate`, adapting it to use our own logger.
- Google Analytics hasn't yet been brought over from `hapi-web-boilerplate`.


Jozzey (approved):
👍🏼

PR #8, opened 2022-10-19
Refactoring summary list for each service

Jozzey:
https://github.com/DEFRA/water-abstraction-team/issues/54

We are going to create a summary list section for each app and we are also going to remove the tables from shared apps like virus scanner


Cruikshanks:
See comments in [last commit](https://github.com/DEFRA/water-abstraction-system/pull/8/commits/840b120035eba9349f0ab7600ed50419fd5fd4f9) for my thoughts on what to do next. But don't let that put anyone off opting for **Require changes**.

test/services/service_status.service.test.js
```diff
@@ -3,38 +3,176 @@
...
+// Normally we'd set this to `= require('../../app/services/service_status.service')`. But to control how
+// `child_process.exec()` behaves in the service, after it's been promisfied we have to use proxyquire.
+let ServiceStatusService // = require('../../app/services/service_status.service')
 
 describe('Service Status service', () => {
-  describe('importRows data', () => {
-    it('returns data in the expected format', async () => {
-      const { importRows } = await ServiceStatusService.go()
+  beforeEach(() => {
+    // These requests will remain unchanged throughout the tests. We do alter the ones to the AddressFace and the
```
StuAA78, line 20: Small typo
```suggestion
    // These requests will remain unchanged throughout the tests. We do alter the ones to the AddressFacade and the
```

test/services/service_status.service.test.js
```diff
@@ -3,38 +3,176 @@
...
+      Nock('http://localhost:8001').get('/health/info').reply(200, { version: '8.0.1', commit: '83d0e8c' })
+
+      // Unfortunately, this convoluted test setup is the only way we've managed to stub how the promisified version of
+      // `child-process.exec()` behaves in the class under test.
+      // We create an anonymous stub, which responds differently for the 1st and 2nd call. We rely on knowing the order
+      // they are called in the class under test for this to work.
+      // We then stub the util library's `promisify()` method and tell it to calll our anonymous stub when invoked.
+      // The bit that makes all this work is the fact we use Proxyquire to load our stubbed util instead of the real
+      // one when we load our class under test
+      const execStub = Sinon.stub().onFirstCall().resolves({ stdout: 'ClamAV 9.99.9/26685/Mon Oct 10 08:00:01 2022\n', stderror: null })
```
StuAA78, line 60: I haven't looked into this too deeply and maybe it's already been considered, but could it be possible to use `stub.withArgs()` in conjunction with a matcher like `sinon.match(string)` to stub specific calls?

app/services/service_status.service.js
```diff
@@ -37,70 +47,110 @@ class ServiceStatusService {
...
   }
 
-  static async _getServiceVersionsData () {
-    return [
+  static async _requestData (url) {
+    // As of v12, the got dependency no longer supports CJS modules. This causes us a problem as we are locked into
+    // using these for the time being. Some workarounds are provided here: https://github.com/sindresorhus/got/issues/1789
+    // We have gone the route of using await import('got'). We cannot do this at the top level as Node doesn't support
+    // top level in CJS so we do it here instead.
+    const { got } = await import('got')
```
StuAA78, line 87: At one point we were importing `got` within `go()` and passing it to each function so that it's only imported once. Happy to go along with this change just as long as we're sure there's no real performance impact from importing it every time `_requestData()` is used?

app/services/service_status.service.js
```diff
@@ -37,70 +47,110 @@ class ServiceStatusService {
...
+        retry: {
+          // We ensure that the only network errors Got retries are timeout errors
+          errorCodes: ['ETIMEDOUT'],
+          // We set statusCodes as an empty array to ensure that 4xx, 5xx etc. errors are not retried
+          statusCodes: []
+        }
+      })
+    } catch (error) {
+      const statusCode = error.response ? error.response.statusCode : 'N/A'
+      result.response = `ERROR: ${statusCode} - ${error.name} - ${error.message}`
```
StuAA78, line 104: Is `ERROR: N/A - errorName - errorMessage` what we want to see if there's no status code or would we prefer to make it conditional so it's just `ERROR: errorName - errorMessage`? (Or do we not care since the page is just us? :grin:)

config/services.config.js
```diff
@@ -0,0 +1,44 @@
...
+  },
+  permitRepository: {
+    url: process.env.PERMIT_REPOSITORY_URL
+  },
+  returns: {
+    url: process.env.RETURNS_URL
+  }
+}
+
+module.exports = config
```
StuAA78, line 44: Just a thought. I think we could directly export our config instead of assigning it to `config` and then exporting it, eg:

```js
module.exports = {
  addressFacade: {
    // etc.
  }
}
```

Is this something we would consider?

test/services/service_status.service.test.js
```diff
@@ -3,38 +3,176 @@
...
-        secondRow
+    Nock('http://localhost:8012').get('/health/info').reply(200, { version: '8.0.12', commit: '83d0e8c' })
+    Nock('http://localhost:8011').get('/health/info').reply(200, { version: '8.0.11', commit: 'a7030dc' })
+    Nock('http://localhost:8007').get('/health/info').reply(200, { version: '8.0.7', commit: 'a181fb1' })
+    Nock('http://localhost:8002').get('/health/info').reply(200, { version: '8.0.2', commit: '58bd0c1' })
+    Nock('http://localhost:8000').get('/health/info').reply(200, { version: '8.0.0', commit: 'f154e3f' })
+    Nock('http://localhost:8008').get('/health/info').reply(200, { version: '8.0.8', commit: 'f154e3f' })
+    Nock('http://localhost:8003').get('/health/info').reply(200, { version: '8.0.3', commit: '2ddff3c' })
+    Nock('http://localhost:8004').get('/health/info').reply(200, { version: '8.0.4', commit: '09d5261' })
+    Nock('http://localhost:8006').get('/health/info').reply(200, { version: '8.0.6', commit: 'b181625' })
```
StuAA78, line 39: Do we want to hardcode these urls or would we prefer to read them in from our config?

StuAA78 (changes requested):

app/services/service_status.service.js
Cruikshanks: It's still only imported once. Like `require()` any modules imported are cached. When another `import` call is made to the same module the ESM module loader will return the cached entry.

The same applies to something like the logger; we `require()` is in hundreds of files but Node is only loading it once. From then on it's returning a cached entry. It's also what makes trying to stub things in unit tests so much fun.

If it helps, we do the same thing in the [sroc-charging-module-api](https://github.com/DEFRA/sroc-charging-module-api/blob/main/app/services/charges/request_rules_service_charge.service.js#L123)

app/services/service_status.service.js
Cruikshanks: I went with 'it's fine as it's just us'! I didn't want to complicate the code any more than I already had 😬

config/services.config.js
Cruikshanks: 💯 I'd be up for that. But I'd go with doing this to all our config in one go. This means I'd merge this change and then add a new PR to cover refactoring all config to follow that pattern.

test/services/service_status.service.test.js
Cruikshanks: 🤷 We hardcode the values in our `.github/ci.yml` so the result would be the same. I guess there is an argument that you'd only have to declare the URL once. 🤷 

That said this is in a unit test. And IMHO I prefer to be very verbose and explicit about any test setup so it's clear what are the parameters of the test.

But honestly, this is not a hill I'm gonna die on so let me know your preference.

test/services/service_status.service.test.js
Cruikshanks: Not sure how `sinon.matcher()` applies. But `withArgs()` was a winner! Have submitted a change. Let me know what you think.

test/services/service_status.service.test.js
StuAA78: Happy to leave it as-is :grin:

StuAA78 (approved):
Happy with this PR as it is, the only potential change I'd carry forward to another PR is exporting config directly.

As an aside, we could probably do with researching the best way to resolve `package-lock.json` conflicts! Deleting and rebuilding it is fine for now while we're still getting it off the ground, my concern in future is that doing this could potentially lose any changes that Dependabot made for us -- to be fair it would probably just create new PRs for us to merge again but if we can prevent that happening in the first place then all the better! :grin:

[According to this post](https://tkdodo.eu/blog/solving-conflicts-in-package-lock-json) `npm install --package-lock-only` may fix this for us (it's not entirely clear but it seems to be suggesting that you run this while you're mid-merge?). So possibly worth a try next time we have this issue?

PR #9, opened 2022-10-21
Switch from CommonJS to ES6 modules

Jozzey:
Currently, when one file references another in the project we use the following

```javascript
const { InvoiceRebillingPresenter } = require('../presenters')
const InvoiceRebillingInitialiseService = require('./invoice_rebilling_initialise.service')
const InvoiceRebillingCopyService = require('./invoice_rebilling_copy.service')
```

This is known as *CommonJS* modules and up until the current LTS version of Node.js was the only type supported. In the meantime, the JavaScript language adopted *ES6* modules as its standard. With native support for them now in Node.js v14, we have the option to switch to them.

By switching we are keeping the code up to date with modern practices. We have also confirmed ES6 modules resolve an issue we have sometimes had to work around (circular dependencies). Finally, they are a bit more performant than CommonJS ones.

So, this change updates the whole project to switch over (our testing found that everything has to go at once; we can't just do a file at a time).

### DIFF
--- a/.labrc.js
+++ b/.labrc.js
@@ -1,5 +1,3 @@
-'use strict'
-
 module.exports = {
   verbose: true,
   coverage: true,
--- a/app/controllers/health/airbrake.controller.js
+++ b/app/controllers/health/airbrake.controller.js
@@ -1,5 +1,3 @@
-'use strict'
-
 class AirbrakeController {
   static async index (req, _h) {
     // First section tests connecting to Airbrake through a manual notification
@@ -18,4 +16,4 @@ class AirbrakeController {
   }
 }
 
-module.exports = AirbrakeController
+export default AirbrakeController
--- a/app/controllers/health/database.controller.js
+++ b/app/controllers/health/database.controller.js
@@ -1,6 +1,4 @@
-'use strict'
-
-const DatabaseHealthCheckService = require('../../services/database_health_check.service.js')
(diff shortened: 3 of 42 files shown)


Cruikshanks:
Myself and @Jozzey pulled this together based on what I learnt in [SROC Charging Module API Switch from CommonJS to ES6 modules](https://github.com/DEFRA/sroc-charging-module-api/pull/466).

As with that attempt, we got the app running grand but were blocked by the test framework. We found it was **Lab** that stopped us and not Sinon.

Looking into solutions the only one (and we had to dig _hard_) was to use gulp to transpile the app code to CommonJS and then have Lab test the transpiled code.

It wasn't much better when we looked at alternates. Jest comes with a [big warning](https://jestjs.io/docs/ecmascript-modules) that ESM modules are experimental. It also is doing the same thing as the Lab workaround, only using Babel to do the transpiling.

A [comparison of test frameworks](https://blog.logrocket.com/comparing-best-node-js-unit-testing-frameworks/) highlights [AVA](https://github.com/avajs/ava) as the only one with Native ESM support. We'd be on our own trying to get it to work with tools like Sinon, and it'd mean a complete rewrite of our tests. As the report highlights, it is the new kid on the block.

So, our conclusion at this point is using ESM modules rather than CommonJS feels like we'd be on the fringe of things, which for a small team with _a lot_ to do is just too much risk to carry at this time.

Cruikshanks:
Superseded by #3522

PR #10, opened 2022-10-26
Refactor dotenv require

Jozzey:
Currently 'dotenv' has been required in multiple places. This only needs to be done once and should be done as early as possible in the application.


Cruikshanks (approved):
thumbs-up

PR #22, opened 2022-11-08
Create test database migrations

Jozzey:

Some migrations scripts are required to create the tables/data for the unit tests in this PR https://github.com/DEFRA/water-abstraction-system/pull/15 in order for the build process to pass.

As no migrations process had yet been set up on this application. This PR has been created to build the migrations process as well as populate the test database with the data required for the PR mentioned above.

This migration process has been based on that used by the [Charging Module](https://github.com/DEFRA/sroc-charging-module-api).


db/migrations/20221108152353_create-charge-versions.js
```diff
@@ -0,0 +1,41 @@
...
+    ON water.${tableName}
+    FOR EACH ROW
+    EXECUTE PROCEDURE update_timestamp();
+  `)
+}
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
```
Cruikshanks, line 35: I think we can drop this documentation. IMHO it's noise we don't need in the migrations. While we at it, we should add `use strict` to the top of the file.

```suggestion
```

Cruikshanks (changes requested):
I think it might benefit from some changes around the comments.

The one thing we do need to ensure is that the files include `use strict`.

But look at us, we have migrations now!

Cruikshanks (approved):

PR #32, opened 2022-11-22
Refactor Supplementary output to test presenter

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3802
https://eaflood.atlassian.net/browse/WATER-3826

We're trying to make it easy to test that we are selecting the right values from the DB when generating the SROC supplementary bill run. Our test endpoint currently lists the charge versions, which includes the licence reference. In [Add new SROC Billing Period service](https://github.com/DEFRA/water-abstraction-system/pull/30) we add the financial years.

But it would be easier to validate the licences that have been selected if the response included a unique list of them rather than having to search through the charge version results. This gives us a prime opportunity to split out the presentation logic from what the service is doing, especially as it is likely to be discarded when we are happy the process is working.

So, in this change, we move formatting the response to a new 'presenter'. As part of the change, we'll add a unique list of the Licences being used.

Example response following this change

```json
{
    "billingPeriods": [
        {
            "startDate": "2022-04-01T00:00:00.000Z",
            "endDate": "2023-03-31T00:00:00.000Z"
        }
    ],
    "licences": [
        {
            "licenceId": "2627a306-23a3-432f-9c71-a71663888285",
            "licenceRef": "AT/SROC/SUPB/01"
        }
    ],
    "chargeVersions": [
        {
            "chargeVersionId": "4b5cbe04-a0e2-468c-909e-1e2d93810ba8",
            "licenceRef": "AT/SROC/SUPB/01",
            "licenceId": "2627a306-23a3-432f-9c71-a71663888285",
            "scheme": "sroc",
            "endDate": null
        },
        {
            "chargeVersionId": "732fde85-fd3b-44e8-811f-8e6f4eb8cf6f",
            "licenceRef": "AT/SROC/SUPB/01",
            "licenceId": "2627a306-23a3-432f-9c71-a71663888285",
            "scheme": "sroc",
            "endDate": null
        }
    ]
}
```

### DIFF
--- a/app/presenters/supplementary.presenter.js
+++ b/app/presenters/supplementary.presenter.js
@@ -62,6 +62,7 @@ class SupplementaryPresenter {
     })
 
     return {
+      billingPeriods: data.billingPeriods,
       licences,
       chargeVersions
     }
--- a/app/presenters/supplementary.presenter.js
+++ b/app/presenters/supplementary.presenter.js
@@ -22,10 +22,14 @@ class SupplementaryPresenter {
    *
    * To make it easier to confirm we are including the right licences in the Supplementary bill run process we want to
    * extract a unique list of licence ID's and references.
+   *
+   * @param {Object[]} chargeVersions Results of a call to SupplementaryService and the charge version info it returns
+   *
+   * @returns {Object[]} Array of objects representing the unique licence details in the charge versions passed in
    */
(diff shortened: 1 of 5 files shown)


Jozzey (approved):
👍🏼

Jozzey (approved):

PR #37, opened 2022-11-25
New billing batch record for SROC supplementary bill run

Jozzey:
https://eaflood.atlassian.net/browse/WATER-3830

 We are starting to create the initial records needed in the SROC Supplementary process.

The first is the billing_batch record. Users can view the status of a bill run on the bill runs page. We need this record to support this, plus link/record other values, for example, the charge module’s Bill run ID.


Jozzey:
I've created a billing-batch test helper but ended up not using it as the billing batch service creates a billing batch so didn't need the helper to do it for me.

I've left the helper in for the time being in case we need it later on, but let me know if you'd like it removed.

app/models/region.model.js
```diff
@@ -23,6 +23,14 @@ class RegionModel extends BaseModel {
           from: 'water.regions.region_id',
           to: 'water.licences.region_id'
         }
+      },
+      billing_batches: {
```
Cruikshanks, line 27: Just a minor tweak. We can use camelcase for the names of the relationships and that will follow what we've already done in the `LicenceModel`.

```suggestion
      billingBatches: {
```

Cruikshanks (changes requested):
Just a real minor change. Over than that, this is bang on the money!

Cruikshanks (approved):

PR #43, opened 2022-12-07
Update licence query to just return SROC

Cruikshanks:

The original acceptance criteria for what licences to consider was anything flagged as 'include in supplementary billing'.

There are reasons why those that are void, deferred, revoked etc still need to be included. But our current scope is to generate SROC supplementary bill runs. So, on review, it makes sense that we only look at licences that have at least 1 SROC charge version.

This change updates our query to filter out anything that isn't SROC.

### DIFF
--- a/app/services/supplementary-billing/fetch-licences.service.js
+++ b/app/services/supplementary-billing/fetch-licences.service.js
@@ -26,8 +26,10 @@ class FetchLicencesService {
 
   static async _fetch (region) {
     const result = await LicenceModel.query()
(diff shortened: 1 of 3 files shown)


Cruikshanks:
Turns out our current support for working with schemas does _not_ support relations. We have implemented [this solution](https://github.com/Vincit/objection.js/issues/85) with success. But rather than pollute this PR we'll make those changes in a new PR and then rebase. Our changes should then build.

Jozzey (approved):

PR #44, opened 2022-12-07
Fix Objection model relations

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3829

Our [first attempt](https://github.com/DEFRA/water-abstraction-system/pull/43) to generate a query that uses relations is erroring.

```text
Licence.relationMappings.charge-versions: join: either `from` or `to` must point to the owner model table and the other one to the related table. It might be that specified table 'water.charge_versions` is not correct
```

Initially, we thought it was the way we're working with schemas as highlighted in [this issue](https://github.com/Vincit/objection.js/issues/85). But actually, it was just a typo.

So, this change fixes that. It also corrects some other details of the relationships and adds some tests just to ensure they are working as expected.

### DIFF
--- a/app/models/charge-version.model.js
+++ b/app/models/charge-version.model.js
@@ -16,7 +16,7 @@ class ChargeVersionModel extends BaseModel {
 
   static get relationMappings () {
     return {
-      licences: {
+      licence: {
         relation: Model.BelongsToOneRelation,
         modelClass: 'licence.model',
         join: {
--- a/app/models/charge-version.model.js
+++ b/app/models/charge-version.model.js
@@ -11,7 +11,7 @@ const BaseModel = require('./base.model.js')
 
 class ChargeVersionModel extends BaseModel {
   static get tableName () {
-    return 'water.chargeVersions'
+    return 'water.charge_versions'
   }
 
   static get relationMappings () {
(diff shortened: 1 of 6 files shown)


Cruikshanks:
Found when working on this our solution _does_ work. Including the schema in the model table names seems to work fine; we just had a typo.

We've opted to stick with our solution as it _should_ mean we can join tables across schemas which the other suggested solution wouldn't allow. But just in case that decision comes back to bite us on the 🍑 , recording the original outline for the PR here

---

Our current solution for working with schemas is to specify the schema name in the `tableName()` getter response.

That has been working fine when just working in isolation. But our [first attempt](https://github.com/DEFRA/water-abstraction-system/pull/43) to generate a query that uses relations is erroring.

```text
Licence.relationMappings.charge-versions: join: either `from` or `to` must point to the owner model table and the other one to the related table. It might be that specified table 'water.charge_versions` is not correct
```

Some digging later we found [this issue](https://github.com/Vincit/objection.js/issues/85) which highlights our solution won't work

> Setting the schema to the tableName doesn't work. knex doesn't interpret the dot as a schema separator but as a part of the table name.

So, this change implements the solution documented in that issue. We extend Objection's `QueryBuilder` with one that defaults all queries to use `withSchema()` taking the schema to use from a new model property `defaultSchema`.

Jozzey (approved):

PR #46, opened 2022-12-08
Replacing classes with modules and functions

Cruikshanks:
> Classes are a template for creating objects.

That is the first sentence of the Mozilla documentation for **Classes**. It is what those of us from other languages understand classes to be for. For some of us though (we're looking at you @Cruikshanks !) classes are _all things_!

Because of that thinking, we've allowed Classes to creep in where they shouldn't really be used. Most noticeably, things like controllers, presenters and services, are formed of purely static methods. There is no real reason to wrap their functionality in a class. They could just be expressed as modules and export their relevant functions.

That's what those experienced with JavaScript are more likely to expect. By not doing things that way we're breaking the [Principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

We believe classes are suitable for our Models, as it follows the convention of [Objection](https://vincit.github.io/objection.js/guide/models.html) and we do create instances of them. Also, classes will be our go-to if we need to create objects in the future.

But before this repo gets too far down the road we're taking this opportunity to refactor things from classes into modules.

### DIFF
--- a/app/controllers/health/airbrake.controller.js
+++ b/app/controllers/health/airbrake.controller.js
@@ -5,22 +5,22 @@
  * @module AirbrakeController
  */
 
-class AirbrakeController {
-  static async index (request, _h) {
-    // First section tests connecting to Airbrake through a manual notification
-    request.server.app.airbrake.notify({
-      message: 'Airbrake manual health check',
-      error: new Error('Airbrake manual health check error'),
-      session: {
-        req: {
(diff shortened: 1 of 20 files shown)


Cruikshanks:
We have #37 ready to be merged. Going to wait for that then will rebase this, make the changes and then make it ready for review.

Jozzey (approved):

PR #48, opened 2022-12-09
Replacing classes with modules and functions Pt.2

Cruikshanks:
This follows on from [Replacing classes with modules and functions](https://github.com/DEFRA/water-abstraction-system/pull/46) as we managed to miss some bits!

This second attempt hopefully covers all the bits we missed on the first pass of refactoring things away from classes.


Jozzey (approved):
👍🏼

PR #53, opened 2022-12-13
Add initial generic http request module

Cruikshanks:

Our `/service-status` page makes `GET` requests to the other services in order to check if they are 'alive'. We now need to start making requests to other external services, for example, the Charging Module API and AWS Cognito (to authenticate with the CHA API).

There are options we want to pass to [got](https://github.com/sindresorhus/got) all the time. Plus we want a standard way of handling the response. **got**, by default, will throw an error for anything other than a `2xx` or `3xx` response. But we normally want the response, including the status code and body as it will determine what we do next.

So, it's time to add a generic HTTP request module to centralise and standardise these things.

### DIFF
--- a/app/lib/request.lib.js
+++ b/app/lib/request.lib.js
@@ -4,6 +4,8 @@
  * @module RequestLib
  */
 
+const { HttpsProxyAgent } = require('hpagent')
+
 const requestConfig = require('../../config/request.config.js')
 
 async function get (url, additionalOptions = {}) {
@@ -48,6 +50,14 @@ async function _importGot () {
  */
 function _requestOptions (additionalOptions) {
   const defaultOptions = {
+    // This uses the spread operator and a logical AND short circuit evaluation to allow us to determine whether the
+    // following columns are added to the options or not. Thanks to https://stackoverflow.com/a/40560953/6117745 for
+    // this
+    ...(requestConfig.httpProxy && {
+      agent: {
+        https: new HttpsProxyAgent({ proxy: requestConfig.httpProxy })
+      }
+    }),
     // If we don't have this setting Got will throw its own HTTPError unless the result is 2xx or 3xx. That makes it
     // impossible to see what the status code was because it doesn't get set on the response object Got provides when
     // an error is thrown. With this set Got will treat a 404 in the same way it treats a 204.
--- a/app/lib/request.lib.js
+++ b/app/lib/request.lib.js
@@ -1,0 +1,78 @@
+'use strict'
+
+/**
+ * @module RequestLib
+ */
+
+const requestConfig = require('../../config/request.config.js')
+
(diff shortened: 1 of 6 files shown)


test/lib/request.lib.test.js
```diff
@@ -0,0 +1,189 @@
...
+        })
+      })
+
+      describe('and a retry succeeds', () => {
+        beforeEach(async () => {
+          // The first response will time out, the second response will return OK
+          Nock(testDomain)
+            .get(() => true)
+            .delay(100)
+            .reply(200, { data: 'hello world' })
```
Jozzey, line 133:
```suggestion
            .reply(200, { data: 'delayed hello world' })
```
With the data the same for both the delayed and rapid replies, if the delayed reply wasn't getting timed out the test would still pass

Jozzey (dismissed):

app/lib/request.lib.js
```diff
@@ -0,0 +1,109 @@
...
+function _requestOptions (additionalOptions) {
+  const defaultOptions = {
+    // This uses the spread operator and a logical AND short circuit evaluation to allow us to determine whether the
+    // following columns are added to the options or not. Thanks to https://stackoverflow.com/a/40560953/6117745 for
+    // this
+    ...(requestConfig.httpProxy && {
+      agent: {
+        https: new HttpsProxyAgent({ proxy: requestConfig.httpProxy })
+      }
+    }),
```
StuAA78, line 81: I don't know if using the ternary would be a slightly clearer way of doing this?

```suggestion
    // This uses the ternary operator to give either an `agent` object or an empty object, and the spread operator to
    // bring the result back into the top level of the `defaultOptions` object.
    ...(requestConfig.httpProxy
      ? {
          agent: {
            https: new HttpsProxyAgent({ proxy: requestConfig.httpProxy })
          }
        }
      : {}),
```

app/lib/request.lib.js
```diff
@@ -0,0 +1,109 @@
...
+    // Got states
+    //
+    // > It is a good practice to set a timeout to prevent hanging requests. By default, there is no timeout set.
+    timeout: {
+      request: requestConfig.timeout
+    }
+  }
+
+  // Copies properties from one object to another; Object.assign(target, source). Matching properties get overriden
+  return Object.assign(defaultOptions, additionalOptions)
```
StuAA78, line 104: Would this be a slightly simpler way of doing it maybe?
```suggestion
  return { ...defaultOptions, ...additionalOptions }
```

StuAA78 (changes requested):
Just a couple of tiny tweaks that might improve readability but I won't argue over it if you think it's fine as it is 😄

Jozzey (approved):

StuAA78 (approved):

PR #56, opened 2022-12-16
Update "create bill run" endpoint to create a bill run

StuAA78:
https://eaflood.atlassian.net/browse/WATER-3854

[When we set up our "create bill run" endpoint](https://github.com/DEFRA/water-abstraction-system/pull/51), we had it simply respond to requests with a dummy response. We now update it to actually create a bill run and an event, and return the required details in the same format as the dummy response.


Cruikshanks (approved):
😁

PR #57, opened 2022-12-16
New charge category model for SROC supplementary bill run

Jozzey:
https://eaflood.atlassian.net/browse/WATER-3861

Behind the scenes, the charge references are held in the billing_charge_categories table. It is a lookup table as the references are static. They are linked to a charge version via the charge_elements table.


app/models/billing-charge-category.model.js
```diff
@@ -0,0 +1,31 @@
...
+
+class BillingChargeCategoryModel extends BaseModel {
+  static get tableName () {
+    return 'water.billing_charge_categories'
+  }
+
+  static get relationMappings () {
+    return {
+      chargeElement: {
+        relation: Model.BelongsToOneRelation,
```
Cruikshanks, line 20: Think this relationship might need tweaking

```suggestion
        relation: Model.HasManyRelation,
```

Cruikshanks (changes requested):
I think just one of the relationships needs double checking 🤷

app/models/billing-charge-category.model.js
Jozzey: Yeah just checked, you are right. I'll commit your change.

Cruikshanks (approved):

PR #61, opened 2022-12-16
Fix missing Objection Id columns

Cruikshanks:
It was spotted whilst working on [Update "create bill run" endpoint to create a bill run](https://github.com/DEFRA/water-abstraction-system/pull/56) that we had an issue with our [Objection models](https://vincit.github.io/objection.js/).

The previous team didn't use the convention of just calling the unique identifier in each table `id`. Instead, they incorporate the table name, for example, `billing_batch_id`. **Objection** is expecting the identifier column to be `id` and is failing when in some cases.

Because of this, we need to go back and specify [idColumn](https://vincit.github.io/objection.js/api/model/static-properties.html#static-idcolumn) in all our models.

### DIFF
--- a/app/models/billing-batch.model.js
+++ b/app/models/billing-batch.model.js
@@ -14,6 +14,10 @@ class BillingBatchModel extends BaseModel {
     return 'water.billing_batches'
   }
 
+  static get idColumn () {
+    return 'billing_batch_id'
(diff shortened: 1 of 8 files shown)


Jozzey (approved):
👍🏼

PR #68, opened 2022-12-20
Refactor supplementary test code

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3854

We are about to [Update "create bill run" endpoint to create a bill run](https://github.com/DEFRA/water-abstraction-system/pull/56). Currently, everything is happening in the controller whereas our convention is to do everything in a service and keep our controllers ['skinny'](https://blog.makersacademy.com/forget-fat-models-its-time-for-skinny-controllers-and-skinny-models-a9b84ec481b7).

So, we need a service, something like `SupplementaryService`. But oh no! It already exists.

This change is about refactoring the code currently in `SupplementaryService` to a more explicit testing version so we can free up the namespace. Whilst we're at it we also rename the `/test` path to `check/` as everyone was getting confused by seeing `test/` folders in various places in the code 😁


Jozzey (approved):
👍🏼

PR #72, opened 2022-12-22
Update test helpers to use Objection

Cruikshanks:
> This is in support of [Improve model tests](https://github.com/DEFRA/water-abstraction-system/pull/71)

Currently, our test helpers create records using Knex directly. This is a result of the initial helpers being added before we implemented [Objection.js](https://vincit.github.io/objection.js/).

We want to expand some of our tests plus there is an argument to be consistent in how we interact with the tables in our DB. So, this change updates the helpers to do things via Objection and fixes any tests we break as a result.


Cruikshanks:
Abusing my tech lead powers so I can get on with #71 😁 !

PR #79, opened 2023-01-06
Replace `standard` with `standardx`

StuAA78:
***NOTE: We will not be implementing this change until either `standardx` is updated to support the latest version of `standard`, or `standard` is amended to allow setting of custom rules. See comment below for further details.***

StandardJS covers 95% of our linting needs in one simple package. However there are some things it has no opinion on which we feel strongly about -- for example, ensuring parameters are always enclosed in brackets (even in cases where a single parameter's brackets are optional).

We'd like to be able to flag these things up during linting and `standardx` sems like the perfect package -- a drop-in replacement for `standard` which allows extra rules to be added. This means we can extend our linting without the pain of installing full-fat ESLint.

We did encounter a couple of problems after removing `standard` and installing `standardx`:
- The first problem was an error `Failed to load config "standard" to extend from`. There was no consensus online how to fix this, some people suggested installing a bunch of ESLint modules, however we found that simply reinstalling `standard` did the trick.
- The second problem was a `Parsing error: Invalid ecmaVersion` error for each file. This seems to be an incompatibility between `standardx` and `eslint`, where `eslint` defaults to ecmaVersion 13 but `standardx` isn't compatible with this. Enforcing ecmaVersion 12 in our `package.json` resolved this. *Note that if `standardx` is updated then this may no longer be necessary.*

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,6 +36,7 @@
         "pino-pretty": "^9.1.1",
         "proxyquire": "^2.1.3",
         "sinon": "^15.0.1",
+        "standard": "^17.0.0",
         "standardx": "^7.0.0"
       }
     },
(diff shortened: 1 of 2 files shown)


StuAA78:
At present we feel that `standardx` isn't currently suitable for us so we're going to put this on hold for the time being, but we'll keep an eye on `standardx` and see if things change:
- It currently uses v16 of `standard` where the latest version is v17. This isn't really an issue as yet, as v17 hasn't introduced any new rules. However until it's fixed to use the latest version, there's always the possibility that when `standard` v18, v19 etc. comes in there will be additional rules that we won't be using. So we won't consider using it until that is resolved. [A PR for this is currently awaiting approval.](https://github.com/standard/standardx/pull/42)
- As an alternative, [an issue exists proposing that `standard` supports additional rules](https://github.com/standard/standard/issues/1536). It seems to have a bit of support behind it but as yet there doesn't seem to be any official yes or no.

So for now, we mark this as "do not merge" and will keep an eye on the state of `standardx` and `standard`.

PR #81, opened 2023-01-10
Add logic 2 determine billable abstraction periods

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3834

In [Implement abstraction period query using Objection](https://github.com/DEFRA/water-abstraction-system/pull/66) we amended the query used in `test/services/supplementary-billing/fetch-charge-versions.service.test.js` to include abstraction period information for each charge version candidate.

We're aiming to get to a 'billable days' value but the next step is determining for each abstraction period whether it is relevant to the current billing period, and if so, how much of it.

This change implements a service which can be used to answer both those questions.


Cruikshanks (commented):
I need to add some documentation and my feelings on my first attempt at solving the problem!

But once I've done that this is all good to go 😁

PR #82, opened 2023-01-10
Determine if licence is 'billed'

Jozzey:
https://eaflood.atlassian.net/browse/WATER-3868

We need to know whether a licence has already been ‘billed’ in the current period. If so we’ll have to ensure it’s credited as part of the bill run before we recalculate the debit.

For each licence that is a candidate for billing, we need to determine if it was included in a `billing_batch` for the current period.

For now, the result of this check should be included in the `/check/supplementary` endpoint results, against each licence. That way we can verify if it is working correctly.


Jozzey:
Initially this ticket referred to the charge version. Following a chat with Alan this was changed to the licence after digging into the structure of the data.

app/presenters/check/supplementary-data.presenter.js
```diff
@@ -21,6 +24,18 @@ function go (data) {
   }
 }
 
+function _licenceExistsInBilling (billingInvoiceLicences) {
+  if (billingInvoiceLicences) {
+    for (let i = 0; i < billingInvoiceLicences.length; i++) {
+      if (billingInvoiceLicences[i].billingInvoice) {
+        return true
+      }
+    }
```
Cruikshanks, line 33: I did a quick test locally (in my console) and I think you can slim this down to the following, whilst IMHO, it still makes sense what it's doing

```suggestion
    return billingInvoiceLicences.some((billingInvoiceLicence) => {
        return billingInvoiceLicence.billingInvoice
    })
```

This might even be the exception to our convention of 'no inline methods'.

```javascript
    return billingInvoiceLicences.some((billingInvoiceLicence) => billingInvoiceLicence.billingInvoice)
```

What do you think?

Cruikshanks (commented):
I've just got a possible suggestion for using one of the built-in Array functions to determine if an array contains something, rather than a for loop.

What do you think?

app/presenters/check/supplementary-data.presenter.js
Jozzey: Yep that works, I'll make the change

Cruikshanks (approved):

PR #83, opened 2023-01-10
Request new bill run in Charging Module API

StuAA78:

As part of the bill run process, we need to send a request to the Charging Module API to create a new bill run. We therefore update `InitiateBillingBatchService` to call the recently-created `ChargingModuleCreateBillRunService` to do this, and store the returned id in the `billing_batch` record as `externalId`.

In terms of error handling we simply rely on the individual stages throwing an error, which we then catch in the controller. The one exception is the result of `ChargingModuleCreateBillRunService`, which always returns a `{ succeeded: ..., response: ... }` tuple. We therefore check for failure and throw an error (which as before is caught by the controller).

The controller takes the caught error and reports a `500` error back. Note that Boom does not expose the message of a `500` error in its response; if we want to see the message we would need to check the `system` logs.


app/services/charging-module/create-bill-run.service.js
```diff
@@ -5,11 +5,12 @@
...
-const ChargeModuleTokenService = require('./charge-module-token.service.js')
-const RegionModel = require('../models/water/region.model.js')
-const RequestLib = require('../lib/request.lib.js')
+const RegionModel = require('../../models/water/region.model.js')
+const RequestLib = require('../../lib/request.lib.js')
 
-const servicesConfig = require('../../config/services.config.js')
+const servicesConfig = require('../../../config/services.config.js')
+
+const ChargingModuleTokenService = require('./token.service.js')
```
Cruikshanks, line 13: With the dotenv fix can this now go back with the rest of the module `require()` statements?

Cruikshanks (changes requested):
Just one minor tweak that I'm hoping is a hangover from before the dotenv issue was resolved.

app/services/charging-module/create-bill-run.service.js
StuAA78: Yes good catch -- reverted this now

Jozzey (approved):

Cruikshanks (approved):

PR #85, opened 2023-01-10
Make timestamps consistent at model layer

Cruikshanks:
https://github.com/DEFRA/water-abstraction-team/issues/69

We identified that there are inconsistencies in the legacy database, timestamps in the tables being one of them. We don't want these issues to make a mess of our code. So, until we're ready for some full-scale data migration we're going to deal with them in our model layer.

Timestamps being different across tables is the main inconsistency we noted and this change deals with it.

We use [Objection.js](https://vincit.github.io/objection.js/api/model/instance-methods.html#parsedatabasejson) `$parseDatabaseJson()` and `$formatDatabaseJson()` to present a consistent API in our models whilst still working with the DB. This is built into the `LegacyBaseModel` with every child model expected to include a `translations()` getter that will be used by both. There is an argument to move `translations()` into `WaterBaseModel` and then just override it for exceptions like `EventModel`. But we have an eye to also use this logic to make the ID property for each model _actually_ `.id`. And that will be different in each which means every model is going to need to set `translations ()`. So, it'd be better to keep the translations together in one place for each model, rather than splitting them.

> WARNING! Queries cannot use the 'code' version of a column name. They must use the database version. We've included a unit test to demonstrate this and documentation to explain the reasons why.


Cruikshanks:
If it helps, most of the changes are documenting what is going on! 😳 Also, for @StuAA78 benefit the key section you need to read is this

> But that leads us to the second problem; at this level translations are global. So, any reference to `date_created` would get translated to `created_at`, irrespective of the table referenced. If we only had to worry about database to model it would be okay. The problem is to make queries work we need to be able to translate back from our names to the database version. That is impossible where `created_at` might need to become `date_created`, `created`, or `modified`. With no access to the table name for context we cannot make that decision.
>
> This is why we have accepted we can only support translations when dealing with instances of a model for now.

Hopefully that explains why I had to abandon what we discussed and accept the 'middle option' 😬

Jozzey (dismissed):

app/models/legacy-base.model.js
```diff
@@ -65,6 +65,159 @@ class LegacyBaseModel extends BaseModel {
...
+   * If there are no translations an empty array is valid.
+   *
+   * @example
+   * ```javascript
+   * static get translations () {
+   *  return [
+   *    { database: 'created', model: 'createdAt' },
+   *    { database: 'modified', model: 'updatedAt' }
+   *  ]
+   *}
```
StuAA78, line 89: This is really picky but once I saw it I couldn't unsee it! It just needs a space to line up with L84
```suggestion
   * }
```

app/models/legacy-base.model.js
```diff
@@ -65,6 +65,159 @@ class LegacyBaseModel extends BaseModel {
...
+
+  /**
+   * Translations for non-standard column names to our standardised ones
+   *
+   * The primary example of non-standard column names in the legacy data is timestamps. For example, we have found the
+   * column which has the timestamp for when a record is created has been named `date_created`, 'created', 'modified',
+   * and 'created_at`.
+   *
+   * Within our code we want to just be using `created_at`, in this example, as this is what the field would have been
+   * called if Knex had been used to generate the migrations and build the tables.
```
StuAA78, line 77: As discussed we probably want to distinguish between what we would want to see the column as (ie. `created_at`) and how we want to refer to it in our code (ie. `createdAt`)

StuAA78 (changes requested):
Looks great! Just a couple of comment tweaks as discussed

Jozzey (approved):

PR #97, opened 2023-01-24
SRoC Supplementary Billing - Determine Abstraction Period

Jozzey:
https://eaflood.atlassian.net/browse/WATER-3834

As part of determining the transactions and Billable Days for Supplementary Billing, the Abstraction Period needs to be taken into account.  Abstraction period is the date range in which Licence Holders can abstract water, in terms of Supplementary Billing, then it needs to be taken into account days when the Licence Holder cannot abstract as the dates are outside of the Abstraction Period and therefore not chargeable.

This PR is to determine the correct number of billable days in the abstraction period.


app/services/supplementary-billing/abstraction-billing-period.service.js
```diff
@@ -137,23 +139,61 @@ function _abstractionPeriods (billingPeriod, chargePurpose) {
...
 function _calculateBillableDays (abstractionPeriods) {
-  let difference
-  let billableDays
   for (const abstractionPeriod of abstractionPeriods) {
-    difference = abstractionPeriod.endDate.getTime() - abstractionPeriod.startDate.getTime() // difference in msecs
-    billableDays = Math.ceil(difference / (1000 * 3600 * 24)) + 1 // (1000 msecs * (60 secs * 60 mins) * 24 hrs)
-    abstractionPeriod.billableDays = billableDays
+    if (abstractionPeriod.billableStartDate) {
+      const difference = abstractionPeriod.billableEndDate.getTime() - abstractionPeriod.billableStartDate.getTime() // difference in msecs
+      const billableDays = Math.ceil(difference / (1000 * 3600 * 24)) + 1 // (1000 msecs * (60 secs * 60 mins) * 24 hrs)
```
StuAA78, line 158: Just for clarity is it worth maybe putting the `24 * 60 * 60 * 1000` calculation as a constant defined at the top, eg. `DAYS_IN_MILLISECONDS` or something like that? And maybe stick a comment in here to explain what the `billableDays` calculation is doing?

app/services/supplementary-billing/abstraction-billing-period.service.js
```diff
@@ -137,23 +139,61 @@ function _abstractionPeriods (billingPeriod, chargePurpose) {
...
+ * @param {Object} billingPeriod Object that has a `startDate` and `endDate` that defines the billing period
+ *
+ * @returns {Object[]} The array abstraction periods each with new `billableStartDate` & `billableEndDate` properties
+ * @returns {Date} periods[].startDate
+ * @returns {Date} periods[].endDate
+ * @returns {boolean} periods[].consider
+ * @returns {Date} periods[].billableStartDate
+ * @returns {Date} periods[].billableEndDate
+ */
+function _calculateBillablePeriods (abstractionPeriods, billingPeriod) {
```
StuAA78, line 177: This function is taking the second param and using it to modify the first; but `_flagPeriodsForConsideration` takes its first param and uses it to modify the second. Personally I think the way this function does it makes more sense (I'm pretty sure it's more consistent in JS that the first param is the "target") but do we maybe want them both to work in the same way for consistency? Either by changing this function or the other one?

app/services/supplementary-billing/abstraction-billing-period.service.js
```diff
@@ -137,23 +139,61 @@ function _abstractionPeriods (billingPeriod, chargePurpose) {
...
+  }
+}
+
+/**
+ * Calculates the 'billable period' for each abstraction billing period
+ *
+ * @param {Object[]} abstractionPeriods An array of abstraction billing periods
+ * @param {Object} billingPeriod Object that has a `startDate` and `endDate` that defines the billing period
+ *
+ * @returns {Object[]} The array abstraction periods each with new `billableStartDate` & `billableEndDate` properties
```
StuAA78, line 170: This applies to `_flagPeriodsForConsideration` as well, but we're saying that this function returns something when really it's modifying the object that's passed to it. More of an observation really as I don't know what the right way to document this behaviour would be!

app/services/supplementary-billing/abstraction-billing-period.service.js
```diff
@@ -137,23 +139,61 @@ function _abstractionPeriods (billingPeriod, chargePurpose) {
...
+ * @returns {Date} periods[].startDate
+ * @returns {Date} periods[].endDate
+ * @returns {boolean} periods[].consider
+ * @returns {Date} periods[].billableStartDate
+ * @returns {Date} periods[].billableEndDate
```
StuAA78, line 175: Is it worth clarifying that `billableStartDate` and `billableEndDate` aren't always set? Only if the start date is on or before the end date? (Same applies for L146 I guess, where we don't always set the `billableDays` property, just if `billableStartDate` is defined)

StuAA78 (changes requested):

app/services/supplementary-billing/abstraction-billing-period.service.js
Cruikshanks: Defo be consistent, but treat my `_flagPeriodsForConsideration` as the bady guy. What you said about 'target' makes sense so I would update it to be consistent with `_calculateBillablePeriods`

app/services/supplementary-billing/abstraction-billing-period.service.js
Cruikshanks: That is a valid point. That's my bad for getting trigger-happy with the JSDoc for `_flagPeriodsForConsideration ()`. Where we are not explicitly returning anything we shouldn't be adding `@returns` 🤦

app/services/supplementary-billing/abstraction-billing-period.service.js
Cruikshanks: See my previous comment. We should just scrub `@returns`. Much less confusing.

app/services/supplementary-billing/abstraction-billing-period.service.js
```diff
@@ -46,6 +46,8 @@ function go (billingPeriod, chargePurpose) {
 
   _flagPeriodsForConsideration(billingPeriod, abstractionPeriods)
 
+  _calculateBillablePeriods(abstractionPeriods, billingPeriod)
+
   _calculateBillableDays(abstractionPeriods)
```
Cruikshanks, line 51: My suggestion would be that we don't bother to call these methods if the abstraction period is not to be considered. IMHO it's a waste of CPU cycles. That would mean some refactoring.

One way would be rather than `_calculateBillablePeriods()` and `_calculateBillableDays()` taking the whole array, instead you would iterate through `abstractionPeriods` and where a period is flagged for consideration, just pass it through to these methods (which would then be refactored to work on a single instance).

Or, use [filter()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) to extract a subset of periods to be considered and just pass that through to these methods.

Cruikshanks (changes requested):
@StuAA78 has raised some good points which are mainly hangovers from by first stab at this service 😳 😬 .

But I also believe there is no point working out billable dates and billable days for an abstraction period that isn't going to be considered.

app/services/supplementary-billing/abstraction-billing-period.service.js
Jozzey: I shall sort that out

app/services/supplementary-billing/abstraction-billing-period.service.js
Jozzey: Agreed. I guess given that this function could potentially be getting called many many times per bill run it would make sense to make it as efficient as possible.

app/services/supplementary-billing/abstraction-billing-period.service.js
Jozzey: I'll get rid of the `@returns`

app/services/supplementary-billing/abstraction-billing-period.service.js
Cruikshanks: Yep, that would defo do the trick as well. Probably clearer that we're only returning out of the service stuff that needs dealing with, so the calling service doesn't have to do any filtering of its own.

app/services/supplementary-billing/abstraction-billing-period.service.js
```diff
@@ -133,56 +126,86 @@ function _abstractionPeriods (billingPeriod, chargePurpose) {
...
  *
- * @param {Object[]} abstractionPeriods An array of abstraction billing periods
+ * @param {Object} abstractionPeriod An abstraction billing period
  *
- * @returns {Object[]} The array abstraction periods each with a new `billableDays` property
- * @returns {Date} periods[].startDate
- * @returns {Date} periods[].endDate
- * @returns {boolean} periods[].consider
- * @returns {number} periods[].billableDays
+ * @returns {Object} The abstraction period with a new `billableDays` property
```
Cruikshanks, line 146: Following on from @StuAA78 original point, if we're amending the value passed in and not explicitly returning anything we shouldn't include `@returns` in our JSDoc.

```suggestion
```

app/services/supplementary-billing/abstraction-billing-period.service.js
```diff
@@ -133,56 +126,86 @@ function _abstractionPeriods (billingPeriod, chargePurpose) {
...
 }
 
 /**
- * Adds the `consider` flag to each abstraction billing period as to whether it should be included in billing
+ * Calculates the 'billable period' for an abstraction billing period
+ *
+ * @param {Object} abstractionPeriod An abstraction billing period
+ * @param {Object} billingPeriod Object that has a `startDate` and `endDate` that defines the billing period
+ *
+ * @returns {Object} The abstraction period with new `billableStartDate` & `billableEndDate` properties
```
Cruikshanks, line 163: Following on from @StuAA78 original point, if we're amending the value passed in and not explicitly returning anything we shouldn't include `@returns` in our JSDoc.

```suggestion
```

app/services/supplementary-billing/abstraction-billing-period.service.js
```diff
@@ -133,56 +126,86 @@ function _abstractionPeriods (billingPeriod, chargePurpose) {
...
+ */
+function _calculateBillablePeriods (abstractionPeriod, billingPeriod) {
```
Cruikshanks, line 165: Now we're only looking at an individual period, I think it would be clearer to tweak the name.

```suggestion
function _calculateBillablePeriod (abstractionPeriod, billingPeriod) {
```

app/services/supplementary-billing/abstraction-billing-period.service.js
```diff
@@ -23,30 +23,22 @@
...
- * @returns {number} periods[].billableDays
  */
 function go (billingPeriod, chargePurpose) {
   const abstractionPeriods = _abstractionPeriods(billingPeriod, chargePurpose)
 
-  _flagPeriodsForConsideration(billingPeriod, abstractionPeriods)
-
-  _calculateBillableDays(abstractionPeriods)
+  for (const abstractionPeriod of abstractionPeriods) {
+    _calculateBillablePeriods(abstractionPeriod, billingPeriod)
```
Cruikshanks, line 39: Now we're only looking at an individual period, I think it would be clearer to tweak the name.

```suggestion
    _calculateBillablePeriod(abstractionPeriod, billingPeriod)
```

Cruikshanks (changes requested):
Just a couple of minor things. I even did them as suggestions to try and make it quick! 😬

Cruikshanks (approved):

StuAA78 (approved):

PR #98, opened 2023-01-25
Do not create SROC bill run if one already exists

StuAA78:
https://eaflood.atlassian.net/browse/WATER-3875

The legacy system will not create a presroc bill run if one exists that is "live" (ie. has a status of `processing`, `ready` or `review`) with the same region, financial year and type. The same logic needs to be implemented for sroc supplementary bill runs (adding our new `queued` status to the list of statuses we consider to be "live").

We do this by creating `CheckLiveBillRunService`, which takes a region id and financial year and returns a boolean to indicate whether there are any "live" bill runs. We then call this from `InitiateBillingBatchService` and throw an error if it returns `true`. Note that we give the error the same message that the `water-abstraction-service` has but we don't format it as a `409` error.

### DIFF
--- a/app/services/supplementary-billing/check-live-bill-run.service.js
+++ b/app/services/supplementary-billing/check-live-bill-run.service.js
@@ -1,0 +1,26 @@
+'use strict'
+
+/**
+ * Checks whether a "live" bill run exists for the specified region, scheme, type and financial year
+ * @module CheckLiveBillRunService
+ */
+
+/**
+ * Check whether a "live" bill run exists for the specified region, scheme, type and financial year
+ *
+ * We define "live" as having the status `processing`, `ready` or `review`
+ *
+ * @param {*} region
+ * @param {*} scheme
+ * @param {*} type
+ * @param {*} financialYeara
+ *
+ * @returns {Boolean} Whether a "live" bill run exists
+ */
+async function go (region, scheme, type, financialYear) {
+  return true
+}
+
+module.exports = {
+  go
+}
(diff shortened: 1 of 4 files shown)


app/services/supplementary-billing/check-live-bill-run.service.js
```diff
@@ -0,0 +1,38 @@
...
+ * We define "live" as having the status `processing`, `ready`, `review` or `queued`
+ *
+ * @param {*} regionId The id of the region to be checked
+ * @param {*} financialYear The financial year to be checked
+ *
+ * @returns {Boolean} Whether a "live" bill run exists
+ */
+async function go (regionId, financialYear) {
+  const numberOfLiveBillRuns = await BillingBatchModel.query()
+    .where({
```
Cruikshanks, line 24: It's a negligible tweak. But do you think there is value in adding a `select('billing_batch_id`)` to the query to limit the amount of data being returned?

We only care about the number of records returned, not the data in those records 🤷

Cruikshanks (commented):
It's not critical. More in the 'every penny helps' category. What do you think?

app/services/supplementary-billing/check-live-bill-run.service.js
StuAA78: Yep that's fair enough, I somehow managed to miss the bit of the docs for `.resultSize()` that says it does actually run that query!

Cruikshanks (approved):

Jozzey (approved):
👍🏼

PR #99, opened 2023-01-27
Flag bill runs as errored on CHA failure

Cruikshanks:

If we fail to create the bill run in the [Charging Module API](https://github.com/DEFRA/sroc-charging-module-api) we are currently handling the error. We ensure the app doesn't fall over, the issue gets logged and a `5xx` is returned to the client.

What we are not doing is updating the `billing_batch` we created with a status of error. This means as far as the UI is concerned it's **QUEUED** waiting to be processed.

This change updates the `billing_batch` record if the request to the CHA fails so that it correctly shows as errored in the UI.


Cruikshanks:
Superseded by https://github.com/DEFRA/water-abstraction-system/pull/104

PR #101, opened 2023-01-27
Determine if licence is 'billed'

Jozzey:
https://eaflood.atlassian.net/browse/WATER-3868

We need to know whether a licence has already been ‘billed’ in the current period. If so we’ll have to ensure it’s credited as part of the bill run before we recalculate the debit.

For each licence that is a candidate for billing, we need to determine if it was included in a billing_batch for the current period.

### DIFF
--- a/test/services/supplementary-billing/check-live-bill-run.service.test.js
+++ b/test/services/supplementary-billing/check-live-bill-run.service.test.js
@@ -24,7 +24,7 @@ describe('Check Live Bill Run service', () => {
   describe('when an sroc supplementary bill run exists for this region and financial year', () => {
     describe('with a status considered to be "live"', () => {
       beforeEach(async () => {
-        billRun = await BillingBatchHelper.add()
+        billRun = await BillingBatchHelper.add({ status: 'queued' })
       })
 
       it('returns `true`', async () => {
@@ -36,7 +36,7 @@ describe('Check Live Bill Run service', () => {
 
     describe('with a status not considered to be "live"', () => {
       beforeEach(async () => {
-        billRun = await BillingBatchHelper.add({ status: 'sent' })
+        billRun = await BillingBatchHelper.add()
       })
 
       it('returns `false`', async () => {
--- a/test/services/supplementary-billing/check-live-bill-run.service.test.js
+++ b/test/services/supplementary-billing/check-live-bill-run.service.test.js
@@ -24,7 +24,7 @@ describe('Check Live Bill Run service', () => {
   describe('when an sroc supplementary bill run exists for this region and financial year', () => {
     describe('with a status considered to be "live"', () => {
       beforeEach(async () => {
-        billRun = await BillingBatchHelper.add({ status: 'queued' })
+        billRun = await BillingBatchHelper.add()
       })
 
       it('returns `true`', async () => {
@@ -36,7 +36,7 @@ describe('Check Live Bill Run service', () => {
 
     describe('with a status not considered to be "live"', () => {
       beforeEach(async () => {
-        billRun = await BillingBatchHelper.add()
+        billRun = await BillingBatchHelper.add({ status: 'sent' })
       })
 
       it('returns `false`', async () => {
--- a/test/services/supplementary-billing/fetch-licences.service.test.js
+++ b/test/services/supplementary-billing/fetch-licences.service.test.js
@@ -32,7 +32,7 @@ describe('Fetch Licences service', () => {
         testLicence = await LicenceHelper.add({ includeInSupplementaryBilling: 'yes' })
       })
 
-      describe('Licence not previously billed', () => {
+      describe('and that have not been previously billed', () => {
         it('returns the expected results', async () => {
           const result = await FetchLicencesService.go(region, billingPeriodFinancialYearEnding)
 
--- a/test/services/supplementary-billing/fetch-licences.service.test.js
+++ b/test/services/supplementary-billing/fetch-licences.service.test.js
@@ -94,7 +94,7 @@ describe('Fetch Licences service', () => {
         })
       })
 
-      describe('and has been billed twice in current period 2023', () => {
+      describe('and that have been billed twice in the current period', () => {
         beforeEach(async () => {
           billingInvoice = await BillingInvoiceHelper.add({ financialYearEnding: billingPeriodFinancialYearEnding })
           await BillingInvoiceLicenceHelper.add({}, testLicence, billingInvoice)
--- a/test/services/supplementary-billing/fetch-licences.service.test.js
+++ b/test/services/supplementary-billing/fetch-licences.service.test.js
@@ -33,7 +33,7 @@ describe('Fetch Licences service', () => {
       })
 
       describe('Licence not previously billed', () => {
-        it('returns results', async () => {
+        it('returns the expected results', async () => {
           const result = await FetchLicencesService.go(region, billingPeriodFinancialYearEnding)
 
           expect(result.length).to.equal(1)
--- a/test/services/supplementary-billing/fetch-licences.service.test.js
+++ b/test/services/supplementary-billing/fetch-licences.service.test.js
@@ -112,7 +112,7 @@ describe('Fetch Licences service', () => {
         })
       })
 
-      // This situation should not occur normally, but as the billing batch is filtered on scheme I'm going to test it
+      // NOTE: This situation will not occur normally. But as the billing batch is filtered on `scheme` we wanted a test to ensure the filter worked
       describe('Licence previously billed in current period 2023 but scheme not sroc', () => {
         beforeEach(async () => {
           billingInvoice = await BillingInvoiceHelper.add(
--- a/test/services/supplementary-billing/fetch-licences.service.test.js
+++ b/test/services/supplementary-billing/fetch-licences.service.test.js
@@ -113,7 +113,7 @@ describe('Fetch Licences service', () => {
       })
 
       // NOTE: This situation will not occur normally. But as the billing batch is filtered on `scheme` we wanted a test to ensure the filter worked
-      describe('Licence previously billed in current period 2023 but scheme not sroc', () => {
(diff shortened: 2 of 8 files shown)


test/services/supplementary-billing/check-live-bill-run.service.test.js
```diff
@@ -24,7 +24,7 @@ describe('Check Live Bill Run service', () => {
   describe('when an sroc supplementary bill run exists for this region and financial year', () => {
     describe('with a status considered to be "live"', () => {
       beforeEach(async () => {
-        billRun = await BillingBatchHelper.add()
+        billRun = await BillingBatchHelper.add({ status: 'queued' })
```
Cruikshanks, line 27: Going to comment here but also applies to line 39 in the same file and the change to `test/support/helpers/water/billing-batch.helper.js`.

`queued` is the default state for a new BillingBatch record and that is why the helper uses that as its default. Is there a reason why we have switched the default to `sent`?

test/services/supplementary-billing/fetch-licences.service.test.js
```diff
@@ -28,29 +27,44 @@ describe('Fetch Licences service', () => {
   })
 
   describe('when there are licences for the matching region', () => {
-    describe('that are flagged to be included in supplementary billing', () => {
+    describe('that are flagged to be included in supplementary billing.', () => {
```
Cruikshanks, line 30: Ideally, the context block descriptions should 'flow', following on from the previous to form documentation on how the app is expected to behave.

This provides the context for subsequent comments, but why we also wouldn't expect to see the description end with '.'

```suggestion
    describe('that are flagged to be included in supplementary billing', () => {
```

test/services/supplementary-billing/fetch-licences.service.test.js
```diff
@@ -28,29 +27,44 @@ describe('Fetch Licences service', () => {
...
       beforeEach(async () => {
         testLicence = await LicenceHelper.add({ includeInSupplementaryBilling: 'yes' })
       })
 
-      describe('and that have an SROC charge version. Licence not previosly billed', () => {
+      describe('Licence not previously billed', () => {
+        it('returns results', async () => {
```
Cruikshanks, line 36:
```suggestion
        it('returns the expected results', async () => {
```

test/services/supplementary-billing/fetch-licences.service.test.js
```diff
@@ -28,29 +27,44 @@ describe('Fetch Licences service', () => {
...
 
   describe('when there are licences for the matching region', () => {
-    describe('that are flagged to be included in supplementary billing', () => {
+    describe('that are flagged to be included in supplementary billing.', () => {
       beforeEach(async () => {
         testLicence = await LicenceHelper.add({ includeInSupplementaryBilling: 'yes' })
       })
 
-      describe('and that have an SROC charge version. Licence not previosly billed', () => {
+      describe('Licence not previously billed', () => {
```
Cruikshanks, line 35:
```suggestion
      describe('and that have not been previously billed', () => {
```

test/services/supplementary-billing/fetch-licences.service.test.js
```diff
@@ -28,29 +27,44 @@ describe('Fetch Licences service', () => {
...
+      })
+
+      describe('Licence previously billed in current period 2023', () => {
         beforeEach(async () => {
-          await ChargeVersionHelper.add({}, testLicence)
+          billingInvoice = await BillingInvoiceHelper.add({ financialYearEnding: billingPeriodFinancialYearEnding })
+          await BillingInvoiceLicenceHelper.add({}, testLicence, billingInvoice)
         })
 
         it('returns results', async () => {
```
Cruikshanks, line 52:
```suggestion
        it('returns the expected results', async () => {
```

test/services/supplementary-billing/fetch-licences.service.test.js
```diff
@@ -28,29 +27,44 @@ describe('Fetch Licences service', () => {
...
+          const result = await FetchLicencesService.go(region, billingPeriodFinancialYearEnding)
+
+          expect(result.length).to.equal(1)
+          expect(result[0].licenceId).to.equal(testLicence.licenceId)
+          expect(result[0].licenceRef).to.equal(testLicence.licenceRef)
+          expect(result[0].numberOfTimesBilled).to.equal(0)
+        })
+      })
+
+      describe('Licence previously billed in current period 2023', () => {
```
Cruikshanks, line 46:
```suggestion
      describe('and that have been billed in the current period', () => {
```

test/services/supplementary-billing/fetch-licences.service.test.js
```diff
@@ -28,29 +27,44 @@ describe('Fetch Licences service', () => {
...
           expect(result.length).to.equal(1)
           expect(result[0].licenceId).to.equal(testLicence.licenceId)
-          expect(result[0].billingInvoiceLicences).to.equal([])
+          expect(result[0].licenceRef).to.equal(testLicence.licenceRef)
+          expect(result[0].numberOfTimesBilled).to.equal(1)
         })
       })
 
-      describe('and that have an SROC charge version. Licence previosly billed in current period 2023', () => {
+      describe('Licence previously billed in current period 2023, billing batch status not `sent`', () => {
```
Cruikshanks, line 62:
```suggestion
      describe("and that are included in an 'unsent' billing batch in the current period", () => {
```

test/services/supplementary-billing/fetch-licences.service.test.js
```diff
@@ -59,13 +73,13 @@ describe('Fetch Licences service', () => {
...
           expect(result.length).to.equal(1)
           expect(result[0].licenceId).to.equal(testLicence.licenceId)
-          expect(result[0].billingInvoiceLicences[0].billingInvoice.financialYearEnding).to.equal(billingPeriodFinancialYearEnding)
+          expect(result[0].licenceRef).to.equal(testLicence.licenceRef)
+          expect(result[0].numberOfTimesBilled).to.equal(0)
         })
       })
 
-      describe('and that have an SROC charge version. Licence previosly billed in previous period 2022', () => {
+      describe('Licence previosly billed in previous period 2022', () => {
```
Cruikshanks, line 81:
```suggestion
      describe('and that have been billed in the previous period', () => {
```

test/services/supplementary-billing/fetch-licences.service.test.js
```diff
@@ -75,29 +89,46 @@ describe('Fetch Licences service', () => {
...
           expect(result.length).to.equal(1)
           expect(result[0].licenceId).to.equal(testLicence.licenceId)
-          expect(result[0].billingInvoiceLicences[0].billingInvoice).to.equal(null)
+          expect(result[0].licenceRef).to.equal(testLicence.licenceRef)
+          expect(result[0].numberOfTimesBilled).to.equal(0)
         })
       })
 
-      describe('and that have multiple SROC charge versions', () => {
+      describe('and has been billed twice in current period 2023', () => {
```
Cruikshanks, line 97:
```suggestion
      describe('and that have been billed twice in the current period', () => {
```

test/services/supplementary-billing/fetch-licences.service.test.js
```diff
@@ -75,29 +89,46 @@ describe('Fetch Licences service', () => {
...
 
           expect(result.length).to.equal(1)
           expect(result[0].licenceId).to.equal(testLicence.licenceId)
+          expect(result[0].licenceRef).to.equal(testLicence.licenceRef)
+          expect(result[0].numberOfTimesBilled).to.equal(2)
         })
       })
 
-      describe('but do not have an SROC charge version', () => {
+      // This situation should not occur normally, but as the billing batch is filtered on scheme I'm going to test it
```
Cruikshanks, line 115:
```suggestion
      // NOTE: This situation will not occur normally. But as the billing batch is filtered on `scheme` we wanted a test to ensure the filter worked
```

test/services/supplementary-billing/fetch-licences.service.test.js
```diff
@@ -75,29 +89,46 @@ describe('Fetch Licences service', () => {
...
+      describe('Licence previously billed in current period 2023 but scheme not sroc', () => {
```
Cruikshanks, line 116:
```suggestion
      describe('and that have been billed in the current period under a different scheme', () => {
```

test/support/helpers/water/billing-batch.helper.js
```diff
@@ -45,7 +45,7 @@ function defaults (data = {}) {
     batchType: 'supplementary',
     fromFinancialYearEnding: 2023,
     toFinancialYearEnding: 2023,
-    status: 'queued',
+    status: 'sent',
```
Cruikshanks, line 48: See the comment about changing from `queued` to `sent` being the default.

Cruikshanks (changes requested):
I'm questioning the change in the default `status` used by the BillingBatchHelper. But if there is good reason for it I'm not going to argue!

But I have made some suggested changes to try and get the unit test descriptions 'flowing' again.

test/services/supplementary-billing/check-live-bill-run.service.test.js
Jozzey: I defaulted it to `sent` as there was only 2 tests that used that default value `queued`. As the majority of the tests use the status of `sent` I found the helper to be much more helpful if this was the default.

Cruikshanks: I think I'd still argue that the default matches with what would happen when the record is newly created and that we then apply the status we need in the tests.

It might be we have more tests that use 'sent' now. But that might change in the future. Plus, the status being `sent` is important in these tests whereas it might be meaningless in others. So, IMHO it's better to set the state where it matters. 🤷

Jozzey: You've talked me round :grin: I shall change the default back to `queued`

Cruikshanks (approved):

PR #109, opened 2023-02-06
Add transaction service to water-abstraction-system

Jozzey:
https://eaflood.atlassian.net/browse/WATER-3894

Thanks to https://github.com/DEFRA/water-abstraction-system/pull/97 we now know what the billable days are for each abstraction period (`charge_purpose`) linked to a charge reference (`charge_element` && `billing_charge_categories`) within a charge version.

The final step is to take that information and start generating the transaction line information. For each charge reference in a charge version, there should be one transaction line. So, if a charge reference is linked to multiple abstraction periods, we need the sum of their billable days.


app/services/supplementary-billing/create-transactions.service.js
```diff
@@ -0,0 +1,41 @@
+'use strict'
+
+/**
+ * Creates transaction lines based on the charge referece and abstraction period
```
Cruikshanks, line 4:
```suggestion
 * Generates the transaction lines for an supplementary bill run
```

app/services/supplementary-billing/create-transactions.service.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+const AbstractionBillingPeriodService = require('./abstraction-billing-period.service.js')
+
+/**
+ * @param {Object} billingPeriod Object that has a `startDate` and `endDate` that defines the billing period
+ * @param {Object[]} chargeElements An array of `chargeElements` containing an array of `chargePurposes` which define
+ * the abstraction periods
+ *
+ * @returns {Object[]} An array that has the `reference` and `billableDays` for each charge element on a charge version
```
Cruikshanks, line 15:
```suggestion
 * Creates a transaction line and calculates the billable days for each charge element
 * 
 * Each SROC charge version linked to a licence will have one or more charge elements that are linked to a charge reference.
 *
 * A transaction line is needed for each of these charge references. Linked to the charge element are one or more abstraction periods.
 *
 * We need to calculate the sum of the billable days for these periods and apply it to the transaction line.
 *
 * @param {Object} billingPeriod Object that has a `startDate` and `endDate` that defines the billing period
 * @param {Object[]} chargeElements An array of `chargeElements` containing an array of `chargePurposes` which define
 * the abstraction periods
 *
 * @returns {Object[]} An array that has the `reference` and `billableDays` for each charge element on a charge version
```

Cruikshanks (changes requested):
It is only the comments! 😁

Cruikshanks (approved):

PR #119, opened 2023-02-13
Add SROC Supplementary Billing Invoice Service

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3896

In order to 'piggy-back' onto the existing functions in the legacy service for reviewing, confirming, sending, viewing, deleting a bill run etc we need to create the same base database records as it does for a Bill run.

This change will add a service to handle creating the `water.billing_batch_invoice` records in the DB. This is the link between the `billing_batch` and the invoice account, and from there to the licences included in the bill run.

We'll need to dig into the legacy code to understand where the previous team extracts the data that forms the `billing_invoice` records. We'll then aim to replicate that here.


app/services/supplementary-billing/create-billing-invoice.service.js
```diff
@@ -0,0 +1,44 @@
...
+
+const BillingInvoiceModel = require('../../models/water/billing-invoice.model.js')
+const FetchInvoiceAccountService = require('./fetch-invoice-account.service.js')
+
+/**
+ * Create the initial billing invoice for the provided charge version, billing period and billing batch prior to
+ * transactions being sent to the charging module
+ *
+ * @param {Object[]} chargeVersion An Object containing the relevant charge version
+ * @param {Object[]} billingPeriod Object that has a `startDate` and `endDate` that defines the billing period
```
Cruikshanks, line 17: Just ensuring we are correct in the documentation. Putting brackets after the type infers we're expecting an array

```suggestion
 * @param {Object} chargeVersion An Object containing the relevant charge version
 * @param {Object} billingPeriod Object that has a `startDate` and `endDate` that defines the billing period
```

https://jsdoc.app/tags-param.html

app/services/supplementary-billing/create-billing-invoice.service.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param {string} billingBatchId GUID of the billing batch
+ *
+ * @returns {Object} The newly-created billing invoice record
+ */
+async function go (chargeVersion, billingPeriod, billingBatchId) {
+  const billingInvoice = await BillingInvoiceModel.query()
+    .insert({
+      invoiceAccountId: chargeVersion.invoiceAccountId,
+      address: {}, // Have left empty as doesn't appear to be used for SROC
```
Cruikshanks, line 26: Now we're committing the code as 'production-ready' we can be more deterministic in our comments.

```suggestion
      address: {}, // Address is set to an empty object for SROC billing invoices
```

test/services/supplementary-billing/create-billing-invoice.service.test.js
```diff
@@ -0,0 +1,40 @@
...
+  const billingBatchId = '95177485-bfe7-48ed-b2f0-dba5a5084b45'
+  const billingPeriod = { startDate: new Date('2022-04-01'), endDate: new Date('2023-03-31') }
+
+  beforeEach(async () => {
+    await DatabaseHelper.clean()
+  })
+
+  it('returns the new billing invoice instance containing the correct data', async () => {
+    const invoiceAccount = await InvoiceAccountHelper.add({ invoiceAccountNumber: 'TEST12345A' })
+    const chargeVersion = { invoiceAccountId: invoiceAccount.invoiceAccountId }
```
Cruikshanks, line 28: This is a personal preference thing. I use `beforeEach()` blocks to handle the 'setup' for a test. The aim is that what happens in the test, i.e. the `it` block, is limited to executing the thing and asserting the result or behaviour.

That preference has leached into our unit tests as a defacto convention, which is why I'm flagging this.

So, I'd suggest declaring `invoiceAccount` and `chargeVersion` where you are already declaring your other values and then instantiating them in the `beforeEach()` instead of in the test.

```javascript
describe('Create Billing Invoice service', () => {
  const billingBatchId = '95177485-bfe7-48ed-b2f0-dba5a5084b45'
  const billingPeriod = { startDate: new Date('2022-04-01'), endDate: new Date('2023-03-31') }

  let invoiceAccount
  let chargeVersion

  beforeEach(async () => {
    await DatabaseHelper.clean()

    invoiceAccount = await InvoiceAccountHelper.add({ invoiceAccountNumber: 'TEST12345A' })
    chargeVersion = { invoiceAccountId: invoiceAccount.invoiceAccountId }
  })

  // ...
```

Cruikshanks (commented):
I can only comment as I raised the PR. But I've flagged a few things for consideration. Nothing major, honest! 😁

test/services/supplementary-billing/create-billing-invoice.service.test.js
Jozzey: Sounds sensible, I shall implement your suggestion.

PR #124, opened 2023-02-17
Generate Billing Invoice Licence record for SROC

StuAA78:
https://eaflood.atlassian.net/browse/WATER-3897

In order to 'piggy-back' onto the existing functions in the legacy service for reviewing, confirming, sending, viewing, deleting a bill run etc we need to create the same base database records as it does for a Bill run. As part of this we need to generate`billing_invoice_licence` entities, which link billing invoices with licences. We already have this entity so this PR implements `CreateBillingInvoiceLicenceService` which can create these.

### DIFF
--- a/test/services/supplementary-billing/create-billing-invoice-licence.service.test.js
+++ b/test/services/supplementary-billing/create-billing-invoice-licence.service.test.js
@@ -16,7 +16,7 @@ const DatabaseHelper = require('../../support/helpers/database.helper.js')
 // Thing under test
 const CreateBillingInvoiceLicenceService = require('../../../app/services/supplementary-billing/create-billing-invoice-licence.service.js')
 
-describe.only('Create Billing Invoice Licence service', () => {
+describe('Create Billing Invoice Licence service', () => {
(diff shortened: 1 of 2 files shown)


test/services/supplementary-billing/create-billing-invoice-licence.service.test.js
```diff
@@ -0,0 +1,43 @@
...
+  describe('when a BillingBatchModel instance is provided', () => {
+    let billingInvoice
+    let licence
+
+    beforeEach(async () => {
+      billingInvoice = await BillingBatchHelper.add()
+      licence = await LicenceHelper.add()
+    })
+
+    it('creates an event record', async () => {
```
Cruikshanks, line 33:
```suggestion
    it('returns the new billing invoice licence instance', async () => {
```

Cruikshanks (changes requested):
I think it's just a correction needed from a copy and paste.

Cruikshanks (approved):

Jozzey (approved):

PR #133, opened 2023-02-27
Update Process Batch service for debit only

Cruikshanks:

We have created the individual 'tools' needed to generate an SROC supplementary bill-run. By this, we mean the `billing_batch`, `billing_invoice`, `billing_invoice_licence` and `billing_transaction` records. Also, (though we're still not sure where it's used!) the event record.

We have also worked to identify the charge versions to consider in an SROC supplementary bill-run, and how to communicate with the charging module API. Now it's time to pull it all together and create a complete bill-run, one that can be confirmed and sent in the WRLS service UI.

The work to credit a previous bill run is still to be done. We also need to properly handle when a bill run is empty (no valid transactions were found) or the process errors. But we have everything we need to get us to a place where we folks can start testing and verifying debit-only bill-runs.

So, we've 'spiked' a solution and this change updates `ProcessBillingBatchService` with it. There are no unit tests because we're still trying to decide on our final approach. But this change means testing can begin.


Cruikshanks:
The fact this is a spike and so we have added a bunch of `TODO:` comments to the code is why SonarCloud is complaining. It is right to but we are ignoring it for now.

Jozzey (approved):

PR #139, opened 2023-03-03
Only log essential info on failed requests

Cruikshanks:

When working on the [Spike - Create a complete debit only SROC bill run](https://github.com/DEFRA/water-abstraction-system/pull/133) a request to the [sroc-charging-module-api](https://github.com/DEFRA/sroc-charging-module-api) failed, as in we got a `4xx/5xx` response.

One of the things that made it hard to track down the cause as we are currently dumping the full response object [Got](https://github.com/sindresorhus/got) returns into the log.

Should **Got** throw an exception, for example, when a network error occurs we'll want to see everything. But if the request fails because the Charging Module has returned an error response, we just want to see that information.

So, this tweaks the logging in `RequestLib` to ensure this.

### DIFF
--- a/app/lib/request.lib.js
+++ b/app/lib/request.lib.js
@@ -86,7 +86,7 @@ async function _importGot () {
  * we need to diagnose the problem.
  *
  * @param {string} method the type of request made, for example, 'GET', 'POST, or 'PATCH'
- * @param {Object} result the result we object generate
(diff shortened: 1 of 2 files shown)


Jozzey (changes requested):
Just 1 typo then all good

Jozzey (approved):
Approved, just the 1 typo

PR #150, opened 2023-03-06
Move determineChargePeriod() to a service

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3930

The charge period is the period of time that the charge version and its start and end date overlap with the financial period being billed (01-Apr to 31-Mar). We need this period to then calculate how many billable days there are for the charge version.

Currently, the `FormatSrocTransactionLineService` has this function. But we also need to know the charge period to [determine whether minimum charge applies](https://github.com/DEFRA/water-abstraction-system/pull/149) to the charge version.

So, this change moves the function out to a service we can then reuse in both places.


app/services/supplementary-billing/determine-minimum-charge.service.js
```diff
@@ -0,0 +1,40 @@
+'use strict'
```
StuAA78, line 1: Was this file meant to be included?

StuAA78 (changes requested):

StuAA78 (dismissed):

Jozzey (dismissed):

Jozzey (approved):

PR #152, opened 2023-03-06
Add Billing Batch Status update service

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3933

When we start processing the SROC supplementary billing batch we need to update its status. This is used by the [water-abstraction-ui](https://github.com/DEFRA/water-abstraction-ui) to indicate to the user at what stage the bill run is. Should the bill run have no transactions, we'll also need to update the status then.

So, this change adds a new `UpdateBillingBatchStatusService` that will update a batch with whatever status we give it.


Cruikshanks:
Decided against this. It adds nothing to just calling Objection directly.

PR #153, opened 2023-03-06
Handle empty supplementary bill runs

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3923

From the charge version, we generate transaction lines. For each charge element (reference) we look at the charge purposes connected to it and calculate the billable days. If that result is 0, we skip adding the transaction to the bill run.

This means there is often a case where having calculated and rejected all the transaction lines, the bill run is 'empty'. The legacy process will update the status of the bill run to `empty`, and remove any `billing_invoice` and `billing_invoice_licence` records created.

We need to do the same, though we intend to take a different approach. We'll process all the transaction lines first and only create `billing_invoice` and `billing_invoice_licence` records for those that remain. Should there be none, we'll go straight to marking the bill run as `empty` and stop the process there.

Our intent is to remove needless calls to the DB and simplify the process.


Cruikshanks:
The issues with SonarCloud are to do with code coverage on the `ProcessBillingBatchService`. We know we have no unit tests for this service. But we're continuing to 'spike' the process and will ensure there are unit tests when we're happy with what the final process will be.

PR #155, opened 2023-03-07
Fix dupe bill. invoice licence in process service

Cruikshanks:

Found in testing where we have a licence with multiple charge versions for the same invoice account we are handling reusing the billing invoice. But we are always trying to create a new billing invoice licence when we should be reusing them (when the Billing Invoice ID and licence ID are the same).

So, this change updates the `app/services/supplementary-billing/create-billing-invoice-licence.service.js` to work in the same way as `app/services/supplementary-billing/create-billing-invoice.service.js`, fetching the existing billing invoice licence record if it exists else creating a new one.


Jozzey (approved):
👍🏼

StuAA78 (approved):

PR #156, opened 2023-03-07
Fix error calculating auth. and billable days

Cruikshanks:

In the first round of testing, we came across a scenario where our `FormatSrocTransactionLineService` service calculated the `authorisedDays` and `billableDays` as `null`.

After some investigation we realised

- we had 2 ways to calculate billable days
- both had a fundamental flaw

`AbstractionBillingPeriodService` would take a billing period and a charge purpose and return 1 or 2 abstraction periods with the billable days calculated. But it didn't take into account any overlap between the abstraction periods it determined. It also didn't account for any overlap across _all_ the charge purposes a charge element has.

`FormatSrocTransactionLineService` was doing its own logic for determining the abstraction period. It was also looking across the charge purposes and dealing with any overlap. But when determining the abstraction period it hadn't accounted for something like **01-JAN** to **30-JUN**. A period like that needs to be factored in twice, especially when determining the authorised days. This is because if, for example, the billing period is **01-APR-2023** to **31-MAR-2023** it overlaps the first few months and the last few months.

We realised we need to combine the logic of `AbstractionBillingPeriodService` with `FormatSrocTransactionLineService` use of `ConsolidateDateRangesService`.

So, this change adds `CalculateAuthorisedAndBillableDaysService`, which completely replaces `AbstractionBillingPeriodService` and the logic in `FormatSrocTransactionLineService`.

This is _the_ most complex bit of the whole process (that we've encountered) and we're hopeful we now have it nailed. This change also adds lots of documentation to try and help our future selves! 😀


test/services/supplementary-billing/calculate-authorised-and-billable-days.service.test.js
```diff
@@ -0,0 +1,401 @@
...
+
+      describe('and the charge period is 01-AUG-2023 to 30-SEP-2023', () => {
+        beforeEach(() => {
+          chargePeriod = {
+            startDate: new Date('2023-08-01'),
+            endDate: new Date('2023-09-30')
+          }
+        })
+
+        it('returns 182 for authorised days and 59 for billable days', () => {
```
Jozzey, line 234:
```suggestion
        it('returns 182 for authorised days and 0 for billable days', () => {
```

Jozzey (dismissed):
👍🏼  Just one typo

PR #157, opened 2023-03-09
Refactor how we create transactions

Cruikshanks:

Having implemented our [Fix error calculating auth. and billable days](https://github.com/DEFRA/water-abstraction-system/pull/156) we clocked that we were calculating charge periods, option flags, and the authorised and billable days multiple times, when they do not change between iterations of the charge elements in a charge version.

We only want to be executing our services when needed as it reduces complexity and the chance of an error occurring. It also means we get a slight performance bump.

So, this change refactors how the `ProcessBillingBatchService` and the `FormatSrocTransactionLineService` interact together to create the transaction lines for a billing batch.


Jozzey (approved):
👍🏼

StuAA78 (approved):

PR #161, opened 2023-03-13
Handle errors in ProcessBillingBatchService

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3924

Should an error occur during the supplementary bill run process it brings down the service. We're not handling the exception and because we moved the process to the background in our bill runs controller, there is nothing to deal with it when thrown.

So, we need to protect the service from _any_ errors during the process, to prevent the service from going down. But we also need to mimic the legacy service and use an 'error code' to provide extra information about the error. We've already used this concept with [Create error bill run when CM fails](https://github.com/DEFRA/water-abstraction-system/pull/104) where we apply a code to the `billing_batch` record which the UI then uses to indicate what happened to the user.

This change deals with both; ensuring we handle any exceptions raised and where possible, setting a relevant error code when updating the `billing_batch` status to 'errored'.

### DIFF
--- a/app/services/supplementary-billing/handle-errored-billing-batch.service.js
+++ b/app/services/supplementary-billing/handle-errored-billing-batch.service.js
@@ -13,20 +13,27 @@ const BillingBatchModel = require('../../models/water/billing-batch.model.js')
  * We keep this in a separate service so we don't need to worry about multiple/nested try-catch blocks in cases where a
  * billing batch fails and setting its status to error also fails.
  *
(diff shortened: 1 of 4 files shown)


StuAA78 (approved):

Cruikshanks:
The SonarCloud fail is now just code coverage, and that is due to our decision to hold off writing unit tests until the `ProcessBillingService` stabilises. This has been proven by the fact we've made large-scale changes to it in this PR, which is why SonarCloud is back with a vengeance!

So, we're ignoring it for now but do intend to write full unit tests in the near future.

PR #162, opened 2023-03-13
Speed up `RequestLib` tests

StuAA78:
https://github.com/DEFRA/water-abstraction-team/issues/98

For those paying attention when running our unit tests, they appear to stall when it comes to `test/lib/request.lib.test.js`. This is because we have tests that ensure the module behaves as expected when requests to other APIs time out or fail.

One of the reasons we love [Got](https://github.com/sindresorhus/got) is because retries are built-in and easy to configure. But that means we have to allow Got to retry failures when running the unit tests.

The reason for the stalling is because the time between retries is based on a 'computed value';

```javascript
((2 ** (attemptCount - 1)) * 1000) + noise
```

This means the delay increases exponentially. That's great for what we do in the app. A 'blip' when talking to an external API is smoothed over and all the user sees is something takes a few seconds longer than normal. But in our unit tests it causes some to appear to stall.

Thankfully, the brilliant @StuAA78 spotted that we can override the computed value by setting the [backoffLimit](https://github.com/sindresorhus/got/blob/main/documentation/7-retry.md#backofflimit).

> The upper limit of the `computedValue`.

If we set this when running the unit tests the time between retries will be limited to 50ms.

So, this change updates `RequestLib` to make it easy to access the default options it uses so we can then apply `backofflimit` to the retry options.

---

## Before changes

- Test duration: 40059 ms
- Test duration: 40057 ms
- Test duration: 40116 ms

Average **40077 ms** (40 secs)

## After changes

- Test duration: 16960 ms
- Test duration: 16948 ms
- Test duration: 17009 ms

Average **16972 ms** (16 secs)

### DIFF
--- a/app/lib/request.lib.js
+++ b/app/lib/request.lib.js
@@ -10,6 +10,61 @@ const { HttpsProxyAgent } = require('hpagent')
 const requestConfig = require('../../config/request.config.js')
 
 /**
+ * Returns an object containing the default options.
+ *
+ * We want to be able to export these options so we can set specific settings that may be multiple objects deep. For
+ * example, if we want to set `retry.backoffLimit` then we can't simply pass `{ retry: { backoffLimit: 5 } }` to the lib
+ * as `additionalOptions` as this would replace the entire `retry` section. We would therefore want to get the exported
+ * `retry` section and add our setting to it before passing it back as `additionalOptions`.
+ *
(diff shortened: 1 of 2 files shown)


Cruikshanks:
If it helps, comparing the unit test times on this branch compared to others it looks like this change will cut 20ish secs of build time 🤷🎉

robertparkinson (approved):

PR #163, opened 2023-03-14
Log time taken to process a bill run

Cruikshanks:

When issues of performance came up in the [charging-module-api](https://github.com/DEFRA/sroc-charging-module-api) we found it immensely useful to have evidence to hand of how long the most contentious process was actually taking.

As we are not in control of the full process with our new SROC supplementary billing process, it would be useful to know exactly how long are part of the journey is taking. We're also about to expand the process to include reversing the previous supplementary bill run. So, it would be useful to get a working idea of how that impacts our timings.

To be clear, this is a very basic metric; the difference between the times at the start and the end of the process. There is no call at this time for factoring the size of a bill run, load on the servers at the time etc. It's all we needed in the Charging Module API, which also means it's some code we can readily steal!

<details>
<summary>Screenshot of log output</summary>

Screenshot 2023-03-14 at 09 34 06

</details>

### DIFF
--- a/app/services/supplementary-billing/process-billing-batch.service.js
+++ b/app/services/supplementary-billing/process-billing-batch.service.js
@@ -36,6 +36,9 @@ async function go (billingBatch, billingPeriod) {
   const { billingBatchId } = billingBatch
 
   try {
+    // Mark the start time for later logging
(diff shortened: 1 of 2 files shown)


Jozzey (approved):
👍🏼

PR #166, opened 2023-03-14
Fix cached data in ProcessBillingBatchService

Cruikshanks:

As part of implementing [Handle errors in ProcessBillingBatchService](https://github.com/DEFRA/water-abstraction-system/pull/161) we refactored the `ProcessBillingBatchService` to make the code cleaner and easier to read.

One of the changes was moving the `generatedInvoices` and `generatedInvoiceLicences` to the module scope rather than just in the method scope. It saved us from having to pass them around the place which removed a bunch of 'cruft'. What we didn't take into account is they would be cached between calls to the service.

Now it's at module scope the data from previous runs is being cached and still exists when a new one starts. So, we're getting to the finalise stage and we're trying to persist old data from a previous run!

This change resolves the issue.

---

For reference, we move the variables into the scope of the `go()` function removing the issue of the data being cached at the global scope.

We then take a halfway house approach; we pass the information into our methods but we accept having to mutate the data rather than pass back a modified copy.

<details>
<summary>Screenshot of logs showing caching</summary>

Screenshot 2023-03-14 at 10 30 20

</details>

### DIFF
--- a/app/services/supplementary-billing/process-billing-batch.service.js
+++ b/app/services/supplementary-billing/process-billing-batch.service.js
@@ -21,9 +21,6 @@ const GenerateBillingInvoiceLicenceService = require('./generate-billing-invoice
 const HandleErroredBillingBatchService = require('./handle-errored-billing-batch.service.js')
 const LegacyRequestLib = require('../../lib/legacy-request.lib.js')
 
-let generatedInvoices = []
-let generatedInvoiceLicences = []
-
 /**
  * Creates the invoices and transactions in both WRLS and the Charging Module API
  *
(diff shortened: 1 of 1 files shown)


Cruikshanks:
The SonarCloud fail is now just code coverage, and that is due to our decision to hold off writing unit tests until the `ProcessBillingService` stabilises. This has been proven by the fact we've made large-scale changes to it in this PR, which is why SonarCloud is back with a vengeance!

So, we're ignoring it for now but do intend to write full unit tests in the near future.

StuAA78 (approved):

PR #174, opened 2023-03-24
Handle no charge versions in ProcessBillingBatch

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3952
https://eaflood.atlassian.net/browse/WATER-3803

To be honest, this change started as our first effort to add unit test coverage for the `ProcessBillingBatchService`. So, for example, our unit test error coverage now covers the changes we made including reversing the previous billing batch.

Whilst working on them we realised we were not handling no charge versions being returned at all. So, we put in protection against this issue. Nice!

Only our QA has come across this issue in testing and needs it resolved so they can continue their work.

Hence, we're derailing the original intent of this PR and getting the fix to handle no charge versions returned (plus tests! 😬)


Jozzey (approved):
👍🏼

PR #177, opened 2023-03-27
Update FetchChargeVersions to use SROC flag

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3948

We have an issue that the current `include_in_supplementary_billing` flag was added at a time there was only 1 charge scheme. A licence gets flagged irrespective of whether the change relates to PRESROC or SROC.

Where that has impacted us is when sending a billing batch. When it gets sent it clears the flag for _all_ licences included. But if we cancel one, for example, the PRESROC bill run, and send the SROC one we lose which licences still need to be processed on the PRESROC one.

[Our solution](https://github.com/DEFRA/water-abstraction-service/pull/2077) is to add a new `include_in_sroc_supplementary_billing` flag to the `licences` table and only set it to 'true' when an SROC charge version is approved. With this in place our `FetchChargeVersionsService` needs to be altered to look at that field instead of the old one to identify charge versions to consider for billing.


Cruikshanks:
https://github.com/DEFRA/water-abstraction-service/pull/2077 needs to be merged before this can be merged.

PR #179, opened 2023-03-27
Handle REPLACED charge v's with different inv. acc

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3950

During the supplementary billing process as we iterate through the charge versions we do so in invoice account and then licence order. This is so we can use 'iterating to the next one' as a trigger to know when to persist the current billing invoice and billing licence records we're dealing with.

Because we have the invoice account and licence ID's during this process it made it simple to look at the last bill run and extract any matching debit billing transactions that might now need crediting.

But during test, a scenario came up that we're not managing. A charge version can be created that supersedes the previous one. It supersedes the previous because the start date of the new charge version is the same as the previous i.e. the previous charge version should be completely ignored.

This most often happens because the billing account was wrong. So, the new charge version will be linked to a different billing account. The problem is when the previous charge version has already been billed. We need to include a credit in the new bill run we're generating for it. But our 'prudent' inclusion of the invoice account ID in our query means we're overlooking the previous debit to the old billing invoice account.

The end result is 2 invoices to 2 billing accounts on different bill runs, for the same licence and abstraction period.

This change resolves the problem, though goodness knows how!

### DIFF
--- a/app/services/supplementary-billing/fetch-charge-versions.service.js
+++ b/app/services/supplementary-billing/fetch-charge-versions.service.js
@@ -38,7 +38,7 @@ async function _fetch (regionId, billingPeriod) {
     ])
     .innerJoinRelated('licence')
     .where('scheme', 'sroc')
-    .where('includeInSrocSupplementaryBilling', 'yes')
+    .where('includeInSrocSupplementaryBilling', true)
     .where('regionId', regionId)
     .where('chargeVersions.status', 'current')
     .where('chargeVersions.startDate', '>=', billingPeriod.startDate)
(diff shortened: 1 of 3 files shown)


Cruikshanks:
@Jozzey @StuAA78 I meant to just push the notes we had but instead started to have a play 😁. With my changes, this is the kind of scenario I have seen it working locally for

Screenshot 2023-04-10 at 17 20 26

It was correctly adding the credits which it wasn't doing before. There is still work to be done to complete the PR. But hopefully a working PR makes more sense than some rough notes 🤷

StuAA78 (approved):

Jozzey (approved):

PR #189, opened 2023-04-12
Export billing_charge_categories

Beckyrose200:
https://eaflood.atlassian.net/jira/software/c/projects/WATER/boards/96?modal=detail&selectedIssue=WATER-3963

This pull request is just a small part of a larger project that involves exporting all our database schemas, converting them into CSV files, and uploading them to our Amazon S3 bucket. Its primary focus is establishing a connection to the database and querying a specific table from the water schema. To expedite the export process and see the output sooner, we are using a vertical slicing approach, rather than a horizontal one, which means exporting a single table at a time from each schema.

### DIFF
--- a/test/services/db-export/fetch-billing-charge-categories-table.service.test.js
+++ b/test/services/db-export/fetch-billing-charge-categories-table.service.test.js
@@ -24,7 +24,7 @@ describe('Billing charge categories table export service', () => {
     await BillingChargeCategoryHelper.add()
   })
 
-  describe.only('when we connect to the db', () => {
+  describe('when we connect to the db', () => {
     it('returns all records in the billing-charge-categories table', async () => {
       const results = await BillingChargeCategoriesTableExportService.go()
 


app/services/db-export/billing-charge-categories-table-export.service.js
Cruikshanks, line 1: I'd suggest renaming the file and service to follow a convention we seemed to have settled on. Where a 'service' is responsible for just returning data from the DB we name it `fetch-foo-bar.service.js`. Take a look in `app/services/supplementary-billing` for current examples.

So, this would become `fetch-billing-charge-categories.service.js`.

app/services/db-export/billing-charge-categories-table-export.service.js
```diff
@@ -0,0 +1,27 @@
+'use strict'
+
+/**
+ * Used by the 'root.controller' currently to determine if we are pulling
+ * the billing_charge_categories table correctly
```
Cruikshanks, line 5: We don't tend to worry about stating where something is used, especially when that usage is temporary as in this case.

I'd suggest

```suggestion
 * Fetches all records from the water.billing_charge_categories table
```

Cruikshanks (changes requested):
The code is fine. Just some conventions you have not been exposed to suggesting some name changes.

Jozzey (dismissed):
With the exception of the issues @Cruikshanks has found it looks good. 👍🏼

Cruikshanks (approved):

StuAA78 (approved):

PR #190, opened 2023-04-13
Standardise how we test the length of an array

Jozzey:
In our unit tests we had been checking the length of an array using the method `expect(result.length).to.equal(1)`. However our preferred way to test the length of an array is `expect(result).to.have.length(1)`.

This PR updates all unit tests to use the `expect(result).to.have.length(1)` method to test the length of an array.

### DIFF
--- a/test/services/supplementary-billing/fetch-charge-versions.service.test.js
+++ b/test/services/supplementary-billing/fetch-charge-versions.service.test.js
@@ -86,7 +86,7 @@ describe('Fetch Charge Versions service', () => {
     it("returns only the 'current' SROC charge versions that are applicable", async () => {
       const result = await FetchChargeVersionsService.go(regionId, billingPeriod)
 
-      expect(result.length).to.equal(1)
+      expect(result).to.have.length(1)
       expect(result[0].chargeVersionId).to.equal(testRecords[0].chargeVersionId)
     })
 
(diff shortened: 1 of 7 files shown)


StuAA78 (approved):

Beckyrose200 (approved):

test/services/supplementary-billing/fetch-charge-versions.service.test.js
```diff
@@ -154,7 +154,7 @@ describe('Fetch Charge Versions service', () => {
       it('returns no applicable charge versions', async () => {
         const result = await FetchChargeVersionsService.go(regionId, billingPeriod)
 
-        expect(result.length).to.equal(0)
+        expect(result).to.have.length(0)
```
StuAA78, line 157: Just a thought, do you think it would maybe read better to use `.to.be.empty()` in place of `.to.have.length(0)` for this and the other tests like it?

StuAA78 (changes requested):
Small change to the tests where we're checking `.to.have.length(0)` -- might read better to use `.to.be.empty()` instead?

test/services/supplementary-billing/fetch-charge-versions.service.test.js
Jozzey: Yes I think it would read better like that. I'll make the changes.

Didn't think you'd be able to find anything on this one :grinning:

Jozzey: I've implemented your suggestion

StuAA78 (approved):
👍

PR #194, opened 2023-04-19
Fix supplementary bill run errors with non-chargeable charge versions

StuAA78:
https://eaflood.atlassian.net/browse/WATER-3972

We found that an sroc supplementary bill run that contains a non-chargeable charge version would immediately error. On investigation we found that we were attempting to create invoices for these charge versions, starting by destructuring the `invoiceAccountId`, causing an error.

Since we do not want to invoice these charge versions at all, we fix this by updating our db query to exclude records where `invoiceAccountId` is `null`.


Cruikshanks (approved):

Jozzey (approved):
👍🏼

PR #202, opened 2023-04-26
Reliable test data clear down OLD

Jozzey:
https://eaflood.atlassian.net/browse/WATER-3982

Using the existing acceptance tests a common issue we face is that we can’t clear down the existing test data. Because of the solution previously implemented, it only focuses on data flagged as `is_test === true`. The problem is when using this data for test purposes you will often create related data that is not flagged as `is_test`. This leads to errors in the existing clear down process that we then have to create a and run manual SQL delete scripts to deal with.

If we’re often having to write SQL delete scripts manually, and these deal with the foreign key constraints that the existing process isn’t handling, it makes sense to us to use them always rather than the existing brittle process.


Jozzey:
Something went wrong with the branch and I couldn't fix it so I have created a new one and switched to this PR https://github.com/DEFRA/water-abstraction-system/pull/215

PR #205, opened 2023-04-28
Compress billing-charge-categories export

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-3965

This pull request is just a small part of a larger project that involves exporting all our database schemas, converting them into CSV files, and uploading them to our Amazon S3 bucket. This PR's primary focus is compressing the billing charge categories table that we have already converted and exported to a CSV file. To expedite the export process and see the output sooner, we are using a vertical slicing approach, rather than a horizontal one, which means exporting a single table at a time from each schema.

### DIFF
--- a/app/services/db-export/export-data-files.service.js
+++ b/app/services/db-export/export-data-files.service.js
@@ -19,11 +19,13 @@ const os = require('os')
  * @returns {Boolean} True if the file is written successfully and false if not
  */
 async function go (tableConvertedToCsv, tableName) {
+  const filePath = _filenameWithPath(tableName)
+
   try {
-    await fs.writeFile(_filenameWithPath(tableName), tableConvertedToCsv)
+    await fs.writeFile(filePath, tableConvertedToCsv)
     global.GlobalNotifier.omg(`${tableName} exported successfully`)
 
-    return _filenameWithPath(tableName)
+    return filePath
   } catch (error) {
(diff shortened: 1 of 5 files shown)


app/services/db-export/export-data-files.service.js
```diff
@@ -23,7 +23,7 @@ async function go (tableConvertedToCsv, tableName) {
     await fs.writeFile(_filenameWithPath(tableName), tableConvertedToCsv)
     global.GlobalNotifier.omg(`${tableName} exported successfully`)
 
-    return true
+    return _filenameWithPath(tableName)
```
Cruikshanks, line 26: We are calling `_filenameWithPath()` twice and expecting to see the same value. There is also 'work' going on within that method that goes beyond adding some numbers or manipulating a string.

So, it would be better to put the result of `_filenameWithPath()` into a const first. Then pass it to `fs.writeFile()` and also return it as our service's result.

Cruikshanks (changes requested):
Just have a few comments on this one. Looks good though!

Jozzey (approved):

Cruikshanks (dismissed):

StuAA78 (approved):

Cruikshanks (approved):

Jozzey (approved):

PR #207, opened 2023-04-28
Inc. changes to agreements & charges in supp bill

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3989

When a new charge version is added to a licence, users can change agreements and charges linked to the licence via the charge version.

So, for example, when the SROC charge version is first applied the licence does **not** have a two-part tariff agreement (2PT). But later in the year, one is added. To do this the business will add a new charge version and apply the 2PT agreement.

Though the number of billable days calculated won't change, the charge that the [charging-module-api](https://gothub.com/DEFRA/sroc-charging-module-api) returns will. So, we need to credit the previous transaction and calculate a new debit. At the moment, our engine does not look at changes to agreements and charges. So, it will compare the billable days, see no change and cancel both the previous and calculated transaction lines before we get a value from the charging module.

This change updates the logic for comparing previous transactions to those we've calculated to include agreements and charges. If there has been a change, the lines won't match and get cancelled so will then progress to the next step; sending them to the charging module to calculate their charge value.

### DIFF
--- a/app/services/supplementary-billing/process-billing-transactions.service.js
+++ b/app/services/supplementary-billing/process-billing-transactions.service.js
@@ -39,21 +39,55 @@ async function go (calculatedTransactions, billingInvoice, billingInvoiceLicence
 }
 
 /**
+ * Compares a calculated transaction to the transactions from previous bill runs
+ *
  * Takes a single calculated debit transaction and checks to see if the provided array of reversed (credit) transactions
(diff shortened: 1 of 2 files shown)


Cruikshanks:
Unfortunately, a relatively small code change has resulted in a number of tests being added. Hence the size of the PR. Sorry folks! 😳😬

Jozzey (approved):

PR #211, opened 2023-05-04
Uploading file to AWS S3 bucket

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-3966

This pull request is just a small part of a larger project that involves exporting all our database schemas, converting them into CSV files, and uploading them to our Amazon S3 bucket. This PR's primary focus is connecting to the amazon s3 bucket and uploading the file. To expedite the export process and see the output sooner, we are using a vertical slicing approach, rather than a horizontal one, which means exporting a single table at a time from each schema.


app/services/db-export/send-to-s3-bucket.service.js
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Sends a file to our AWS S3 bucket
+ * @module SendToS3BucketService
+ */
+
+const { S3Client, PutObjectCommand } = require('@aws-sdk/client-s3')
+const fs = require('fs')
+const S3Config = require('../../../config/s3.config')
+const path = require('path')
```
Cruikshanks, line 11: Only a minor thing but this doesn't quite meet our [team convention](https://github.com/DEFRA/water-abstraction-team/blob/main/coding_conventions.md#top-of-js-files)
```suggestion
const fs = require('fs')
const path = require('path')
const { PutObjectCommand, S3Client } = require('@aws-sdk/client-s3')

const S3Config = require('../../../config/s3.config')
```

app/services/db-export/send-to-s3-bucket.service.js
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @param {String} filePath A string containing the path of the file to send to the S3 bucket
+ *
+ * @returns {Boolean} True if the file is uploaded successfully and false if not
+ */
+async function go (filePath) {
+  if (!filePath) {
+    global.GlobalNotifier.omfg('ERROR uploading file to S3 bucket. No file path given')
+    return false
+  }
```
Cruikshanks, line 24: Typically a service like this won't be used in isolation. It will be called by a marshalling service or function, that is taking the result of one thing and passing it to this.

When this is the case we let that 'thing' handle situations where something is unexpected. It's fine to code the `SendToS3BucketService` with the assumption that it will always be given a path, and should it not or that path be invalid then things will go 💥 !

```suggestion
```

app/services/db-export/send-to-s3-bucket.service.js
```diff
@@ -0,0 +1,63 @@
...
+  const bucketName = S3Config.s3.bucket
+  const folderName = 'export'
+  const fileName = path.basename(filePath)
+  const fileContent = fs.readFileSync(filePath)
+
+  const params = {
+    Bucket: bucketName,
+    Key: `${folderName}/${fileName}`,
+    Body: fileContent
+  }
```
Cruikshanks, line 35: I think we only use `folderName` when generating the `Key:` value. If so then there is no harm in just using it directly.

```suggestion
  const fileName = path.basename(filePath)
  const fileContent = fs.readFileSync(filePath)

  const params = {
    Bucket: bucketName,
    Key: `export/${fileName}`,
    Body: fileContent
  }
```

test/services/db-export/send-to-s3-bucket.service.test.js
```diff
@@ -0,0 +1,94 @@
...
+      const calledCommand = s3Stub.getCall(0).firstArg
+      expect(calledCommand instanceof PutObjectCommand).to.be.true()
+
+      expect(result).to.equal(true)
+      expect(notifierStub.omg.calledWith(`The file ${fileName} was uploaded successfully`)).to.be.true()
+    })
+  })
+
+  describe('when unsuccessful', () => {
+    describe('an invalid file name is given', () => {
```
Cruikshanks, line 59:
```suggestion
    describe('because an invalid file name is given', () => {
```

test/services/db-export/send-to-s3-bucket.service.test.js
```diff
@@ -0,0 +1,94 @@
...
+      })
+
+      it('logs the error', async () => {
+        await SendToS3BucketService.go()
+
+        expect(notifierStub.omfg.calledWith('ERROR uploading file to S3 bucket. No file path given')).to.be.true()
+      })
+    })
+
+    describe('there is an issue with the upload', () => {
```
Cruikshanks, line 73:
```suggestion
    describe('because there is an issue with the upload', () => {
```

Cruikshanks (changes requested):
I think it just needs a few tweaks. Shout if you need to chat through any of them!

test/services/db-export/send-to-s3-bucket.service.test.js
```diff
@@ -0,0 +1,94 @@
...
+
+      // Test that the S3 Client was called once
+      expect(s3Stub.calledOnce).to.be.true()
+
+      // Get the first call and test that it was called with PutObjectCommand
+      const calledCommand = s3Stub.getCall(0).firstArg
+      expect(calledCommand instanceof PutObjectCommand).to.be.true()
+
+      expect(result).to.equal(true)
+      expect(notifierStub.omg.calledWith(`The file ${fileName} was uploaded successfully`)).to.be.true()
```
StuAA78, line 54: These could be tested in separate `it` blocks eg.:
```js
    it('returns true', async () => {
      ...
    })

    it('logs a success message', async () => {
      ...
    })

File: test/services/db-export/send-to-s3-bucket.service.test.js [hunk 7]
```diff
@@ -0,0 +1,94 @@
...
+
+    it('uploads a file to the S3 bucket', async () => {
+      const result = await SendToS3BucketService.go(filePath)
+
+      // Test that the S3 Client was called once
+      expect(s3Stub.calledOnce).to.be.true()
+
+      // Get the first call and test that it was called with PutObjectCommand
+      const calledCommand = s3Stub.getCall(0).firstArg
+      expect(calledCommand instanceof PutObjectCommand).to.be.true()
```
StuAA78, line 51: There's an [`instanceof()` assertion](https://hapi.dev/module/code/api/?v=9.0.3#instanceoftype) we can use here:
```suggestion
      expect(calledCommand).to.be.an.instanceof(PutObjectCommand)
```

StuAA78 (changes requested):
Looks good! Just a couple of suggested tweaks to the unit tests.

Jozzey (commented):
Looking good. I see @Cruikshanks has a few things for you to look at. I can't see anything other than what he has found

Cruikshanks (approved):

StuAA78 (approved):

PR #214, opened 2023-05-09
Refactor `ProcessBillingBatchService` - first pass

StuAA78:
https://eaflood.atlassian.net/browse/WATER-3996

While reviewing `ProcessBillingBatchService` we realised there is scope to make it more performant. This is a first pass which tackles the issue of us fetching invoice accounts one at a time within the main loop. We refactor so that we now fetch all the invoice accounts we will need in a single query prior to entering the main loop and hold them in a keyed object:
```js
{
  '610ca07a-3bd3-4ea2-8cb1-82f6977cdc71': { invoiceAccountId: '610ca07a-3bd3-4ea2-8cb1-82f6977cdc71', ... },
  '03cec07f-9ad3-4c9a-bad8-1958cec0d61f': { invoiceAccountId: '03cec07f-9ad3-4c9a-bad8-1958cec0d61f', ... }
}
```
This means that we can easily retrieve a given invoice account once we're in the main loop:
```js
const invoiceAccountId = invoiceAccounts['610ca07a-3bd3-4ea2-8cb1-82f6977cdc71']
```


Cruikshanks (approved):

test/services/supplementary-billing/fetch-invoice-account-numbers.service.test.js
```diff
@@ -0,0 +1,55 @@
...
+  })
+
+  describe('when the service is called with an array of charge versions', () => {
+    let expectedResult
+    let invoiceAccounts
+
+    beforeEach(async () => {
+      // We create three invoice accounts but we will only be fetching the first two
+      invoiceAccounts = await Promise.all([
+        await InvoiceAccountHelper.add(),
```
Jozzey, line 29: Don't think we need to await the `InvoiceAccountHelper` if using the `Promise.all`

Jozzey (changes requested):

Jozzey (approved):

PR #215, opened 2023-05-09
Reliable test data clear down

Jozzey:
https://eaflood.atlassian.net/browse/WATER-3982

Using the existing acceptance tests a common issue we face is that we can’t clear down the existing test data. Because of the solution previously implemented, it only focuses on data flagged as `is_test === true`. The problem is when using this data for test purposes you will often create related data that is not flagged as `is_test`. This leads to errors in the existing clear down process that we then have to create a and run manual SQL delete scripts to deal with.

If we’re often having to write SQL delete scripts manually, and these deal with the foreign key constraints that the existing process isn’t handling, it makes sense to us to use them always rather than the existing brittle process.

This PR creates an endpoint at `/data/tear-down`. To run locally, send a `POST` request to `http://localhost:8013/data/tear-down`.


Jozzey:
I've had a word with @Cruikshanks about the test coverage on the schema services. As they are just for deleting test data we've agreed that they are not worth spending time on full unit testing since this would involve significant work in creating the new schemas/data in the test DB.

Testing has been added but just at the high level tear-down service and the controller.

app/controllers/data/data.controller.js
```diff
@@ -0,0 +1,23 @@
+'use strict'
+
+/**
+ * Controller for /data endpoints
+ * @module DataController
+ */
+
+const Boom = require('@hapi/boom')
```
Cruikshanks, line 8: Tiny thing! We've got [a convention](https://github.com/DEFRA/water-abstraction-team/blob/main/coding_conventions.md#top-of-js-files) that splits external modules from internal ones with a new line.

```suggestion
const Boom = require('@hapi/boom')

```

app/routes/data.routes.js
```diff
@@ -0,0 +1,17 @@
...
+
+const DataController = require('../controllers/data/data.controller.js')
+
+const routes = [
+  {
+    method: 'POST',
+    path: '/data/tear-down',
+    handler: DataController.tearDown,
+    options: {
+      description: 'Used to remove the integration test data from the database',
```
Cruikshanks, line 11: Potato / potato I know. But we name our UI based test repos 'acceptance tests'. So, that's the name I would use here as well.

 ```suggestion
      description: 'Used to remove the acceptance test data from the database',
```

File: test/controllers/data/data.controller.test.js [hunk 3]
```diff
@@ -0,0 +1,62 @@
...
+    server = await init()
+    Sinon.stub(server.logger, 'error')
+  })
+
+  afterEach(() => {
+    Sinon.restore()
+  })
+
+  describe('POST /data/tear-down', () => {
+    describe('when a request is valid', () => {
```
Cruikshanks, line 36: Perhaps ...

```suggestion
    describe('when the request succeeds', () => {
```

Cruikshanks (changes requested):
Just some minor changes! I can't wait to give it a spin when done 😁

app/plugins/router.plugin.js
```diff
@@ -13,17 +13,19 @@
...
 const AirbrakeConfig = require('../../config/airbrake.config.js')
 
 const routes = [
-  ...RootRoutes,
   ...AssetRoutes,
+  ...BillRunRoutes,
+  ...DataRoutes,
   ...HealthRoutes,
-  ...BillRunRoutes
+  ...RootRoutes
```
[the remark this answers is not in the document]
Jozzey: I did wonder why they were in that order. I'll put it back to how it was.

Cruikshanks (approved):

PR #221, opened 2023-05-12
Unflag all processed licences not in bill run

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4007

> This came about dues to an issue found during UAT of SROC supplementary billing

The issue was that when the SROC bill run was sent licences that were processed in the bill run retained a 'include in SROC supplementary billing' flag. This means the licence will be processed again in the next supplementary bill run.

We've had trouble replicating the issue. So, we went back to the logic to see if that can shine any light.

***The current logic***

At the very end of the **send** process the legacy code will fire off this query

```sql
update water.licences l
  set include_in_sroc_supplementary_billing = :to
  where l.date_updated <= :batchCreatedDate
    and l.region_id = :regionId
    and l.licence_id not in (select licence_id from water.charge_version_workflows where date_deleted is null)
    and l.include_in_sroc_supplementary_billing = :from
```

This is the original query used for PRESROC which we copied and tweaked for SROC. Most folks on the team assumed we unflagged only those licences included in the bill run. But that's not the case.

The problem is a licence when processed may result in no need for a `billing_invoice` to be generated. Because of the queueing architecture in the legacy code at the point this query is run, you've lost which licences never made it into the billing_batch. This is why the legacy code has defaulted to unsetting the flag for any licence last updated prior to the billing_batch (bill run) being created.

***The problem***

Should a licence record get updated after the bill run is created, the query won't include it. We've not been able to pin down how this happens. But our primary suspect is the import process, which seems to update all licence records at approximately the same time.

***The fix***

We're updating our system repo's billing engine to unflag any licences that have been processed but didn't result in a `billing_invoice` record being created. Should the bill run get cancelled and re-generated it won't matter these are no longer flagged as the result will be the same.

With this in place, we can then update the query we're asking the legacy code to fire to be much more intelligent. It can use the `billing_invoice_licence` records linked to the bill run to identify precisely which licences were processed and part of the bill run. It won't matter when they got updated.

With both steps in place, the new SROC supplementary process will only unflag those licences processed during the bill run.

### DIFF
--- a/app/services/supplementary-billing/unflag-unbilled-licences.service.js
+++ b/app/services/supplementary-billing/unflag-unbilled-licences.service.js
@@ -1,0 +1,59 @@
+'use strict'
+
+/**
+ * Unflag all licences in a bill run that did not result in a billing invoice (they are unbilled)
+ * @module UnflagUnbilledLicencesService
+ */
+
+const LicenceModel = require('../../models/water/licence.model.js')
+
+/**
+ * Unflag any licences that were not billed as part of a bill run
+ *
+ * Some licences will not result in an invoice (`billing_invoice_licence`) being created. For example, you could add a
+ * new charge version that does nothing but change the description of the previous one. In isolation, this would result
+ * in an `EMPTY` bill run. If others are being processed at the same time, it would just mean no records are added to
+ * the bill run for this licence.
+ *
+ * If this is the case, we can remove the 'Include in SROC Supplementary Billing' flag from the licence now. Even if
+ * the bill run gets cancelled and re-run later the result will be the same.
+ *
+ * What it also means is we can be accurate with which licences get unflagged when the bill run is finally **SENT**. In
+ * the old logic they simply unflag any licence in a region where `date_updated` is less than the bill run's
+ * created date. But we know this is flawed because a licence can be updated after a bill run is created, for example
+ * the NALD import process updates them all. If this happens the flag remains on.
+ *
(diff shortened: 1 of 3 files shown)


app/services/supplementary-billing/unflag-unbilled-licences.service.js
```diff
@@ -0,0 +1,59 @@
...
+ * The query we run during the **SEND** process unflags only those which are linked to the bill run being sent. We can
+ * do this because we know this service has handled anything that was unbilled and not represented.
+ *
+ * @param {*} billingBatchId The ID of the bill run (billing batch) being processed
+ * @param {module:ChargeVersionModel[]} chargeVersions All charge versions being processed in the bill run
+ * @returns {Number} count of records updated
+ */
+async function go (billingBatchId, chargeVersions) {
+  const allLicenceIds = _extractUniqueLicenceIds(chargeVersions)
+  allLicenceIds.push('310f05c0-98d9-4e70-9857-71e5106e7559')
```
Jozzey, line 35: Is this line supposed to be here?

Jozzey (changes requested):

app/services/supplementary-billing/unflag-unbilled-licences.service.js
Jozzey: I'm referring to `allLicenceIds.push('310f05c0-98d9-4e70-9857-71e5106e7559')`

Jozzey (approved):

PR #223, opened 2023-05-15
Add endpoint for Db export

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4008

This pull request is just a small part of a larger project that involves exporting all our database schemas, converting them into CSV files, and uploading them to our Amazon S3 bucket. This PR's primary focus is creating the endpoint which can kick off the entire db-export service. To expedite the export process and see the output sooner, we are using a vertical slicing approach, rather than a horizontal one, which means exporting a single table at a time from each schema.


.env.example
```diff
@@ -43,3 +43,9 @@ LEGACY_AUTH_TOKEN=longvalueofnumbersandletters.inbothcases.splitin3by2periods
 CHARGING_MODULE_TOKEN_URL=https://myinstance.amazoncognito.com
 CHARGING_MODULE_TOKEN_USERNAME=valuefullofnumbersandlettersinlowercase
 CHARGING_MODULE_TOKEN_PASSWORD=longvaluefullofnumbersandlettersinlowercase
+
+# AWS S3 config
+AWS_REGION=eu-west-1
+AWS_ACCESS_KEY_ID=uploadaccesskey
+AWS_SECRET_ACCESS_KEY=uploadsecretkey
+AWS_MAINTENANCE_BUCKET=upload-bucket-gov-uk
```
Cruikshanks, line 51: Just to avoid the scary 'no entry' sign in GitHub we try and ensure all files end with a new line.

Screenshot 2023-05-16 at 12 48 37

I think there is even a setting in VSCode that will ensure this is there when you save a file.

```suggestion
AWS_MAINTENANCE_BUCKET=upload-bucket-gov-uk

```

test/controllers/data/data.controller.test.js
```diff
@@ -59,4 +60,40 @@ describe('Data controller', () => {
...
+
+      it('displays the correct message', async () => {
+        const response = await server.inject(options)
+        const payload = JSON.parse(response.payload)
+
+        expect(response.statusCode).to.equal(200)
+        expect(payload.status).to.equal('successful')
+      })
+    })
+    describe('when the service fails', () => {
```
Cruikshanks, line 83:
```suggestion

    describe('when the service fails', () => {
```

Cruikshanks (changes requested):
Just 2 quick changes to do with missing new lines. That's it!

Cruikshanks (approved):

PR #227, opened 2023-05-16
Adding logging to the db export service

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-3968

This pull request is just a small part of a larger project that involves exporting all our database schemas, converting them into CSV files, and uploading them to our Amazon S3 bucket. This PR's primary focus is to add logging to the db-export service. To expedite the export process and see the output sooner, we are using a vertical slicing approach, rather than a horizontal one, which means exporting a single table at a time from each schema.


Cruikshanks (approved):
Look! No changes! 😁

PR #228, opened 2023-05-17
Multi-Year SROC Supplementary Billing

Jozzey:

Supplementary billing for SRoC currently only handles Financial Year 2022 - 2023, going forward it needs to pick up any changes to charge versions in any year from 2022. Once we have passed Annual Billing 2027 then the engine should only pick up changes to charge versions in the current year and the previous 5.

All transactions in the available billing period should be picked up in 1 bill run, and be triggered by a single action.

This PR adds the remainder of the functionality required for Multi-Year SROC Supplementary Billing not covered in the previous PR https://github.com/DEFRA/water-abstraction-system/pull/226


app/services/supplementary-billing/process-billing-batch.service.js
```diff
@@ -10,19 +10,35 @@ const ProcessBillingPeriodService = require('./process-billing-period.service.js
...
-    const isPopulated = await ProcessBillingPeriodService.go(billingBatch, currentBillingPeriod, chargeVersions)
+    for (const billingPeriod of billingPeriods) {
+      const chargeVersions = await _fetchChargeVersions(billingBatch, billingPeriod)
+      const isPeriodPopulated = await ProcessBillingPeriodService.go(billingBatch, billingPeriod, chargeVersions)
 
-    await _finaliseBillingBatch(billingBatch, chargeVersions, isPopulated)
+      const licenceIdsForPeriod = _extractLicenceIds(chargeVersions)
+      const previousLicenceIds = accumulatedLicenceIds
+
+      accumulatedLicenceIds = previousLicenceIds.concat(licenceIdsForPeriod)
```
StuAA78, line 30: We could do this in a couple of lines I think (but worth double-checking before committing!)
```suggestion
      const licenceIdsForPeriod = _extractLicenceIds(chargeVersions)
      accumulatedLicenceIds.push(...licenceIdsForPeriod)
```
Doing this would also let us declare `accumulatedLicenceIds` with `const`

StuAA78 (changes requested):

app/services/supplementary-billing/process-billing-batch.service.js
Jozzey: Cheers @StuAA78 I thought you'd have a better way of doing that :star:

app/services/supplementary-billing/unflag-unbilled-licences.service.js
```diff
@@ -27,12 +27,10 @@ const LicenceModel = require('../../models/water/licence.model.js')
  * do this because we know this service has handled anything that was unbilled and not represented.
  *
  * @param {*} billingBatchId The ID of the bill run (billing batch) being processed
- * @param {module:ChargeVersionModel[]} chargeVersions All charge versions being processed in the bill run
+ * @param {Array} allLicenceIds All licence IDs being processed in the bill run
```
StuAA78, line 30: I think we've settled on using the `[]` syntax now eg. `{String[]}` for an array of strings.

Jozzey:
As this PR is going to significantly impact the supplementary billing testing that is currently being carried out. It shouldn't be merged until we get the OK from the business/QA.

Cruikshanks (changes requested):
The specific change is just about the order of `const` and `let`.

But I would also push for updating the first test ('when there are charge versions that should be considered for the next supplementary billing') in `test/services/supplementary-billing/fetch-charge-versions.service.test.js`. We should update the billing period to 2023-04-01 - 2024-03-31. The existing test SROC charge version should still be picked up. But we can also include one that starts in the billing period just for completeness. 

The tests will then need to be updated to confirm both SROC charge versions are returned.

app/services/supplementary-billing/process-billing-batch.service.js
```diff
@@ -10,19 +10,33 @@ const ProcessBillingPeriodService = require('./process-billing-period.service.js
...
 
 async function go (billingBatch, billingPeriods) {
-  const currentBillingPeriod = billingPeriods[billingPeriods.length - 1]
   const { billingBatchId } = billingBatch
 
   try {
     // Mark the start time for later logging
     const startTime = process.hrtime.bigint()
+    let isBatchPopulated = false
+    const accumulatedLicenceIds = []
```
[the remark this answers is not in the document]
Jozzey: You're getting as bad as Stuart :laughing:

Cruikshanks (approved):

PR #230, opened 2023-05-17
Process billing batch move db writes

StuAA78:
https://eaflood.atlassian.net/browse/WATER-3996

This PR aims to make `ProcessBillingBatchService` more performant by moving database writes out of the main loop. As part of this we undertake additional refactoring to clean up the main loop.


Cruikshanks (approved):
In my opinion I think we could pass around more specific values than the bulk data objects in some places. I also think with all that is going on in this service there are a few things we could do elsewhere.

But I also know we need to move some stuff out ASAP to support the multi-year work. And with that done what remains I think screams out for splitting into smaller, more focused services.

So, I'm keeping those details to myself and approving this! We can catch up after on the final refactoring/cleanup 😁 

Other than that, this is going to really help 💪

PR #232, opened 2023-05-18
Generalising fetch billing charge categories service

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4015

This pull request is just a small part of a larger project that involves exporting all our database schemas, converting them into CSV files, and uploading them to our Amazon S3 bucket. This PR's primary focus is to generalise the current service we have so that it can take any table as an argument. To expedite the export process and see the output sooner, we are using a vertical slicing approach, rather than a horizontal one, which means exporting a single table at a time from each schema.


app/services/db-export/db-export.service.js
```diff
@@ -1,28 +1,20 @@
...
-  const filePath = await ExportDataFilesService.go(tableConvertedToCsv, data.tableName)
-
-  const compressedFilePath = await CompressFilesService.go(filePath)
-
-  await SendToS3BucketService.go(compressedFilePath)
+  const tableName = 'billing_charge_categories'
+  const schemaName = 'water'
 
-  await DeleteFileService.go(filePath)
+  return await TableExportService.go(tableName, schemaName)
```
Cruikshanks, line 17: Having had a look at the `TableExportService` it doesn't appear to be returning anything explicitly. So, if that isn't returning anything there isn't really anything to return here. I'd suggest

```suggestion
  await TableExportService.go(tableName, schemaName)
```

test/services/db-export/table-export.service.test.js
```diff
@@ -13,26 +13,26 @@ const ConvertToCSVService = require('../../../app/services/db-export/convert-to-
...
 const ExportDataFilesService = require('../../../app/services/db-export/export-data-files.service.js')
-const FetchBillingChargeCategoriesService = require('../../../app/services/db-export/fetch-billing-charge-categories.service.js')
+const FetchTableService = require('../../../app/services/db-export/fetch-table.service.js')
 const SendToS3BucketService = require('../../../app/services/db-export/send-to-s3-bucket.service.js')
 
 // Thing under test
-const DbExportService = require('../../../app/services/db-export/db-export.service.js')
+const TableExportService = require('../../../app/services/db-export/table-export.service.js')
 
 describe('Db Export service', () => {
```
Cruikshanks, line 22: Think this top-level describe needs updating

```suggestion
describe('Table export service', () => {
```

Cruikshanks (changes requested):
Just a couple of suggested tweaks!

Cruikshanks (approved):

StuAA78 (approved):

PR #233, opened 2023-05-18
Refactor start of processing a supp. billing batch

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3977

Both to help resolve some issues found when cleaning up our existing process billing batch service and to support changes need for handling multi-year bill runs we need to make some changes.

The key one is `ProcessBillingBatchService` is no longer appropriate because it is actually focused on processing a single billing period. So, we renamed it to `ProcessBillingPeriodService` and have created a new `ProcessBillingBatchService` responsible for orchestrating the _whole_ process.

As the primary service now, we've updated the controller to call it instead of `InitiateBillingBatchService`. This resulted in us needing to make further changes in it and the services it uses.

Along the way we spotted opportunities to clean up some other things; passing in params only to find they are hard coded elsewhere, or having params named differently to how they are used within a service.


test/controllers/bill-runs/bill-runs.controller.test.js
```diff
@@ -58,12 +58,14 @@ describe('Bill Runs controller', () => {
       }
 
       beforeEach(async () => {
-        Sinon.stub(InitiateBillingBatchService, 'go').resolves(validResponse)
+        Sinon.stub(ProcessBillingBatchService, 'go').resolves(validResponse)
       })
 
       it('returns a 200 response including details of the new billing batch', async () => {
         const response = await server.inject(options())
+        console.log('🚀 ~ file: bill-runs.controller.test.js:66 ~ it ~ response:', response)
```
Jozzey, line 66: Don't think you meant to leave this `console.log()` in.

test/controllers/bill-runs/bill-runs.controller.test.js
```diff
@@ -58,12 +58,14 @@ describe('Bill Runs controller', () => {
...
         const payload = JSON.parse(response.payload)
+        console.log('🚀 ~ file: bill-runs.controller.test.js:68 ~ it ~ payload:', payload)
```
Jozzey, line 68: ...or this `console.log()`

Jozzey (changes requested):
Found a couple of minor issues

StuAA78 (approved):

PR #236, opened 2023-05-21
Refactor, unit tests & add NewBillingBatchService

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3977

This is a follow on from [Refactor start of processing a supp. billing batch](https://github.com/DEFRA/water-abstraction-system/pull/233). In there we needed to refactor the `ProcessBillingBatchService` to not mix processing a given billing period and the overall billing batch. So, we created `ProcessBillingPeriodService` and split the functionality between the 2.

But we parked writing any unit tests for our new `ProcessBillingBatchService` till this PR, else the change would have been massive. Only we hit an issue when we started writing them.

`ProcessBillingBatchService` in its new form was determining the billing periods, initiating the billing batch record, and generating a response back to the client all whilst kicking off the process for each billing period and handling any errors thrown. It was also responsible for finalising the bill run.

A bit much! The final issue was that background processing was not `awaited`. This made it a nightmare to try and test all this behaviour.

We realised we needed to get the 'processing' bit in its own service if only to make it testable. So, this change adds a new `NewBillingBatchService` (yes - we really need to sit down and firm up our naming conventions!)

_This_ is now what is called when kicking off a new billing batch. It deals with determining the billing periods, initiating the billing batch, and generating the response for the client. Essentially, it manages everything to do with a new bill run. It also kicks off `ProcessBillingBatchService` in the background, which is now solely focused on processing a billing batch; generating the data, handling errors, updating the billing batch status, and finalising it.

With this in place, it meant we could also write some unit tests!


app/services/supplementary-billing/new-billing-batch.service.js
```diff
@@ -0,0 +1,37 @@
...
+const InitiateBillingBatchService = require('./initiate-billing-batch.service.js')
+const ProcessBillingBatchService = require('./process-billing-batch.service.js')
+
+async function go (regionId, userEmail) {
+  const billingPeriods = BillingPeriodsService.go()
+  const financialYearEndings = _financialYearEndings(billingPeriods)
+
+  const billingBatch = await InitiateBillingBatchService.go(financialYearEndings, regionId, userEmail)
+
+  ProcessBillingBatchService.go(billingBatch, billingPeriods)
```
StuAA78, line 19:
```suggestion
  // We do not `await` the billing batch being processed so we can leave it to run in the background while we return an immediate response
  ProcessBillingBatchService.go(billingBatch, billingPeriods)
```

app/services/supplementary-billing/new-billing-batch.service.js
```diff
@@ -0,0 +1,37 @@
...
+
+  return _response(billingBatch)
+}
+
+function _financialYearEndings (billingPeriods) {
+  return {
+    fromFinancialYearEnding: billingPeriods[billingPeriods.length - 1].endDate.getFullYear(),
+    toFinancialYearEnding: billingPeriods[0].endDate.getFullYear()
```
StuAA78, line 27: We should be able to use `billingPeriods.at(-1)` to get the last item in a slightly cleaner way. We could also use `billingPeriods.at(0)` for consistency.

```suggestion
    fromFinancialYearEnding: billingPeriods.at(-1).endDate.getFullYear(),
    toFinancialYearEnding: billingPeriods.at(0).endDate.getFullYear()
```

app/services/supplementary-billing/process-billing-batch.service.js
```diff
@@ -1,39 +1,106 @@
...
+const UnflagUnbilledLicencesService = require('./unflag-unbilled-licences.service.js')
 
-async function go (region, user) {
-  // NOTE: It will be required in the future that we cater for a range of billing periods, as changes can be back dated
-  // up to 5 years. For now though, our delivery scope is only for the 2022-2023 billing period so the final record is
-  // extracted from the `billingPeriods` array which will currently always be for the 2022-2023 billing period.
-  const billingPeriods = BillingPeriodsService.go()
-  const financialYearEndings = _financialYearEndings(billingPeriods)
+async function go (billingBatch, billingPeriods) {
+  const currentBillingPeriod = billingPeriods[billingPeriods.length - 1]
```
StuAA78, line 13: Again, we should be able to use `.at(-1)` for a slightly cleaner way of getting the last item.
```suggestion
  const currentBillingPeriod = billingPeriods.at(-1)
```

StuAA78 (changes requested):
Great stuff 😁 Just a couple of small tweaks might neaten things up a little.

PR #237, opened 2023-05-22
Getting table names from schema

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4017

This pull request is just a small part of a larger project that involves exporting all our database schemas, converting them into CSV files, and uploading them to our Amazon S3 bucket. This PR's primary focus is to get the table names from the database using only the schema name provided. To expedite the export process and see the output sooner, we are using a vertical slicing approach, rather than a horizontal one, which means exporting a single table at a time from each schema.


Cruikshanks (changes requested):
Just some really minor formatting issues. But the meat of the change is all gravy!

Cruikshanks (approved):

PR #238, opened 2023-05-22
Export a schema

Beckyrose200:

This pull request is just a small part of a larger project that involves exporting all our database schemas, converting them into CSV files, and uploading them to our Amazon S3 bucket. This PR's primary focus is to export an entire database schema from start to finish. To expedite the export process and see the output sooner, we are using a vertical slicing approach, rather than a horizontal one, which means exporting a single table at a time from each schema.


app/services/db-export/send-to-s3-bucket.service.js
```diff
@@ -47,7 +74,7 @@ async function _uploadToBucket (params, fileName) {
 
     return true
   } catch (error) {
-    return false
+    throw new Error()
```
StuAA78, line 77: This doesn't need to be a `try...catch` block now that we're no longer handling the error in this function. The calling function has it wrapped in a `try...catch` so if `s3Client.send()` throws an error then it will be caught by that; we don't need to catch it and then throw a new error.

StuAA78 (changes requested):
Great stuff! Most of what I've suggested is pretty minor. And there's also the stuff we discussed earlier about replacing the sync versions of file handling functions with the async ones where promises.

app/services/db-export/export-data-files.service.js
```diff
@@ -7,22 +7,22 @@
...
-async function go (tableConvertedToCsv, tableName) {
-  const filePath = _filenameWithPath(tableName)
+async function go (tableConvertedToCsv, tableName, schemaFolderPath) {
+  const filePath = _filenameWithPath(tableName, schemaFolderPath)
 
   try {
-    await fs.writeFile(filePath, tableConvertedToCsv)
+    fs.writeFileSync(filePath, tableConvertedToCsv)
 
     return filePath
```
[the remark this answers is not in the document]
Beckyrose200: It returns the filePath as where its being called it needs the file path to give as an argument to the next service

app/services/db-export/export-data-files.service.js
```diff
@@ -7,42 +7,44 @@
 
 const path = require('path')
 
-const fs = require('fs').promises
-const os = require('os')
+const fs = require('fs')
+const fsp = fs.promises
```
StuAA78, line 11: Thinking about it, `fsPromises` is probably more in line with how we'd normally name this?

app/services/db-export/export-data-files.service.js
```diff
@@ -7,42 +7,44 @@
...
  *
  * @returns {String} The full file path
  */
-function _filenameWithPath (tableName) {
-  const temporaryFilePath = os.tmpdir()
+function _filenameWithPath (tableName, schemaFolderPath) {
+  const schemaFolderExists = fs.existsSync(schemaFolderPath)
+
+  if (!schemaFolderExists) {
+    fs.mkdirSync(schemaFolderPath)
```
StuAA78, line 42: I just spotted the docs do seem to mention [`fsPromises.mkdir`](https://nodejs.org/docs/latest-v16.x/api/fs.html#fspromisesmkdirpath-options), if that would work here?

app/services/db-export/schema-export.service.js
```diff
@@ -0,0 +1,50 @@
...
+ * @module SchemaExportService
+ */
+
+const path = require('path')
+const os = require('os')
+
+const DeleteFolderService = require('./delete-folder.service.js')
+const FetchTableNames = require('../db-export/fetch-table-names.service.js')
+const SendToS3BucketService = require('../db-export/send-to-s3-bucket.service.js')
+const ExportCompressedTableService = require('./export-compressed-table.service.js')
```
StuAA78, line 14: Just a small tweak to alphabetise the requires.
```suggestion
const os = require('os')
const path = require('path')

const DeleteFolderService = require('./delete-folder.service.js')
const ExportCompressedTableService = require('./export-compressed-table.service.js')
const FetchTableNames = require('../db-export/fetch-table-names.service.js')
const SendToS3BucketService = require('../db-export/send-to-s3-bucket.service.js')
```

StuAA78 (changes requested):
There might be a couple more places we could use `fsPromises` but otherwise I think we're good 😁

app/services/db-export/export-data-files.service.js
```diff
@@ -5,44 +5,45 @@
...
  * @returns {String} The full file path
  */
-function _filenameWithPath (tableName) {
-  const temporaryFilePath = os.tmpdir()
+async function _filenameWithPath (tableName, schemaFolderPath) {
+  const schemaFolderExists = fs.existsSync(schemaFolderPath)
+
+  if (!schemaFolderExists) {
+    fsPromises.mkdir(schemaFolderPath)
+  }
```
StuAA78, line 42: We may be able to replace this with `await fsPromises.mkdir(schemaFolderPath, { recursive: true })` to simply try creating the folder with no error thrown if it already exists. And then we can drop `fs` from this file completely in favour of `fsPromises`

StuAA78 (changes requested):
I spotted a couple of things in the docs that may be able to do things in a slightly easier fashion 😄

StuAA78 (approved):

PR #250, opened 2023-05-31
Uploading compressed tarball file to S3

Beckyrose200:

This pull request is just a small part of a larger project that involves exporting all our database schemas, converting them into CSV files, and uploading them to our Amazon S3 bucket. 
This PR's primary focus is to upload a single compressed tarball file to our S3 bucket rather then individual table files. 
To expedite the export process and see the output sooner, we are using a vertical slicing approach, rather than a horizontal one, which means exporting a single table at a time from each schema.


StuAA78 (changes requested):
This is great! I noticed we're using `readFileSync` which we'd be better off avoiding if possible. The  "define filename as a const" comment is just a small thing we could maybe refactor. The comment about `_uploadToBucket()` is more of a nice to have and it's a little dependent on what was decided on about error handling so I'm not going to insist we make changes to it at the moment if it needs any more than a few mins of work.

app/services/db-export/compressed-tarball.service.js
Cruikshanks, line 1: It's a common convention that when we code, we do so in verbs (a word used to describe an action) and nouns (a word that refers to a thing).

- CreateBillingBatchService
- DetermineChargePeriodService
- FetchRegionService

We're essentially saying 'do x with y' when we use a service, call a method etc. Because of this, we avoid past tense in anything, for example, `CreatedBillingBatchService`.

The next thing I'd flag is that a tarball is a feature of the service but it's probably not the most obvious property. The aim of this service was to take our folder of tables, derived from a schema and compress them into a single file.

So, with both those ideas in mind, I'd suggest renaming this service to something like

- `CompressFolderService`
- `CompressSchemaFolderService`
- `CompressSchemaService`

app/services/db-export/export-compressed-table.service.js
Cruikshanks, line 1: I might have this wrong but I don't think this service is doing any compression anymore. If so we should ideally rename it.

Cruikshanks (changes requested):
I've got some requested changes around service names and I can see Stuart has made some suggestions.

I didn't dig into the unit tests just yet but will do once these suggestions have been reviewed and rolled in.

app/services/db-export/send-to-s3-bucket.service.js
```diff
@@ -1,79 +1,55 @@
...
- *
- * @returns {[]} An array of file paths within the folder
- */
-async function _getFilesFromFolder (folderPath) {
-  const files = await fsPromises.readdir(folderPath)
 
-  return files.map((file) => {
-    return path.join(folderPath, file)
-  })
+  return _uploadToBucket(params)
```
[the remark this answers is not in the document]
Beckyrose200: Thank you, I will be looking at the erroring handling over the whole service in my next PR so this will be a change I consider for that :)

app/services/db-export/compress-schema-folder.service.js
```diff
@@ -0,0 +1,31 @@
...
+  const file = `${schemaFolderPath}.tgz`
+
+  await tar.create(
+    {
+      gzip: true,
+      file
+    },
+    [schemaFolderPath]
+  )
+  return `${schemaFolderPath}.tgz`
```
Cruikshanks, line 26: As we have defined `file` as a const we can just return that as the result rather than workout the file path again.

We also have a convention of putting a newline between the body of a method and its `return` statement.

```suggestion
  )
  
  return file
```

app/services/db-export/send-to-s3-bucket.service.js
```diff
@@ -1,79 +1,55 @@
...
 
 /**
- * Uploads a folders worth of files to our S3 bucket
+ * Sends a file to our AWS S3 bucket
  * @module SendToS3BucketService
  */
 
 const fsPromises = require('fs').promises
 const path = require('path')
 const { PutObjectCommand, S3Client } = require('@aws-sdk/client-s3')
```
Cruikshanks, line 10: Eeek! I can see you removed the line but actually, our [convention](https://medium.com/selleo/essential-rubyonrails-patterns-part-1-service-objects-1af9f9573ca1) _is_ to put a break before the config.

app/services/db-export/send-to-s3-bucket.service.js
Cruikshanks, line 1: I think we just need to work through what this service is returning, how and why.

@StuAA78 was right to point out that when we call `SendToS3BucketService` in `SchemaExportService` we don't capture its return value. So, at this time we don't need to return anything.

This means we don't need to wrap the `await s3Client.send(command)` call in `_uploadToBucket()` in a try-catch to figure out a `true` or `false` value. In fact, we're swallowing the error and losing the detail of what went wrong by doing so.

So, we can drop the return from `_uploadToBucket()`. We can also drop the return in `go()`; it just becomes `await _uploadToBucket()`.

test/services/db-export/compress-schema-folder.service.test.js
```diff
@@ -0,0 +1,37 @@
...
+const { describe, it, beforeEach, afterEach } = exports.lab = Lab.script()
+const { expect } = Code
+
+// Things we need to stub
+const tar = require('tar')
+
+// Thing under test
+const CompressSchemaFolderService = require('../../../app/services/db-export/compress-schema-folder.service.js')
+
+describe('Compressed schema folder service', () => {
```
Cruikshanks, line 17:
```suggestion
describe('Compress schema folder service', () => {
```

Cruikshanks (changes requested):
Sorry, bit more feedback. The main points are around the `SendToS3BucketService` which I'm happy to pair on to get sorted and the PR approved as soon as you have time.

app/services/db-export/send-to-s3-bucket.service.js
Beckyrose200: I knew that the services were not returning the right values and the error handling was something I wanted to work on in the next PR. However I have added a few commits to this PR to fix the return value issues and added the error handling that I think works. 

The way the service handle the errors now is with a single try-catch block in the schema-export.service file.

So if an error occurs during the export of a schema or its individual tables, the error will be caught, and the process will return from that specific schema. This will prevents the possibility of attempting to upload a broken or incomplete schema file, as it halts the process when encountering errors.

In case of an error, the schema-export.service will capture the error and send a notification message indicating the specific schema that failed using the global notifier. Afterwards, it will proceed to perform cleanup operations by deleting the schema folder that was created, which contains all the individual table CSV files. Additionally, if the schema.tgz file exists, it will also be deleted.

This cleanup process remains the same whether the file is successfully uploaded to the S3 bucket or if an error occurs during the export.

Cruikshanks (approved):

StuAA78 (approved):

PR #256, opened 2023-06-06
Implement SROC invoice reissuing

StuAA78:

We have [created the new services we need](https://github.com/DEFRA/water-abstraction-system/pull/255) to send the appropriate requests to the Charging Module. We now implement invoice reissuing for SROC.

This is done with `ReissueInvoicesService`, which calls `FetchInvoicesToBeReissuedService` to get the invoices for reissuing then passes each one to `ReissueInvoiceService` (note singular `Invoice`) which handles the actual reissuing of the invoice. The resulting invoices, invoice licences and transactions are persisted, and a `true` value returned to indicate that db changes have been made.

The intent is that this service will be called from `ProcessBillingBatchService`, and its return value used to set the `isBatchPopulated` boolean. This PR initially made some changes to `ProcessBillingBatchService` to accommodate this; however due to a merge conflict (and the fact that the changes don't really belong in this PR) the changes in the [`Refactor to accommodate future service`](https://github.com/DEFRA/water-abstraction-system/pull/256/commits/b251f8a4ba2bd77c74d55021badf4c54c02633da) commit have been reverted. They will be reinstated in the future PR which adds reissuing to the process billing batch process.

### DIFF
--- a/app/services/supplementary-billing/fetch-invoices-to-be-reissued.service.js
+++ b/app/services/supplementary-billing/fetch-invoices-to-be-reissued.service.js
@@ -35,11 +35,20 @@ async function go (regionId) {
       })
 
     return result
-  } catch (_error) {
+  } catch (error) {
     // If getting invoices errors then we log the error and return an empty array; the db hasn't yet been modified at
     // this stage so we can simply move on to the next stage of processing the billing batch.
 
-    // TODO: add logging
+    global.GlobalNotifier.omfg('Could not fetch reissue invoices', {
+      region: regionId,
+      error: {
+        name: error.name,
+        message: error.message,
+        stack: error.stack,
+        code: error.code
+      }
+    })
+
     return []
   }
 }
--- a/app/services/supplementary-billing/fetch-invoices-to-be-reissued.service.js
+++ b/app/services/supplementary-billing/fetch-invoices-to-be-reissued.service.js
@@ -39,15 +39,11 @@ async function go (regionId) {
     // If getting invoices errors then we log the error and return an empty array; the db hasn't yet been modified at
     // this stage so we can simply move on to the next stage of processing the billing batch.
 
-    global.GlobalNotifier.omfg('Could not fetch reissue invoices', {
-      region: regionId,
-      error: {
-        name: error.name,
-        message: error.message,
-        stack: error.stack,
-        code: error.code
-      }
-    })
+    global.GlobalNotifier.omfg(
+      'Could not fetch reissue invoices',
+      { region: regionId },
+      error
+    )
 
     return []
   }
--- a/app/services/supplementary-billing/fetch-invoices-to-be-reissued.service.js
+++ b/app/services/supplementary-billing/fetch-invoices-to-be-reissued.service.js
@@ -1,0 +1,40 @@
+'use strict'
+
+/**
+ * Fetches billing invoices to be reissued
+ * @module FetchInvoicesToBeReissuedService
+ */
+
+const BillingInvoiceModel = require('../../models/water/billing-invoice.model.js')
+
+/**
+ * Takes a region and fetches billing invoices in that region marked for reissuing, along with their transactions
+ *
+ * @param {String} regionId The uuid of the region
+ *
+ * @returns {module:BillingInvoiceModel[]} Array of billing invoices to be reissued
+ */
+async function go (regionId) {
+  try {
+    // TODO: optimise this to only return what we need from the billing invoice licences and the transactions
+    const result = await BillingInvoiceModel.query()
+      .joinRelated('billingBatch')
+      .withGraphFetched('billingInvoiceLicences.billingTransactions')
+      .where({
+        'billingBatch.regionId': regionId,
+        isFlaggedForRebilling: true
+      })
+
+    return result
+  } catch (_error) {
+    // If getting invoices errors then we log the error and return an empty array; the db hasn't yet been modified at
+    // this stage so we can simply move on to the next stage of processing the billing batch.
+
+    // TODO: add logging
+    return []
+  }
+}
+
+module.exports = {
+  go
+}
--- a/app/services/supplementary-billing/fetch-invoices-to-be-reissued.service.js
+++ b/app/services/supplementary-billing/fetch-invoices-to-be-reissued.service.js
@@ -16,13 +16,22 @@ const BillingInvoiceModel = require('../../models/water/billing-invoice.model.js
  */
 async function go (regionId) {
   try {
-    // TODO: optimise this to only return what we need from the billing invoice licences and the transactions
     const result = await BillingInvoiceModel.query()
+      .select(
+        'billingInvoices.externalId',
+        'invoiceAccountId',
+        'invoiceAccountNumber',
+        'originalBillingInvoiceId'
+      )
+      .where('isFlaggedForRebilling', true)
       .joinRelated('billingBatch')
+      .where('billingBatch.regionId', regionId)
       .withGraphFetched('billingInvoiceLicences.billingTransactions')
-      .where({
-        'billingBatch.regionId': regionId,
-        isFlaggedForRebilling: true
+      .modifyGraph('billingInvoiceLicences', (builder) => {
+        builder.select(
+          'licenceRef',
+          'licenceId'
+        )
       })
 
     return result
--- a/app/services/supplementary-billing/reissue-invoice.service.js
+++ b/app/services/supplementary-billing/reissue-invoice.service.js
@@ -48,16 +48,16 @@ async function go (sourceInvoice, reissueBillingBatch) {
 
   // When a reissue request is sent to the Charging Module, it creates 2 new invoices (one to cancel out the original
   // invoice and one to be the new version of it) and returns their ids
-  const chargingModuleReissueResponses = await _sendReissueRequest(
+  const chargingModuleReissueResponse = await _sendReissueRequest(
     reissueBillingBatch.externalId,
     sourceInvoice.externalId
   )
 
-  for (const chargingModuleReissueResponse of chargingModuleReissueResponses) {
+  for (const chargingModuleReissueInvoiceInfo of chargingModuleReissueResponse) {
     // Because we only have the CM invoice's id we now need to fetch its details via the "view invoice" endpoint
     const chargingModuleReissueInvoice = await _sendViewInvoiceRequest(
       reissueBillingBatch,
-      chargingModuleReissueResponse
+      chargingModuleReissueInvoiceInfo
     )
 
     const reissueBillingInvoice = _retrieveOrGenerateBillingInvoice(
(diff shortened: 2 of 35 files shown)


app/services/supplementary-billing/fetch-invoices-to-be-reissued.service.js
```diff
@@ -0,0 +1,58 @@
...
+
+    global.GlobalNotifier.omfg('Could not fetch reissue invoices', {
+      region: regionId,
+      error: {
+        name: error.name,
+        message: error.message,
+        stack: error.stack,
+        code: error.code
+      }
+    })
```
Cruikshanks, line 50: Having just done [Better handle errors in Notifiers](https://github.com/DEFRA/water-abstraction-system/pull/273) the way the object is being created and passed to the notifier means we'd loose pino and Errbit seeing the original error.

I'm guessing you want to include the region ID in the packet to help when debugging. If this is the case we should be calling it like this now

```suggestion
    global.GlobalNotifier.omfg(
      'Could not fetch reissue invoices',
      { region: regionId },
      error
    )
```

Have a look at https://github.com/DEFRA/water-abstraction-system/blob/main/app/lib/base-notifier.lib.js#L83 and it should make more sense.

Cruikshanks (changes requested):
I've not finished the full review @StuAA78 

But spotted this change and thought I'd get it across whilst I review the rest 😬

Cruikshanks (changes requested):
A couple more typos.

app/services/supplementary-billing/reissue-invoices.service.js
```diff
@@ -0,0 +1,68 @@
...
+ * one of these we call `ReissueInvoiceService` which handles the actual reissuing process and collect the returned
+ * invoice, invoice licence and transaction data which we batch persist one all invoices have been reissued
+ *
+ * @param {module:BillingBatchModel} originalBillingBatch The billing batch to check for invoices to be reissued
+ * @param {module:BillingBatchModel} reissueBillingBatch The billing batch that the reissued invoices will be created on
+ *
+ * @returns {Boolean} `true` if any invoices were reissued; `false` if not
+*/
+
+async function go (originalBillingBatch, reissueBillingBatch) {
```
Cruikshanks, line 27: This comment is going to apply to the use of `originalBillingBatch` both here and in `ReissueInvoiceService`.

I don't believe we need it. The 'source' for what to reissue comes from `FetchInvoicesToBeReissuedService` and all it needs is the region ID.

We can get that from `reissueBillingBatch` which then makes `originalBillingBatch` redundant here.

app/services/supplementary-billing/reissue-invoice.service.js
```diff
@@ -0,0 +1,267 @@
...
+ * @param {module:BillingBatchModel} originalBillingBatch The billing batch that the original invoice belongs to
+ * @param {module:BillingBatchModel} reissueBillingBatch The billing batch that the new invoices should belong to
+ *
+ * @returns {Object} dataToReturn Data that has been generated while reissuing the invoice
+ * @returns {Object[]} dataToReturn.billingInvoices Array of billing invoices
+ * @returns {Object[]} dataToReturn.billingInvoiceLicences Array of billing invoice licences
+ * @returns {Object[]} dataToReturn.transactions Array of transactions
+ */
+
+async function go (sourceInvoice, originalBillingBatch, reissueBillingBatch) {
```
Cruikshanks, line 43: This is tied to my other comment about the `originalBillingBatch`.

When we send the request to the Charging Module we need to provide it with the bill run ID that we want the invoices reissued against. In this context that will be `reissueBillingBatch.externalId`.

The Charging Module will then create the invoices against it, which means we should use the same `reissueBillingBatch.externalId` in `_sendViewRequest()`.

So, again I think we can remove `originalBillingBatch` as it is redundant in this context.

Cruikshanks (changes requested):

app/services/supplementary-billing/reissue-invoices.service.js
```diff
@@ -0,0 +1,69 @@
...
+  return true
+}
+
+/**
+ * Iterate over each key in `dataToPersist` and add to it the corresponding data in `newData`
+ */
+function _addNewDataToDataToPersist (dataToPersist, newData) {
+  Object.keys(dataToPersist).forEach((key) => {
+    dataToPersist[key].push(...newData[key])
+  })
```
Cruikshanks, line 58: Assuming I've got it right, I don't think this is a case where iterating the keys is adding anything (IMHO).

I think it'd be clearer and require no thinking to just see

```suggestion
  dataToPersist.billingInvoices.push(...newData.billingInvoices)
  dataToPersist.billingInvoiceLicences.push(...newData.billingInvoiceLicences)
  dataToPersist.transactions.push(...newData.transactions)
```

app/services/supplementary-billing/reissue-invoices.service.js
```diff
@@ -0,0 +1,69 @@
...
+  const sourceInvoices = await FetchInvoicesToBeReissuedService.go(reissueBillingBatch.regionId)
+
+  if (sourceInvoices.length === 0) {
+    return false
+  }
+
+  const dataToPersist = {
+    billingInvoices: [],
+    billingInvoiceLicences: [],
+    transactions: []
```
Cruikshanks, line 38: I was just wondering why the invoices and invoice-licences keep the `billing` prefix but `transactions` doesn't. Have I let this convention slip through into other areas!? 😰

app/services/supplementary-billing/reissue-invoice.service.js
```diff
@@ -0,0 +1,267 @@
...
+ * object with the new invoices, invoice licences and transactions.
+ *
+ * Once the invoice has been handled, we mark the source invoice as `rebilled` in the db (note that "rebill" is legacy
+ * terminology for "reissue") along with the `originalBillingInvoiceId` field; if this is empty then we update it with
+ * the source invoice's billing invoice id, or if it's already filled in then we leave it as-is. This ensures that if an
+ * invoice is reissued, and that reissuing invoice is itself reissued, then `originalBillingInvoiceId` will still point
+ * directly to the original source invoice.
+ *
+ * @param {module:BillingInvoiceModel} sourceInvoice The invoice to be reissued
+ * @param {module:BillingBatchModel} originalBillingBatch The billing batch that the original invoice belongs to
```
Cruikshanks, line 34: This one can go
```suggestion
```

app/services/supplementary-billing/reissue-invoice.service.js
```diff
@@ -0,0 +1,267 @@
...
+
+  return newBillingInvoiceLicence
+}
+
+async function _sendReissueRequest (billingBatchExternalId, invoiceExternalId) {
+  try {
+    const result = await ChargingModuleReissueInvoiceService.go(billingBatchExternalId, invoiceExternalId)
+    return result.response.invoices
+  } catch (error) {
+    throw new Error(`Charging Module reissue request failed: ${error.message}`)
```
Cruikshanks, line 252: I'm not sure we would ever actually hit these error blocks. The ChargingModule*Services all depend on `ChargingModuleRequestLib` which in turn depends on `RequestLib`.

They don't add much logic between the call and what `Requestlib` does and `RequestLib` ensures all errors are caught, logged, and then the result object that's returned is populated with the error.

We should properly be checking `result.succeeded` and then deciding whether an error needs to be logged and thrown.

app/services/supplementary-billing/reissue-invoice.service.js
```diff
@@ -0,0 +1,267 @@
...
+  }
+}
+
+async function _sendViewInvoiceRequest (billingBatch, reissueInvoice) {
+  try {
+    const result = await ChargingModuleViewInvoiceService.go(billingBatch.externalId, reissueInvoice.id)
+    return result.response.invoice
+  } catch (error) {
+    throw new Error(`Charging Module view invoice request failed: ${error.message}`)
```
Cruikshanks, line 261: Dito my previous comment about try/catch around Charging Module service calls

app/services/supplementary-billing/reissue-invoice.service.js
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  // When a reissue request is sent to the Charging Module, it creates 2 new invoices (one to cancel out the original
+  // invoice and one to be the new version of it) and returns their ids
+  const chargingModuleReissueResponses = await _sendReissueRequest(
+    reissueBillingBatch.externalId,
+    sourceInvoice.externalId
+  )
+
+  for (const chargingModuleReissueResponse of chargingModuleReissueResponses) {
```
Cruikshanks, line 57: In `_sendReissueRequest()` we're returning `result.response.invoices` so I'd suggest simply calling them `invoices` here.

```suggestion
  const invoices = await _sendReissueRequest(
    reissueBillingBatch.externalId,
    sourceInvoice.externalId
  )

  for (const invoice of invoices) {
```

Cruikshanks (changes requested):
Give me a shout if you need to dig into/discuss any of my comments. I think the error handling is the key one we need to straighten out.

app/services/supplementary-billing/reissue-invoice.service.js
```diff
@@ -0,0 +1,276 @@
...
+      for (const sourceTransaction of sourceInvoiceLicence.billingTransactions) {
+        // Get the original transaction from the charging module data so we can create our new one
+        const chargingModuleReissueTransaction = _retrieveChargingModuleTransaction(
+          chargingModuleLicence,
+          sourceTransaction.externalId
+        )
+
+        // We need to know if this is an invoice which is cancelling out the original so we can create our transactions
+        // accordingly (ie. creating credit transactions to cancel out debits and vice-versa)
+        const isCancellingInvoice = _isCancellingInvoice(chargingModuleReissueInvoiceInfo)
```
Cruikshanks, line 94: Fortunately, the Charging Module has our back on this! When it creates the 2 invoices it works out creating a debit for a credit and vice-versa. So, premise is we need to 'copy' what the charging module has created into our DB but fill in the blanks for anything it doesn't provide.

The legacy code appears to have gone with an approach of

- start with the source transaction
- overwrite matching properties from the CM transaction
- strip out the ID

We can follow the same or go with starting with the CM transaction and filling in the blanks from the source. Either way, `isCredit` is sorted for us so we don't need to flip anything.

Cruikshanks (changes requested):

test/services/supplementary-billing/reissue-invoice.service.test.js
```diff
@@ -0,0 +1,268 @@
...
+        const result = await ReissueInvoiceService.go(sourceInvoice, reissueBillingBatch)
+
+        const credits = result.billingTransactions.filter(transaction => transaction.isCredit)
+
+        credits.forEach((transaction) => {
+          expect(transaction.netAmount).to.be.below(0)
+        })
+      })
+
+      it('positive for credits', async () => {
```
Cruikshanks, line 206:
```suggestion
      it('positive for debits', async () => {
```

Cruikshanks (changes requested):
Just a teeny-weeny typo!

Cruikshanks (approved):

PR #265, opened 2023-06-13
Break up `GenerateBillingTransactionsService` into separate services

StuAA78:

Currently `GenerateBillingTransactionsService` does the work of calculating the number of billable days, determining if a compensation transaction is required, then converting provided data into the format required by a transaction record.

Doing all this in one service means we can't reuse the data conversion code elsewhere (ie. when creating transactions for reissuing).

This PR aims to break `GenerateBillingTransactionsService` into two separate services: one that calculates the number of billable days and determines if a compensation transaction is required, and one that converts data into the required format. This second service can then be called from elsewhere.


StuAA78:
While developing this and cross-referencing against the current work we have on reissuing invoices, we realised there isn't much point in this refactor right now, as the transactions we're creating for reissuing are pretty much fully-formatted so in actual fact we wouldn't really make use of a separate "generate single transaction" service. Although this refactoring may be something we want to consider in future, we will shelve it for the time being.

PR #266, opened 2023-06-13
SROC Supplementary flag not cleared for 'non-chargeable' licence

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4041

It has been found in QA that when a licence has no other SROC charge versions and is made 'non-chargeable', this correctly sets the SROC supplementary billing flag on the licence. However, the problem is that when a bill run that includes this licence is created, the SROC supplementary billing flag is not un-set for it.

This issue is being caused because when a licence is made “non-chargeable” it creates a charge version without any charge elements, but also without an `invoice_account_id`.

Our `fetch-charge-versions` service only fetches charge versions which have an `invoice_account_id`. This in effect means that if a licence only has the “non-chargeable” charge version associated with it. That licence will never be looked at, since the licence ID is extracted from the results of the fetch-charge-versions service. As the licence is never included by the SROC supplementary service the flag is never un-set.

This PR fixes that issue by initially fetching the charge versions with no `invoice_account_id`, rather than filtering them out in the query as was done previously.

The `licence_id`s are then extracted from the results of this query before finally stripping out the charge versions that have no `invoice_account_id`.

### DIFF
--- a/app/services/supplementary-billing/fetch-charge-versions.service.js
+++ b/app/services/supplementary-billing/fetch-charge-versions.service.js
@@ -115,6 +115,14 @@ async function _fetch (regionId, billingPeriod) {
   return uncleansedChargeVersions
 }
 
+/**
+ * Extract the `licence_id`s from the charge versions before removing charge versions that have no `invoice_account_id`
+ *
+ * When a licence is made "non-chargeable" the supplementary billing flag gets set and a charge version created that has
+ * no `invoice_account_id`. For the purpose of billing we are not interested in charge versions with no
+ * `invoice_account_id` as these will not be charged. We are however interested in the associated licences to ensure
+ * that "non-chargeable" licences have thier supplementary billing flag un-set.
+ */
 function _extractLicencesAndCleanseChargeVersions (uncleansedChargeVersions) {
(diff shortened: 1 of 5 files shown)


app/services/supplementary-billing/fetch-charge-versions.service.js
```diff
@@ -113,7 +112,30 @@ async function _fetch (regionId, billingPeriod) {
...
+  return uncleansedChargeVersions
+}
+
+/**
+ * Extract the `licence_id`s from the charge versions before removing charge versions that have no `invoice_account_id`
+ *
+ * When a licence is made "non-chargeable" the supplementary billing flag gets set and a charge version created that has
+ * no `invoice_account_id`. For the purpose of billing we are not interested in charge versions with no
+ * `invoice_account_id` as these will not be charged. We are however interested in the associated licences to ensure
+ * that "non-chargeable" licences have thier supplementary billing flag un-set.
```
Cruikshanks, line 124:
```suggestion
 * that "non-chargeable" licences have their supplementary billing flag un-set.
```

Cruikshanks (dismissed):

Cruikshanks (approved):

PR #274, opened 2023-06-20
Supplementary bill run doesn't credit old account correctly

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4044

An issue was found when a Charge Version was charged in the 2024 FY but later superseded in the 2023 FY due to a Billing contact change. When the supplementary billing is re-run, the charge made to the original contact in 2024 is not being considered as the endDate of the Charge Version is before the start of the 2024 FY. This means that the Charge Version is not picked up, resulting in previous charges against the replaced invoice account being ignored rather than credited.


Cruikshanks:
This has been replaced by https://github.com/DEFRA/water-abstraction-system/pull/280

PR #296, opened 2023-07-06
Add new custom ExpandedError

Cruikshanks:
When working on [Implement SROC invoice reissuing](https://github.com/DEFRA/water-abstraction-system/pull/256) we had to deal with a situation where we wanted to throw an error that contained details of the records in the current context.

But we were faced with the issue that we knew the error would eventually find its way to our notifiers. If we include the details in the error message we break Errbit's ability to group instances of an error. But the standard [Error constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error) doesn't support passing in additional values.

To avoid making things complicated for ourselves we did this

```javascript
const error = new Error('Charging Module reissue request failed')
error.billingBatchExternalId = billingBatchExternalId
error.invoiceExternalId = invoiceExternalId

throw error
```

We know both Pino our logger will include these extra properties in its output and Errbit can still do its stuff.

Then when working on [Fix omfg() in HandleErroredBillingBatch](https://github.com/DEFRA/water-abstraction-system/pull/295) we did a review of all our `omfg()` calls. This led us to find this in `InitiateBillingBatchService`

```javascript
if (liveBillRunExists) {
  throw Error(`Batch already live for region ${regionId}`)
}
```

This is an example of an error message that will block Errbit from grouping these errors. We could fix this by doing what we did in `ReissueInvoiceService`. But it's probably time to add our own custom error that allows us to do it properly.

This change adds the error and applies it where applicable.


Jozzey (approved):
👍🏼

PR #304, opened 2023-07-11
Move and make generic some billing services

Cruikshanks:

Continuing our theme of getting things ready for two-part tariff (2PT) bill runs we have already spotted that `app/services/billing/supplementary/fetch-region.service.js` will be needed by both types.

So, this PR either moves 'generic' billing services to the root `app/services/billing/` level, updates them to be generic, or both!


Cruikshanks:
This looks bigger than it is. With files moving you need to dip into both the main an test files of others to update the references. Also, git has decided there was enough change to warrant a couple of services to appear deleted and then added.

But I'm happy to walk through it if anyone has concerns/is struggling. I just felt making the create bill run journey generic to the point of kicking off processing the billing batch was better done as one change.

Jozzey (approved):

PR #310, opened 2023-07-17
Fix reissuing bugs

StuAA78:

Reissuing failed QA. The following fixes have therefore been applied:

- After sending the reissue request to the CM, we wait for the bill run status to leave `pending` before proceeding;
- We ensure the correct data is being passed back from the CM;
- We pull some additional fields from the db for invoices to ensure the records are completed;
- We found that inserting transactions where some data is in object format would fail. This is because even though the field's datatype is `jsonb`, Objection expects us to stringify the data first. We initially explored doing this, however we found that Objection allows us to define JSON fields in the model (by adding a `jsonAttributes` getter) which means it will stringify the data for us if needed. We therefore update `BillingTransactionModel` to make use of this.

### DIFF
--- a/test/models/water/billing-transaction.model.test.js
+++ b/test/models/water/billing-transaction.model.test.js
@@ -110,5 +110,22 @@ describe('Billing Transaction model', () => {
         })
       ).to.not.reject()
     })
+
+    it('returns an object from a json field regardless of whether the inserted object was stringified first', async () => {
+      const objectTransaction = await BillingTransactionModel.query().insert({
+        ...BillingTransactionHelper.defaults(),
+        purposes: [{ test: 'TEST' }]
+      })
+      const stringifyTransaction = await BillingTransactionModel.query().insert({
+        ...BillingTransactionHelper.defaults(),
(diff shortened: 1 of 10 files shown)


test/models/water/billing-transaction.model.test.js
```diff
@@ -97,4 +97,18 @@ describe('Billing Transaction model', () => {
       })
     })
   })
+
+  describe('Inserting', () => {
+    // Objection doesn't normally allow us to insert an object directly into a json field unless we stringify it first.
+    // However if we define jsonAttributes in our model with the json fields then we don't need to stringify the object.
+    // This test is therefore to check whether jsonAttributes is correctly working.
+    it('can insert an object directly into a json field', async () => {
```
Jozzey, line 105: As well as/or instead of checking that it doesn't blow up when we insert the `purposes`, can we check that the `purposes` data is returned correctly when we query the table following the insert?

Jozzey (changes requested):
Just 1 comment

PR #311, opened 2023-07-19
Fix reissuing ALCS invoice bug

StuAA78:

Reissuing was confirmed working in our local environments but was continuing to fail in `dev`. When we checked the logs we realised that it was failing because it was fetching alcs invoices that were awaiting reissuing as well as sroc ones! This then causes an error to come back from the CM as it won't reissue an alcs invoice onto an sroc bill run.

We resolve this by updating `FetchInvoicesToBeReissuedService` to only fetch invoices linked to an sroc billing batch.

### DIFF
--- a/test/services/billing/supplementary/fetch-invoices-to-be-reissued.service.test.js
+++ b/test/services/billing/supplementary/fetch-invoices-to-be-reissued.service.test.js
@@ -101,6 +101,7 @@ describe('Fetch Invoices To Be Reissued service', () => {
         const result = await FetchInvoicesToBeReissuedService.go(billingBatch.regionId)
 
         expect(result).to.have.length(1)
+        expect(result[0].billingInvoiceId).to.equal(billingInvoice.billingInvoiceId)
       })
     })
   })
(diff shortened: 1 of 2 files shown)


test/services/billing/supplementary/fetch-invoices-to-be-reissued.service.test.js
```diff
@@ -30,6 +31,15 @@ describe('Fetch Invoices To Be Reissued service', () => {
     billingInvoice = await BillingInvoiceHelper.add({ billingBatchId: billingBatch.billingBatchId })
     const { billingInvoiceLicenceId } = await BillingInvoiceLicenceHelper.add({ billingInvoiceId: billingInvoice.billingInvoiceId })
     await BillingTransactionHelper.add({ billingInvoiceLicenceId })
+
+    // Set up a separate billing batch which we will manually patch to be alcs
+    const alcsBillingBatch = await BillingBatchHelper.add()
```
Jozzey, line 36: Don't think we need line 42 if we write line 36 as `const alcsBillingBatch = await BillingBatchHelper.add({ scheme: 'alcs' })`

Jozzey (changes requested):
Just 1 comment.

Jozzey (approved):
👍🏼

PR #319, opened 2023-07-25
Using streams on schema export service

Beckyrose200:

This PR's primary focus is to fix the schema export service. The service is currently not working due to the fact it reaches its memory capacity when it exports the returns table. This is due to the fact the full table is brought into memory. This fix introduced streams into the service to allow the table to be written immediately to a file and not be held in full in memory.


app/services/data/export/write-stream-to-file.service.js
```diff
@@ -0,0 +1,66 @@
+'use strict'
+
+/**
+ * Export converted data to a temporary file
+ * @module WriteStreamToFileService
```
StuAA78, line 5: The name `WriteStreamToFileService` isn't quite correct any more; we're writing some headers and _then_ a stream to a file. So I wonder if `WriteTableToFileService` is more accurate? Alternatively, it's possible to get the table headers from the db as a stream, and then we change `ExportTableService` to call `WriteStreamToFileService` twice -- once with the header stream, and once with the rows stream. It should be fairly straightforward to do and might be a little cleaner overall, but since what we've got does work then I'm not sure if it's worth making the changes?

app/services/data/export/convert-to-csv.service.js
```diff
@@ -8,42 +8,31 @@
...
  * @returns {String} A CSV formatted string
  */
-function go (headers, rows) {
-  const transformedHeaders = _transformDataToCSV([headers])[0]
-
-  if (!rows) {
-    return transformedHeaders
+function go (data) {
+  if (!data) {
+    return
```
StuAA78, line 17: Do we want this to return an empty string instead of `undefined`? Because then we're consistent with the type that the service returns?

StuAA78 (changes requested):
This is looking great, just a few changes to consider.

app/services/data/export/write-stream-to-file.service.js
Beckyrose200: I think I prefer the name change to writeTableToFileService rather then calling the stream twice :)

app/services/data/export/write-stream-to-file.service.js
```diff
@@ -0,0 +1,66 @@
...
+const util = require('util')
+
+const ConvertToCSVService = require('./convert-to-csv.service.js')
+
+/**
+ * Converts data into CSV format and writes it to a file
+ *  @param {Object} data The knex query to fetch the table
+ * @param {String} schemaFolderPath The folder path of the schema
+ */
+async function go (data, schemaFolderPath) {
```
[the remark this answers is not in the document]
Beckyrose200: With this do you mean passing the arguments in like so (data.rows, data.headers, schemaFolderPath, TableName) or do you mean the data object not being made?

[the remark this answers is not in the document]
StuAA78: It's up to you -- in `ExportTableService` you could either do this as you say:
```js
  const data = await FetchTableService.go(tableName, schemaName)

  await WriteTableToFileService.go(data.rows, data.headers, schemaFolderPath, tableName)
```
Or you could destructure them into separate variables first:
```js
  const { headers, rows } = await FetchTableService.go(tableName, schemaName)

  await WriteTableToFileService.go(rows, headers, schemaFolderPath, tableName)
```

StuAA78 (approved):

PR #320, opened 2023-07-25
Improve reissuing error logging

StuAA78:

We found that checking the logs for reissuing errors caused by the Charging Module returning an error response wasn't as straightforward as it could be; instead of having all the info logged in one error, we had to find the separately-logged CM error response to see why the request errored. This change adds the response body to our logged error (as `responseBody`) so we don't need to go hunting for the reason for the error (eg. "Invoice has already been rebilled").


Jozzey (approved):
👍🏼

PR #328, opened 2023-08-01
Match CV data to returns data for 2PT billing

Jozzey:

First iteration of mathcing returns to charge versions. This covers the first example which is a one-to-one match.

We also have updated the results to include charge versions, charge elements and charge purposes. We intend to refine this in future iterations but we are adding them to confirm the matching algorithm is working correctly.


Cruikshanks:
We just need to add unit tests for the new model hence put back to draft.

Cruikshanks (dismissed):

Cruikshanks (approved):

StuAA78 (approved):

PR #330, opened 2023-08-02
Update migrations to follow team convention

Cruikshanks:
As a team, we have agreed that [our convention](https://github.com/DEFRA/water-abstraction-team/blob/main/coding_conventions.md#arrow-functions) when using arrow functions is to always wrap arguments in brackets even if there are only 1.

We spotted with a recent migration that we were not doing this. At the same time, we also clocked we were doing the following in some migrations.

```javascript
exports.up = async function (knex) {
  await knex
    .schema

// vs

exports.down = function (knex) {
  return knex
    .schema
```

The [Knex example](https://knexjs.org/guide/migrations.html#transactions-in-migrations) implies what we are doing in `down()` is correct. Looking at some of our simpler migrations, for example, when creating schemas we saw that's what we were already doing.

As we tend to copy and paste previous migrations when adding new ones we've decided to correct what we have to save repeating the infractions in future changes!


Jozzey (dismissed):

db/migrations/20221108152353_create-water-charge-versions.js
```diff
@@ -6,7 +6,7 @@ exports.up = async function (knex) {
   await knex
```
StuAA78, line 6: Not strictly related to this PR but I noticed while reviewing that we have a lot of migrations where the `up` method is `await knex...` while the down method is `return knex...`. The [example here](https://knexjs.org/guide/migrations.html#transactions-in-migrations) implies that `return knex...` is the preferred way of doing it:
```js
exports.up = function(knex) {
  return knex.schema
    .createTable('users', function (table) {
        // ...
    })
};
```
So, do we want to change this? And if so, do we change it in this PR or a separate one? Alternatively, "if it ain't broke don't fix it" and we maybe just follow the convention going forwards?

Cruikshanks: I think we should fix it for the same reason I gave in the PR; we tend to copy existing migrations that match what we are trying to do and use them as a template for the new one.

So, I might as well get it done in this PR as our team convention is to be consistent!

db/migrations/20221114104700_create-water-regions.js
```diff
@@ -6,7 +6,7 @@ exports.up = async function (knex) {
   await knex
```
StuAA78, line 6: I think chaining multiple things using `.then()` should work if we wanted to be consistent about using `return knex` everywhere:
```js
  return knex
    .schema
    .withSchema('water')
    .createTable(tableName, table => {
      ...
    })
    .then(() => {
      knex.raw(`
        ...
      `)
    })
```
But is it actually worth it? 🤷

Cruikshanks: I'll give it a go!

Jozzey (approved):
👍🏼

StuAA78 (approved):

PR #337, opened 2023-08-03
Add crm_v2 to DatabaseHelper clean() method

Cruikshanks:
When we added the first `crm_v2` schema models to our repo we forgot to add the schema to our `DatabaseHelper` and its `clean()` method.

We use this to ensure the DB is empty of existing data before each test to ensure test isolation. This change adds the schema to the helper.

Note - we also had to tweak the query to wrap the schema name in quotes. It's because PostgreSQL doesn't actually like the name because it is case-sensitive 🤦😩. Without the quotes, everything falls over.


Jozzey (approved):

test/support/helpers/database.helper.js
```diff
@@ -32,7 +32,7 @@ async function _tableNames (schema) {
     .select('tablename')
     .where('schemaname', schema)
 
-  return result.map((table) => `${schema}.${table.tablename}`)
+  return result.map((table) => `"${schema}".${table.tablename}`)
```
StuAA78, line 35:
```suggestion
  return result.map((table) => {
    return `"${schema}".${table.tablename}`
  })
```

StuAA78 (changes requested):
One small tweak to reflect our current standards while we're changing this file!

test/support/helpers/database.helper.js
Cruikshanks: You had mentioned that and I forgot about it. Good spot!

StuAA78 (approved):

PR #340, opened 2023-08-07
Switch to Charge Version as base for 2PT matching

Cruikshanks:

Update the existing matching algorithm for matching 2PT returns to charge versions to be based on `charge_versions` as the primary table instead of licences.

For example, a charge version may have 2 charge elements that match and therefore currently cause our results to spit out the same licence twice. Also, we know eventually the data will be used to feed billing which uses charge versions as the primary data source.


Cruikshanks:
We are getting told off by SonarCloud on code coverage but that is to be expected @StuAA78

PR #347, opened 2023-08-09
Create endpoint to generate fake data

StuAA78:
For prototype and test purposes we want to be able to generate realistic-looking fake data. We therefore create an endpoint to do this.

### DIFF
--- a/app/controllers/data/data.controller.js
+++ b/app/controllers/data/data.controller.js
@@ -8,6 +8,7 @@
 const Boom = require('@hapi/boom')
 
 const ExportService = require('../../services/data/export/export.service.js')
+const MockService = require('../../services/data/mock/mock.service.js')
 const SeedService = require('../../services/data/seed/seed.service.js')
 const TearDownService = require('../../services/data/tear-down/tear-down.service.js')
 
@@ -20,6 +21,17 @@ async function exportDb (_request, h) {
   ExportService.go()
 
   return h.response().code(204)
+}
+
+async function mock (request, h) {
+  try {
+    const { type, id } = request.params
+    const mockData = await MockService.go(type, id)
+
+    return h.response(mockData)
+  } catch (error) {
+    return Boom.badImplementation(error.message)
+  }
 }
 
 async function seed (_request, h) {
@@ -44,6 +56,7 @@ async function tearDown (_request, h) {
 
 module.exports = {
   exportDb,
+  mockData: mock,
   seed,
   tearDown
 }
--- a/app/controllers/data/data.controller.js
+++ b/app/controllers/data/data.controller.js
@@ -22,14 +22,10 @@ async function exportDb (_request, h) {
 }
 
 async function mock (request, h) {
-  try {
-    const { type, id } = request.params
-    const mockData = await MockService.go(type, id)
+  const { type, id } = request.params
+  const mockData = await MockService.go(type, id)
 
-    return h.response(mockData)
-  } catch (error) {
-    return Boom.badImplementation(error.message)
-  }
+  return h.response(mockData)
 }
 
 async function seed (_request, h) {
--- a/app/presenters/data/mock-bill-run.presenter.js
+++ b/app/presenters/data/mock-bill-run.presenter.js
@@ -1,0 +1,205 @@
+'use strict'
+
+/**
+ * Formats the response for the GET `/data/mock/{bill-run}` endpoint
+ * @module MockBillRunPresenter
+ */
+
+const { formatAbstractionPeriod, formatLongDate, formatNumberAsMoney } = require('../base.presenter.js')
+
+function go (billingBatch) {
+  const {
+    billingInvoices,
+    billRunNumber,
+    createdAt,
+    fromFinancialYearEnding,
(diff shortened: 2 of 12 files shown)


app/controllers/data/data.controller.js
```diff
@@ -44,6 +56,7 @@ async function tearDown (_request, h) {
 
 module.exports = {
   exportDb,
+  mockData: mock,
```
StuAA78, line 59: Do we want to rename our `mock` function to be `mockData` instead of aliasing it in the exports?

app/presenters/data/mock-bill-run.presenter.js
```diff
@@ -0,0 +1,205 @@
+'use strict'
+
+/**
+ * Formats the response for the GET `/data/mock/{bill-run}` endpoint
+ * @module MockBillRunPresenter
+ */
+
+const { convertPenceToPounds, formatAbstractionPeriod, formatLongDate, formatNumberAsMoney } = require('../base.presenter.js')
```
StuAA78, line 8: This line is just a little too long and should probably be split up.

app/presenters/data/mock-bill-run.presenter.js
```diff
@@ -0,0 +1,205 @@
...
+  return {
+    dateCreated: formatLongDate(createdAt),
+    status,
+    region: region.name,
+    type,
+    chargeScheme: scheme === 'sroc' ? 'Current' : 'Old',
+    transactionFile,
+    billRunNumber,
+    financialYear: `${fromFinancialYearEnding} to ${toFinancialYearEnding}`,
+    debit: formatNumberAsMoney(convertPenceToPounds(netTotal)),
```
StuAA78, line 34: We do `formatNumberAsMoney(convertPenceToPounds())` in several places. Do we want to break it out into a function of its own, `_formatPenceAsPounds()` or something?

app/presenters/data/mock-bill-run.presenter.js
```diff
@@ -0,0 +1,205 @@
...
+}
+
+function _formatChargePurposes (chargePurposes) {
+  return chargePurposes.map((chargePurpose) => {
+    const {
+      purposesUse,
+      abstractionPeriodStartDay: startDay,
+      abstractionPeriodStartMonth: startMonth,
+      abstractionPeriodEndDay: endDay,
+      abstractionPeriodEndMonth: endMonth,
```
StuAA78, line 189: Do we need to alias these seeing as we're passing them to a function and not using them directly? Or do we want to keep that cos if we don't alias them, our call to the function would be too big for one line?

app/services/data/mock/generate-bill-run.service.js
```diff
@@ -0,0 +1,190 @@
...
+  if (!billingBatch) {
+    throw new Error('No matching bill run exists')
+  }
+
+  _mockBillingInvoices(billingBatch.billingInvoices)
+
+  return _response(billingBatch)
+}
+
+async function _fetchBillingBatch (id) {
```
StuAA78, line 24: Not 100% certain this needs to be `async` seeing as it returns a query, and it's the returned query that we `await`; but not going to quibble over it!

app/services/data/mock/generate-bill-run.service.js
```diff
@@ -0,0 +1,190 @@
...
+/**
+ * Generates a mock bill run based on a real one
+ * @module GenerateBillRunService
+ */
+
+const BillingBatchModel = require('../../../models/water/billing-batch.model.js')
+const GenerateMockDataService = require('./generate-mock-data.service.js')
+const MockBillRunPresenter = require('../../../presenters/data/mock-bill-run.presenter.js')
+
+async function go (id) {
```
StuAA78, line 12: Do we need some brief docs here?

app/services/data/mock/generate-bill-run.service.js
```diff
@@ -0,0 +1,190 @@
...
+ * Calculate the totals for a licence based on the transaction values
+ *
+ * We don't hold totals in the `billing_invoice_licence` record. But the UI shows them. We found it is calculating
+ * these on the fly in the UI code so we need to replicate the same behaviour.
+ *
+ * Another thing to note is that if a transaction is flagged as a credit, then `netAmount` will be held as a signed
+ * value, for example -213.40. This is why it might look confusing we are always adding on each iteration but the
+ * calculation will be correct.
+ * @param {*} transactions
+ * @returns
```
StuAA78, line 166: I think I'd want to see us either flesh out `@param` with the type and `@returns` with some details, or remove them completely.

StuAA78 (commented):
All looking good, just a few small comments (it won't let me Request Changes since it's my own PR)

Cruikshanks (approved):
Paired on this with @StuAA78 . Admittedly, the later changes were done just by me so ideally would be confirmed as all good before merging.

But he's had the audacity to take time off (!!) and it would be unfair to ask the rest of the team to review this massive change cold.

So, making the executive decision to approve and merge-away! This is not a feature that is intended for use in `production` anyway 😬

PR #351, opened 2023-08-11
Create authentication plugin

StuAA78:
https://eaflood.atlassian.net/browse/WATER-4085

We implement `AuthenticationPlugin` to authenticate users. This is purely intended for use by frontend routes, ie. those where a request is passed from the UI to the system. 

> Note that you will need to add `COOKIE_SECRET` to your `.env` file. This should have the same value as `COOKIE_SECRET` in your UI's `.env`.

The plugin creates an authentication strategy `session` which implements the `cookie` auth scheme.

The main bit of authentication is done by the strategy before our validation handler runs. It checks that the `sid` cookie passed on from the UI is valid, and if not it redirects the user to `/signin`.

If the cookie is valid then the plugin's validation handler calls on `AuthenticationService` to do its own authentication, which in turn makes use of `FetchUserRolesAndGroupsService` to fetch user data from the IDM.

`AuthenticationService` resolves to an object populated with the data from `FetchUserRolesAndGroupsService`, which the `AuthenticationPlugin` passes on to the underlying auth strategy:

- `isValid` boolean is set `true` if the user exists in the IDM, or `false` if they don't. The auth strategy makes use of this to reject the request with a 403 error if it's set to `false`.
- The `credentials` object is populated with the `user`, `groups` and `roles` which were fetched from the IDM. At present we intend to make use of `user` to read things like the user's email address; at present the `groups` and `roles` are simply "nice to have", although this may change in future.
- `credentials` is also populated with an array named `scope`, which contains the names of the roles the user has (which it gets from the `role` property of each role object). `credentials.scope` is used by hapi for authentication; if a route has `options.auth.strategy.scope` with an array of scopes then hapi will reject requests if the user's `credentials.scope` does not contain at least one of those strings. In other words, putting one or more role names into a route's `options.auth.strategy.scope` will prevent requests if the user doesn't have at least one of the roles. More details on scope can be found in the [hapi docs](https://hapi.dev/api/?v=21.3.2#-routeoptionsauthaccessscope).

If authentication succeeds, the auth strategy will populate `request.auth` with `credentials` amongst other things; see the [hapi docs](https://hapi.dev/api/?v=21.3.2#-requestauth) for more details. This means for example that we can access the user's email via `request.auth.credentials.user.userName`.


app/services/idm/fetch-user-roles-and-groups.service.js
```diff
@@ -16,11 +16,12 @@ const UserModel = require('../../models/idm/user.model.js')
  * This service looks up the user in the `idm` (identity management) schema and returns a combined array of all roles
  * (deduped in case the user is given the same role multiple times; for example, by being assigned a role directly, then
  * later added to a group which also includes that role). It also returns an array of groups that the user is a member
- * of.
+ * of, along with `userFound` to explicitly indicate whether or not the user id exists.
  *
  * @param {Number} userId The user id to get roles and groups for
  *
  * @returns {Object} result The resulting roles and groups
+ * @returns {UseModel[]} result.user Returns the UserModel representing the user, or `null` if the user is not found
```
Cruikshanks, line 24: I found something! Think the JSDoc just needs tweaking as `result.user` is a single object rather than an array.

```suggestion
 * @returns {UseModel} result.user Returns the UserModel representing the user, or `null` if the user is not found
```

app/services/plugins/authentication.service.js
```diff
@@ -0,0 +1,43 @@
...
+ * Used by `AuthenticationPlugin` to authenticate and authorise users
+ * @module AuthenticationService
+ */
+
+const FetchUserRolesAndGroupsService = require('../idm/fetch-user-roles-and-groups.service.js')
+
+/**
+ * This service is intended to be used by our `AuthenticationPlugin` to authenticate and authorise users.
+ *
+ * We take a user id and look it up in the `idm` table using `FetchUserRolesAndGroupsService`. This gives us a user
```
Cruikshanks, line 13: I think this is what you mean. 🤔 

```suggestion
 * We take a user id and look it up in the `idm` schema using `FetchUserRolesAndGroupsService`. This gives us a user
```

Cruikshanks (changes requested):
Had a good look through and I can only find some possible tweaks needed to the JSDoc comments.

app/services/idm/fetch-user-roles-and-groups.service.js
StuAA78: Good catch! And I notice I missed the `r` out of `UserModel` too! 🙄

Cruikshanks (approved):

Cruikshanks (approved):

PR #354, opened 2023-08-15
Add Got retry hook to log when retries happen

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4083

We had an issue with the Midlands SROC supplementary bill run. It would error and when checking the logs we found it was because the [charging-module-api](https://github.com/DEFRA/sroc-charging-module-api) was rejecting a `POST /transaction` request.

```json
{
  "result": {
    "succeeded": false,
    "response": {
      "statusCode": 409,
      "body": {
        "statusCode": 409,
        "error": "Conflict",
        "message": "A transaction with Client ID 'ebee5d48-3191-4615-b86a-5bebb9026bfe' for Regime 'wrls' already exists.",
        "clientId": "ebee5d48-3191-4615-b86a-5bebb9026bfe"
      }
    }
  }
}
```

But this should be impossible. We generate the ID's during the bill run. If the bill run errors all transaction records get deleted. We then found a second attempt to run the bill run succeeded.

We've been unable to replicate the failed scenario but have a working theory. We believe a `POST /transaction` request is failing due to a timeout. That doesn't mean the CHA didn't receive the request, just that [Got](https://github.com/sindresorhus/got) timed out waiting for it. We have it configured to automatically retry these. So, in this scenario, the CHA receives the first request and creates the transaction but **Got** times out. It then sends the same request a second time, by which point the record already exists in the CHA so it rejects it.

To confirm this, and see when **Got** is retrying this change adds a new [beforeRetry](https://github.com/sindresorhus/got/blob/main/documentation/9-hooks.md#beforeretry) hook to the **Got** instance we create. In it, we can log that a retry has taken place and details about it.

### DIFF
--- a/app/lib/request.lib.js
+++ b/app/lib/request.lib.js
@@ -46,33 +46,6 @@ function _beforeRetryHook (error, retryCount) {
   global.GlobalNotifier.omg('Retrying HTTP request', { error, retryCount })
 }
 
-async function _sendRequest (method, url, additionalOptions) {
-  const got = await _importGot()
-  const result = {
-    succeeded: false,
-    response: null
-  }
-
-  try {
-    const options = _requestOptions(additionalOptions)
-
-    result.response = await got[method](url, options)
-
-    // If the result is not 2xx or 3xx Got will mark the result as unsuccessful using the response object's `ok:`
-    // property
-    result.succeeded = result.response.ok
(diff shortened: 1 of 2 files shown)


Jozzey (approved):
👍🏼

PR #358, opened 2023-08-16
Add proxy support to SendToS3BucketService

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4086

While testing the data export service on the development environment, it was identified that the upload process was unsuccessful due to the lack of support for proxy settings.
This change adds support for proxy by checking if a proxy server is configured. If a server is configured, we add these settings to a new customConfig object. We pass this object to the S3 client.

During testing on the development environment, we were waiting 6 minutes for the S3 client to error. This was due to the default set timeout being 6 minutes long. This was far too long during testing as we had to wait until we had a readable error. This change sets the timeout to be a maximum of 10 seconds now. This config is always passed to the S3 client.


app/services/data/export/send-to-s3-bucket.service.js
```diff
@@ -5,40 +5,224 @@
...
+ *
+ * @param {*} s3Client The AWS S3 Client instance used to interact with the service
+ * @param {String} bucketName The name of the bucket where the object will be stored
+ * @param {String} key The key (path) of the object within the bucket
+ * @param {Buffer} buffer The buffer containing the data to be uploaded in parts
+ * @param {String} uploadId The uploadId associated with the created multipart upload
+ *
+ * @returns An array of responses from the upload promises for each part.
+ */
+async function _uploadPartCommand (s3Client, bucketName, key, buffer, uploadId) {
```
Cruikshanks, line 173: Apologies, I've got a few comments going on. But I don't think the name of this method corresponds to what it is doing.

My understanding is this is responsible for uploading _all_ the parts, not just a single one? If so it might help with an additional thought; this function is doing too much.

I would suggest renaming and splitting out the work. For example, this becomes `_splitIntoPartsAndUpload()`. Then everything in the `for()` loop gets moved to a function called `_uploadPart()`. 🤷

app/services/data/export/send-to-s3-bucket.service.js
```diff
@@ -5,40 +5,224 @@
...
+          })
+        )
+        .then((response) => {
+          // Once the promise is resolved return the response metadata (this includes the ETag)
+          return response
+        })
+    )
+  }
+
+  return await Promise.all(uploadPromises)
```
Cruikshanks, line 202: As `_uploadPartCommand()` is currently being awaited I think we can just do

 ```suggestion
  return Promise.all(uploadPromises)
```

File: app/services/data/export/send-to-s3-bucket.service.js [hunk 3]
```diff
@@ -5,40 +5,224 @@
...
+}
+
+/**
+ * Uploads a single file content to the specified S3 bucket with the provided key
+ *
+ * @param {Buffer} buffer The file content as a buffer object
+ * @param {String} bucketName Name of the S3 bucket to upload the file to
+ * @param {String} key The path under which the file will be stored in the bucket
```
Cruikshanks, line 210: These params don't match the args of the function.

Cruikshanks (changes requested):
More comments now away from our conventions.

A key thing to think about is

- are the JSDoc params adding value where they are used
- if their primary value is explaining that `x` actually means `y`, should we just be thinking of a better name for the `x` param

There are also some corrections and suggestions. Shout if you have any questions.

Cruikshanks (approved):

PR #359, opened 2023-08-16
Implement TwoPartService 'friendly' response

Cruikshanks:

We have our `TwoPartService` which is our spike on coming up with a working two-part tariff matching algorithm. But the data it matches has issues

- there is _lots_ of it. The records have numerous fields and many are JSONB which means when presented they expand out even more
- for 'reasons' few of the tables and hence the entity names match what you see in the UI

Reducing the amount of data output to just the key elements will help when assessing whether a match is right. And as you'll generally have the UI open and the relevant record up it would _really_ help if the response matched what you saw on screen.

So, we have created a `FriendlyResponseService`. We fully intend to delete it once we have confirmation the matching algorithm is good. This is our excuse for why it looks horrendous and has no tests!

But it means the default response from our `/check` endpoint will make it a lot easier to assess the critical bit; the matching algorithm.

### DIFF
--- a/app/controllers/check/check.controller.js
+++ b/app/controllers/check/check.controller.js
@@ -5,18 +5,12 @@
  * @module CheckController
  */
 
-const Boom = require('@hapi/boom')
-
 const CheckTwoPartService = require('../../services/check/two-part.service.js')
 
 async function twoPart (request, h) {
-  try {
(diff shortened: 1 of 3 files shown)


Cruikshanks:
SonarCloud is perfectly right to call out the lack of unit tests for this one. But as I stated, I hand on heart promise this will be deleted before two-part tariff billing goes live! 😁

PR #360, opened 2023-08-16
Get Return Status as part of Matching Algorithm

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4081

As a billing and data user, I need the status of the returns to be assessed.

Once the charge element is matched to a return, the next step is to assess the status of the return, so that the relevant volumes and status can be recorded or error recorded.

This PR adds a new `returnStatus` array to the `chargeVersions` object displaying the unique statuses of the returns matched to the charge version.


Cruikshanks (changes requested):
Just a really minor tweak if its ok @Jozzey

Cruikshanks (approved):

PR #375, opened 2023-08-24
Add automatic payload cleaning

Cruikshanks:

We really should have added this when working on supplementary billing. But now we are building a new endpoint to handle changing the address for a billing account we really need it.

We'll get into why we need to build a new endpoint when we start adding the code for it. But there are a series of changes we need to make in preparation.

The first of these is to implement automatic payload cleaning. This is something we did in the [Charging Module API](https://github.com/DEFRA/sroc-charging-module-api). By cleaning payloads automatically when they come in, we avoid polluting the [Joi](https://joi.dev/) schemas and our controllers with lots of calls to `trim()` or checks to see if something is `null` or `undefined`.

By cleaning we mean removing empty, null or undefined properties and then removing extraneous whitespace from the values left. It also handles sanitizing values to protect us from things like [cross-site scripting](https://owasp.org/www-community/attacks/xss/).

With this in place, controllers can get on and use `request.payload` without having to do any of these other checks.


Jozzey (commented):
Since we are shackled to Hapi it looks like we may be able to achieve the same outcome using the `disinfect` plugin https://www.npmjs.com/package/disinfect

What do you think?

Cruikshanks:
Ok @Jozzey . As discussed I took a look at **Disinfect** in https://github.com/DEFRA/water-abstraction-system/pull/377 . I could have lived with still having to trim whitespace from values. But the incorrect translation of integers and some boolean values to strings means we need to pass on it.

> Pause whilst I go to check what package it uses for sanitizing to see if we could use it

😱😱😱😱😱

Right! Definitely passing on **Disinfect** because guess what it uses? [sanitizer](https://github.com/theSmaw/Caja-HTML-Sanitizer) !!!!

Jozzey (approved):

PR #392, opened 2023-08-30
Create `FetchUserRolesAndGroupsService`

StuAA78:
https://eaflood.atlassian.net/browse/WATER-4085

We create a service `FetchUserRolesAndGroupsService` to look up a user in the `idm` schema and return the roles and groups assigned to them.

Roles are assigned as follows:
- Users can have roles assigned to them in the `user_roles` table
- Users can have groups assigned to them in the `user_groups` table
  - Groups have one or more roles assigned to them in the `group_roles` table

We want to collate all roles, no matter how they're assigned, and return the resulting list, along with the groups the user belongs to.


test/models/idm/group.model.test.js
```diff
@@ -19,16 +27,142 @@ describe('Group model', () => {
...
+      })
+
+      it('can successfully run a related query', async () => {
+        const query = await GroupModel.query()
+          .innerJoinRelated('users')
+
+        expect(query).to.exist()
+      })
+
+      it('can eager load the roles', async () => {
```
Cruikshanks, line 153:
```suggestion
      it('can eager load the users', async () => {
```

Cruikshanks (changes requested):
Just copy & paste typo. Other than that all gravy!

Cruikshanks (approved):

PR #395, opened 2023-08-30
Add new CHA create customer change presenter

Cruikshanks:

We are working on migrating the change billing account address functionality from the legacy code to this project. A key part of this is being able to tell the [Charging Module API](https://github.com/DEFRA/sroc-charging-module-api) about the change!

This change adds a new Charging Module presenter which handles converting the data collected by the UI during the change address journey into the format the Charging Module expects.


Cruikshanks:
@Beckyrose200 @Jozzey @StuAA78

PR #400, opened 2023-09-01
Add new Billing Account change address service

Cruikshanks:

We are working on migrating the change billing account address functionality from the legacy code to this project. This change is the final part of the backend logic!

In this change, we add the `ChangeAddressService`. It will be responsible for orchestrating the change address process; letting the [Charging Module API](https://github.com/DEFRA/sroc-charging-module-api) know about the change and updating our own DB.

It then needs to generate a response that the [water-abstraction-ui](https://github.com/DEFRA/water-abstraction-ui) will understand so it can complete the journey for the user. We're making those changes in [Send change address requests to system in future](https://github.com/DEFRA/water-abstraction-ui/pull/2423).


Cruikshanks:
Just need PR #404 to be merged then the tests should pass and this will be ready to go.

Cruikshanks:
@Jozzey @Beckyrose200 @StuAA78 This is a big one and I've broken out as much as I could into other PR's. So, if you need me to walk through it before you review do let me know.

app/services/billing-accounts/change-address.service.js
```diff
@@ -0,0 +1,340 @@
...
+
+async function _patchExistingInvoiceAccountAddressEndDate (trx, invoiceAccountId, timestamp) {
+  // The timestamp represents the current date and time we're making this change, i.e. today. So, the new invoice
+  // account address will start from today. To show that the old record is no longer current, we need to set its
+  // `endDate` to be today - 1 (yesterday). The following works it all out even if we're over a month or year boundary
+  // and no moment() in sight! Thanks to https://stackoverflow.com/a/1296374 for how to do this
+  const endDate = new Date()
+  endDate.setDate(timestamp.getDate() - 1)
+
+  return InvoiceAccountAddressModel.query(trx)
```
StuAA78, line 151: Do we need to use `return` here seeing as we don't use the return value?
```suggestion
  await InvoiceAccountAddressModel.query(trx)
```

app/services/billing-accounts/change-address.service.js
```diff
@@ -0,0 +1,340 @@
...
+ * (end date is null).
+ *
+ * The object we return has all 3 entities whether they were persisted or not. This gets passed back to the UI via
+ * the controller and is what it needs to then be able to redirect the user to the correct page.
+ *
+ * @param {Date} timestamp the timestamp to be used for any date created or updated values when persisting
+ * @param {module:InvoiceAccountModel} invoiceAccount the invoice (billing) account having its address changed
+ * @param {module:AddressModel} address the new address to be persisted (expected to be populated)
+ * @param {module:CompanyModel} company the new agent company to be persisted (not expected to be populated)
+ * @param {module:ContactModel} contact the new contact to be persisted (not expected to be populated)
```
StuAA78, line 112: It took me a little figuring out to realise that when we say "expected to be populated" and "not expected to be populated", it's because `agentCompany` and `contact` are optional when calling the service but we always transform the day we receive (empty or not) into `companyInstance` and `contactInstance`. A little note in the comments to note why those two may not be populated might be useful.

test/services/billing-accounts/change-address.service.test.js
```diff
@@ -0,0 +1,424 @@
+'use strict'
```
StuAA78, line 1: Just as a general point, all of our tests call `ChangeAddressService.go()` with `agentCompany` and `contact`. Presumably some of these tests are passing in empty values but is it worth having at least one test that doesn't pass these in, just to confirm/document the defaulting of these to empty objects?

StuAA78 (changes requested):
Didn't have time for a full review but on a first pass it all looks good! Just a few minor comments 😁

app/services/billing-accounts/change-address.service.js
Cruikshanks: Agreed, but think I need to have another convo with you, the team Node maestro 🤓, about what we should be doing in these cases.

app/services/billing-accounts/change-address.service.js
Cruikshanks: Added extra comments. Hopefully that helps.

test/services/billing-accounts/change-address.service.test.js
Cruikshanks: Added a test and it made me think about what the service returns. So, also helped me clean things up a bit. Thank you!

app/services/billing-accounts/change-address.service.js
```diff
@@ -0,0 +1,371 @@
...
+/**
+ * Persist the company (Agent) entered during the change address process
+ *
+ * If the company has a `companyId:` we assume it was an existing company selected during the journey. So, the company
+ * is already persisted hence we just return `company`. If the company name is not set then the user has opted not to
+ * change the agent in the journey. So, we just return the empty `CompanyModel` instance.
+ *
+ * Else we attempt to insert a new company record. If the company has a `companyNumber:` it will either be an existing
+ * company record selected by the user, or they will have been required to enter the company number. The previous team
+ * added a unique constraint on `company_number` in the table so we cannot insert 2 records records with matching
```
Beckyrose200, line 247: It might just be my bad English skills but do we need records twice in this sentence?

test/controllers/billing-accounts.controller.test.js
```diff
@@ -8,10 +8,22 @@ const Sinon = require('sinon')
 const { describe, it, beforeEach, afterEach } = exports.lab = Lab.script()
 const { expect } = Code
 
+// Things we need to stub
+const ChangeAddressService = require('../../app/services/billing-accounts/change-address.service.js')
```
Beckyrose200, line 12: Does the order need to be alphabetical for the 'things we need to stub' list? :)

Beckyrose200 (commented):
Looks good from what I can tell, added a few comments :D

test/controllers/billing-accounts.controller.test.js
Cruikshanks: Yes!

test/services/billing-accounts/change-address.service.test.js
```diff
@@ -0,0 +1,441 @@
...
+const { expect } = Code
+
+// Test helpers
+const AddressHelper = require('../../support/helpers/crm-v2/address.helper.js')
+const AddressModel = require('../../../app/models/crm-v2/address.model.js')
+const CompanyHelper = require('../../support/helpers/crm-v2/company.helper.js')
+const CompanyModel = require('../../../app/models/crm-v2/company.model.js')
+const ContactModel = require('../../../app/models/crm-v2/contact.model.js')
+const DatabaseHelper = require('../../support/helpers/database.helper.js')
+const InvoiceAccountHelper = require('../../support/helpers/crm-v2/invoice-account.helper.js')
```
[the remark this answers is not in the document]
Cruikshanks: Yes!

Jozzey (approved):

PR #376, opened 2023-08-24
Add missing information to `/health/info` page

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4099

Finish off the work previously started on the `/health/info` page. The data that is missing that needs to be added in this PR is:
- Confirm Redis is running
- Display the latest results for the import app

The changes to add the Redis info have been moved into a separate PR https://github.com/DEFRA/water-abstraction-system/pull/402 so that this PR is just focusing on the import information.

PR #411, opened 2023-09-06
Fix `/health/info` Redis connection

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4099

After adding the missing information to the `/health/info` page in PR https://github.com/DEFRA/water-abstraction-system/pull/376 It was discovered that whilst it worked fine locally, when viewed on the `dev` server Redis was unable to connect. This was down the Redis connection on the AWS servers needing TLS to be enabled.

This PR will enable TLS by default for the Redis connection & add an environment variable to enable TLS to be disabled locally.

Whilst debugging this issue it was also found that if a connection to Redis could not be established the page would error rather than just display a connection failure message for Redis and continue to show the status of the other services. This behaviour was not present locally. After much hacking, a fix has not be found for this issue. Given we have more urgent work currently this issue will be added to our issues backlog to investigate further at a later date.


Cruikshanks (approved):

Jozzey:
PR linked to this issue https://github.com/DEFRA/water-abstraction-team/issues/54

PR #430, opened 2023-09-22
Investigate switching to jest

manabshy:
https://eaflood.atlassian.net/browse/WATER-4144

For those not aware, the dev team are now doing all new work in water-abstraction-system. This is how we are side-stepping the issues of working with the legacy code base and moving towards a simpler, better future!

When putting together the project we selected Hapi Lab for the unit test framework. The main framework for all Defra Node-based projects is Hapi. So, early Defra projects also selected Lab for unit testing. It was also the framework for Charging Module API so the dev team were familiar with it.

But it has limitations. Other Node unit test frameworks include more functionality that we have to add to Lab ourselves. Documentation as well as online examples are limited. Plus we’ve been keeping an eye on other Defra projects. Nowadays Jest is the framework of choice. All the Future Farming and Rod Licencing repos use it plus anything new we’ve seen appear on GitHub. It probably helps that Jest is the most popular unit test framework for Node. Certainly, most contract devs who join the team are likely to have experience with it. We’d be surprised if they’d even heard of Lab!

So, before we get too far down the path of our bright and shiny future now is the time to look at what it would take to switch to Jest.


Cruikshanks:
Superseded by #3489

PR #436, opened 2023-09-27
Allocate returns volumes to charge elements

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4127

This PR will take the returns that have been matched to charge elements and distribute their calculated volumes amongst the matched charge elements.


Jozzey:
SonarCloud is failing due to a lack of unit tests. It's been agreed that while we are figuring out the 2PT process we won't spend time on the unit tests as they will probably need to be completely re-written once the process is finalised and refactored.

Cruikshanks (approved):

PR #438, opened 2023-09-28
Even more jest

Cruikshanks:
Some ideas about [Investigate switching to jest](https://github.com/DEFRA/water-abstraction-system/pull/430) to discuss before applying to that branch.


Cruikshanks:
These changes got merged into #430

PR #442, opened 2023-10-02
Fix issue allocating return volume

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4127

A issue has been spotted in the PR which deals with allocating the Returns volumes to the Charge Elements which was merged in https://github.com/DEFRA/water-abstraction-system/pull/436

The issue is that if the return volume exactly matches the `authorisedAnnualQuantity` the volume isn't allocated to that Charge Elelment. This PR will fix that bug.


Jozzey:
It worked after all so didn't need to do this :poop:

PR #443, opened 2023-10-02
Remove licences with charge elements with approaching time limits from billing

Jozzey:
https://eaflood.atlassian.net/browse/WATER-3486

The purpose of this PR is to build an endpoint, that when hit will remove from billing licences with charge elements that have time limits which are coming to an end so that incorrect invoices are not sent.


test/controllers/charge-elements.controller.test.js
```diff
@@ -0,0 +1,68 @@
...
+          Sinon.stub(AddLicencesToWorkflowService, 'go').rejects()
+        })
+
+        it('returns a 500 status', async () => {
+          const response = await server.inject(options)
+
+          expect(response.statusCode).to.equal(500)
+        })
+      })
+    })
```
Cruikshanks, line 66: This test needs to be removed as it's not really valid. We have a `try/catch` in `AddLicencesToWorkflowService`, which means if it did throw an error the controller would never know about it.

test/services/charge-elements/add-licences-to-workflow.service.test.js
```diff
@@ -0,0 +1,97 @@
...
+  beforeEach(() => {
+    // The service depends on GlobalNotifier to have been set. This happens in app/plugins/global-notifier.plugin.js
+    // when the app starts up and the plugin is registered. As we're not creating an instance of Hapi server in this
+    // test we recreate the condition by setting it directly with our own stub
+    notifierStub = { omfg: Sinon.stub() }
+    global.GlobalNotifier = notifierStub
+  })
+
+  afterEach(async () => {
+    await DatabaseHelper.clean()
```
Cruikshanks, line 34: `await DatabaseHelper.clean()` needs to go in the `beforeEach()` not the `afterEach()`.

That way

- the test is always starting with a clean slate
- should the test fail the data remains for possible debugging purposes

app/services/charge-elements/fetch-licences-for-workflow.service.js
Cruikshanks, line 1: Might I suggest a different name? What we are interested in is licences with time limited elements (purposes). So, perhaps `FetchTimeLimitedLicencesService` / `fetch-time-limited-licences.service.js`. 🤷

app/services/charge-elements/add-licences-to-workflow.service.js
Cruikshanks, line 1: It _is_ adding licences to workflow, but it is doing this for a specific reason. So, might I suggest a different name? `ProcessTimeLimitedLicencesService` / `process-time-limited-licences.service.js` 🤷

app/controllers/charge-elements.controller.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
+
+/**
+ * Controller for /charge-elements endpoints
+ * @module ChargeElementsController
+ */
+
+const AddLicencesToWorkflowService = require('../services/charge-elements/add-licences-to-workflow.service.js')
+
+async function workflowTimeLimitedLicences (request, h) {
```
Cruikshanks, line 10: If we can, I favour naming the controller handler after the endpoint.

For example

```javascript
  path: '/health/airbrake',
  handler: HealthController.airbrake,

  path: '/health/database',
  handler: HealthController.database,

  path: '/health/info',
  handler: HealthController.info,
```

So, I'd suggest renaming this

```suggestion
async function timeLimited (request, h) {
```

Cruikshanks (changes requested):
Got some changes plus suggestions for naming. If grabbing me to pair will help get this over the line let me know.

test/services/charge-elements/add-licences-to-workflow.service.test.js
Jozzey: I think I was lazy and used another test as a template, so I suspect this issue also exists in other tests. I'll have a look and fix anything I find.

app/services/charge-elements/fetch-licences-for-workflow.service.js
Jozzey: 💯

app/services/charge-elements/add-licences-to-workflow.service.js
Jozzey: Sounds good to me 👍🏼

Cruikshanks (approved):

PR #447, opened 2023-10-04
Implement mock return review

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4152

We are currently working on the SROC two-part tariff bill. The first step of the process in the UI is to review how the returns have been matched. Our superstar design team are working on an improved design for the returns review screen but to do that they need data.

Enter our mock data service!

We previously [Create endpoint to generate fake data](https://github.com/DEFRA/water-abstraction-system/pull/347) which was focused on a bill run. The design team would love it if we could do the same for the data the current returns review page uses.

This change completes the work started in [Add new bill run volume model](https://github.com/DEFRA/water-abstraction-system/pull/440) to take an existing 2PT bill run and generate mock returns review data from it.

### DIFF
--- a/app/presenters/data/mock-returns-review.presenter.js
+++ b/app/presenters/data/mock-returns-review.presenter.js
@@ -1,0 +1,142 @@
+'use strict'
+
+/**
+ * Formats the response for the GET `/data/mock/{returns-review}` endpoint
+ * @module MockReturnsReviewPresenter
+ */
+
+const { capitalize, formatAbstractionPeriod, formatLongDate } = require('../base.presenter.js')
+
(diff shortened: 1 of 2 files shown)


Cruikshanks:
This was really useful to understand what the existing service is doing with the returns review screens. But with the new design almost complete and the fact it'll need the data in a different way, I'm no longer sure there is any value investing in tidying up and writing unit tests for this.

Just going to leave it here for a little while and 'wait-and-see'. Will then make a decision.

Cruikshanks:
Closing this based on the decision I've made to remove the mock data generation feature. See #575

PR #451, opened 2023-10-05
Add capitalize() method to base presenter

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4152

We are currently working on the SROC two-part tariff bill. The first step of the process in the UI is to review how the returns have been matched. Our superstar design team are working on an improved design for the returns review screen but to do that they need data.

This led us to start [Implement mock return review](https://github.com/DEFRA/water-abstraction-system/pull/447). A requirement that came out of that was the ability to capitalize certain values, for example, loss is recorded in the DB as `high` but when presenting it we want to see `High`.

As we start to build views and 'present' the data for them we're going to need this in more and more cases.

So, this change adds a `capitalize()` method to our base presenter.

### DIFF
--- a/app/presenters/base.presenter.js
+++ b/app/presenters/base.presenter.js
@@ -1,6 +1,25 @@
 'use strict'
 
 /**
+ * Capitalize the first letter of each word in a string
+ *
+ * Will work for strings containing multiple words or only one.
+ *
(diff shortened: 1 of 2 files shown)


Jozzey (approved):
👍🏼

PR #452, opened 2023-10-05
Make test data helpers more 'unique'

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4144

We've been investigating changing our unit test framework recently (see [Investigate switching to jest](https://github.com/DEFRA/water-abstraction-system/pull/430) and [water-abstraction-ava](https://github.com/DEFRA/water-abstraction-ava)).

A key issue we'll have if we switch is that most other frameworks run tests asynchronously. This means our pattern of clean-setup-assert won't work. At best we can clean before any tests run but then the test data we set up must not clash.

To achieve that we need to make the test data the helpers generate as unique as possible, which means moving away from hard-coded values.

We're still playing with frameworks but there is nothing stopping us from making this change now ahead of a decision.

So, this change covers making the helpers generate more unique and fixing any tests that break.


Cruikshanks:
This is bigger than I'd like it to be. But it made sense to do all the helpers at the same time, especially as the changes are generally consistent across them.

Happy to walk/talk through anything if anyone has questions.

test/services/billing/supplementary/fetch-previous-transactions.service.test.js
```diff
@@ -12,15 +12,22 @@ const DatabaseHelper = require('../../../support/helpers/database.helper.js')
 const BillRunHelper = require('../../../support/helpers/water/bill-run.helper.js')
```
Beckyrose200, line 12: Feel free to ignore since not a change you've made but the order for these isn't alphabetical.

test/services/billing/supplementary/fetch-previous-transactions.service.test.js
```diff
@@ -41,8 +48,9 @@ describe('Fetch Previous Transactions service', () => {
   describe('when there is a bill run', () => {
     describe('for the same licence and invoice account', () => {
       beforeEach(async () => {
-        const billingInvoiceLicenceId = await _createBillRunAndBillAndBillLicence(invoiceAccountId, licenceId)
-        await TransactionHelper.add({ billingInvoiceLicenceId })
+        const billingInvoiceLicenceId = await _createBillRunAndBillAndBillLicence(billRunSetupValues)
+        await TransactionHelper.add({ billingInvoiceLicenceId, chargeCategoryCode })
+        console.log('🚀 ~ file: fetch-previous-transactions.service.test.js:52 ~ beforeEach ~ billingInvoiceLicenceId:', billingInvoiceLicenceId)
```
Beckyrose200, line 53: Just checking is this console log meant to be here? :)

test/support/helpers/water/licence.helper.js
```diff
@@ -4,15 +4,17 @@
  * @module LicenceHelper
  */
 
+const { generateUUID } = require('../../../../app/lib/general.lib.js')
 const LicenceModel = require('../../../../app/models/water/licence.model.js')
+const { randomInteger } = require('../general.helper.js')
```
Beckyrose200, line 9: Think the order for these need changing :)

test/support/helpers/water/purpose.helper.js
```diff
@@ -5,13 +5,14 @@
  */
 
 const PurposeModel = require('../../../../app/models/water/purpose.model.js')
+const { randomInteger } = require('../general.helper.js')
```
Beckyrose200, line 8: Think the order for these need changing :)

Beckyrose200 (commented):
Just a few things I noticed, but looks good :D

Jozzey (changes requested):
Just spotted 1 minor infraction :grin:

Jozzey (approved):
👍🏼

PR #456, opened 2023-10-09
Create migration script, model, helper and unit tests for `licence_versions`

Jozzey:
https://eaflood.atlassian.net/browse/WATER-3486

Whilst working on the unit tests for the enhancement in this PR https://github.com/DEFRA/water-abstraction-system/pull/443 it was found that we would need the migration script, model, helper and unit tests for table `licence_versions`.

These will be created in this PR.

### DIFF
--- a/test/support/helpers/water/licence-version.helper.js
+++ b/test/support/helpers/water/licence-version.helper.js
@@ -4,7 +4,7 @@
  * @module LicenceVersionHelper
  */
 
-const { generateUUID } = require('../../../../app/lib/general.lib.js')
+const { generateUUID, timestampForPostgres } = require('../../../../app/lib/general.lib.js')
 const LicenceVersionModel = require('../../../../app/models/water/licence-version.model.js')
 
 /**
@@ -42,6 +42,8 @@ async function add (data = {}) {
  * @param {Object} [data] Any data you want to use instead of the defaults used here or in the database
  */
 function defaults (data = {}) {
+  const timestamp = timestampForPostgres()
+
   const defaults = {
     licenceId: generateUUID(),
     issue: 1,
@@ -49,8 +51,8 @@ function defaults (data = {}) {
     status: 'current',
     startDate: new Date('2022-01-01'),
     externalId: '9:99999:1:0',
-    createdAt: new Date(),
-    updatedAt: new Date()
+    createdAt: timestamp,
+    updatedAt: timestamp
   }
 
   return {
--- a/test/support/helpers/water/licence-version.helper.js
+++ b/test/support/helpers/water/licence-version.helper.js
@@ -23,7 +23,7 @@ const LicenceVersionModel = require('../../../../app/models/water/licence-versio
  *
  * @param {Object} [data] Any data you want to use instead of the defaults used here or in the database
  *
- * @returns {module:LicenceModel} The instance of the newly created record
+ * @returns {module:LicenceVersionModel} The instance of the newly created record
  */
 async function add (data = {}) {
   const insertData = defaults(data)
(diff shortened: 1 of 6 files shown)


test/support/helpers/water/licence-version.helper.js
```diff
@@ -0,0 +1,65 @@
...
+async function add (data = {}) {
+  const insertData = defaults(data)
+
+  return LicenceVersionModel.query()
+    .insert({ ...insertData })
+    .returning('*')
+}
+
+/**
+ * Returns the defaults used when creating a new licence version
```
Beckyrose200, line 37: https://github.com/DEFRA/water-abstraction-system/commit/1f972f33f9accb029de6393d4fcf177ca98ce772#diff-327dd98523aaf70b0b601aa777be715df63cb4922a9061e29006f4512ce60c3d

I think in one of Alan's recent PR's we are removing anything after 'Returns the defaults used'

Beckyrose200 (commented):
Looks good, just 1 comment form me :)

test/support/helpers/water/licence-version.helper.js
Jozzey: Alan made the default data more random to help with the Jest switch. He hasn't removed it though from what I can tell. Is that right @Cruikshanks ?

Beckyrose200: Sorry not the actual data just the comment on line 37 'Returns the defaults used when creating a new licence version'

Cruikshanks: Nope,@Beckyrose200 is right. I used the fact I needed to touch most helpers to make the comment generic and remove the possibility of copy-and-paste errors creeping in.

Jozzey: Your comment has ended up with me spotting an issue with the helper though. We have a timestamp method Alan created to generate the `createdAt` and `updatedAt` that I'm not using. I'll add that

Jozzey: Cheers @Beckyrose200 I've sorted it out now. Thought you were saying to remove the defaults 😀

test/support/helpers/water/licence-version.helper.js
```diff
@@ -0,0 +1,67 @@
...
+function defaults (data = {}) {
+  const timestamp = timestampForPostgres()
+
+  const defaults = {
+    licenceId: generateUUID(),
+    issue: 1,
+    increment: 0,
+    status: 'current',
+    startDate: new Date('2022-01-01'),
+    externalId: '9:99999:1:0',
```
Cruikshanks, line 53: From my experience making the data generic, we are probably going to need to generate this value rather than using the same value for every record.

I'd suggest taking a look at `test/support/helpers/returns/return.helper.js` for an example of how licence references are generated and apply something similar here.

Cruikshanks (changes requested):
Think we'll need a final tweak before this gets merged 😁

Cruikshanks (changes requested):

test/support/helpers/water/licence-version.helper.js
```diff
@@ -0,0 +1,68 @@
...
+ * Add a new licence version
+ *
+ * If no `data` is provided, default values will be used. These are
+ *
+ * - `licenceId` - [random UUID]
+ * - `issue` - 1
+ * - `increment` - 0
+ * - `status` - 'current'
+ * - `startDate` - new Date('2022-01-01')
+ * - `externalId` - '9:99999:1:0'
```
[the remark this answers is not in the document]
Jozzey: Doh!

Jozzey:
I'll get there eventually 😄

PR #458, opened 2023-10-10
Implement alternate version of 2PT matching

Cruikshanks:

We've had a number of PRs merged and changes made building up a new SROC two-part tariff matching engine.

We need to match returns to licences, via charge versions, in order to generate the two-part tariff bills for them. We're not using the legacy code as a reference for 2 reasons.

- it works in the context of matching 'returns -> charging'. We're being asked to use 'charging -> returns'
- it's rubbish 😬

So, as a team, we're having to build up our knowledge of how the matching should work and the engine itself as we go. However, the latest tranche of stories has asked us to deal with the scenario of allocating a return _across_ multiple charge references. This threw a spanner in the works because our current engine starts by matching returns to charge references. If a return is matched to more than one charge reference our current engine won't handle it.

Out of this, we've had further discussions about what it is the engine actually needs to do. Added to that we have a clearer idea of what the new returns review screens need in order to work. We've taken this information away, mulled it over, and iterated an alternate two-part tariff matching engine.

It has been built with these factors in mind

- returns must be matched at the charge element level
- a return is matched to a charge element by code (purpose) and overlapping abstraction period (it doesn't have to be exact)
- a return can match and be allocated to one or more elements
- we care about elements with no matches. We don't care about returns with no matches
- we want to know about issues with the element
- we want to know which returns were matched to an element
- a return should be matched to the charge reference with the highest subsistence charge first
- a return should be matched to the charge element with the highest authorised quantity first
- we match a return's lines to a charge element based on the charge element's abstraction period (does the line's start and end date overlap the element's abstraction period)

This isn't the final version of the engine. We've focused on making it work rather than clean, simple code and its output has been designed for testing purposes. But hopefully, this version covers all the scenarios we now know about and is better able to handle those we don't!


Cruikshanks:
All the changes needed from this PR have now been brought into main. We can close this spike PR down now.

PR #478, opened 2023-10-21
Implement nav bar for internal users

Cruikshanks:

So, pass-through authentication should be all sorted. We can now read the cookie passed in by the legacy UI, and confirm the user is authenticated and authorised to access a page. We pass auth, user and scope details to the global context so all view templates have access to this information. We also have our first use of this; deciding whether to display the header links to **Change password** and **Sign out**.

The final piece that all our pages should handle is the nav bar. This is where internal users can switch between **Search**, **Bill runs** and **Manage** assuming they have the right permissions.

This change implements the nav bar.

<details>
<summary>Screenshot of nav bar</summary>

Screenshot 2023-10-21 at 13 08 38

</details>


Beckyrose200 (approved):

client/sass/application.scss
```diff
@@ -1,3 +1,63 @@
 $govuk-global-styles: true;
 
 @import "node_modules/govuk-frontend/govuk/all";
+
```
Jozzey, line 4: Assume we have added some custom styling to the nav bar as there is nothing suitable in the `govuk-frontend`?

Jozzey (approved):

client/sass/application.scss
Cruikshanks: You got it! From speaking with our designer the design for this came from the design system itself, but on their docs page 😁 

Screenshot 2023-10-25 at 16 30 34

It's not actually a component of the frontend. He'd like to update it to match what the design system has now but we'd need to wait until all pages were migrated to system.

PR #485, opened 2023-10-26
Add new format money with commas to presenters

Cruikshanks:

To support our work on building a new bill page we need to display money values to the user in the UI for the first time. We've done stuff in our API endpoints before hence we have `formatNumberAsMoney()` in `app/presenters/base.presenter.js`.

But this will only take numbers like `1599.4` and `15350` and ensure they have 2 decimal places; `1599.40` and `15350.00`. For the UI we need to be _even fancier_!

In the UI we need to delineate the hundreds and thousands i.e. stick in commas. So, using our examples we need to display them as `1,599.40` and `15,350.00`.

The legacy [water-abstraction-ui](https://github.com/DEFRA/water-abstraction-ui) of course does this with a package because you can never have too many! 😜 [comma-number](https://github.com/elidoran/comma-number) if you are interested.

But as with most things we can do this with just the tools Node.js and JavaScript provides. So, this change adds a new `formatNumberAsMoneyWithCommas()` function to our base presenter to do the same thing.

### DIFF
--- a/test/presenters/base.presenter.test.js
+++ b/test/presenters/base.presenter.test.js
@@ -171,7 +171,7 @@ describe('Base presenter', () => {
 
     describe('when no £ symbol is requested', () => {
       it('correctly returns the value as a money string with no symbol, for example, 1149.50', async () => {
-        const result = BasePresenter.formatNumberAsMoney(valueInPence)
+        const result = BasePresenter.formatNumberAsMoney(valueInPounds)
 
         expect(result).to.equal('1149.50')
       })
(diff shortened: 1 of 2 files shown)


Jozzey (dismissed):

test/presenters/base.presenter.test.js
```diff
@@ -186,6 +186,26 @@ describe('Base presenter', () => {
     })
   })
 
+  describe('#formatNumberAsMoneyWithCommas()', () => {
+    const valueInPence = 1149.5
```
Beckyrose200, line 190: Should this be valueInPounds not pence? If it was pence it would be equal to 114950?

Cruikshanks: Doh!

Jozzey (approved):

PR #493, opened 2023-11-01
Add new fetch billing account service

Cruikshanks:

We are building our first proper page which will replace the bill summary page the [water-abstraction-ui](https://github.com/DEFRA/water-abstraction-ui) currently provides. Rather than try to show all licences _and_ all their transactions for a bill on one page we'll instead list the licences. Users can then see the transaction details by selecting a licence.

We've already made several other changes to support this. This change adds a new service we need to fetch details about the billing account for displaying on the page.


Jozzey (approved):
👍🏼

PR #501, opened 2023-11-06
Add new fetch bill licence service

Cruikshanks:

This adds a service that will get the data needed for the page. Mainly that is the transaction detail but also information about the bill and licence needed for context on the page.

One thing to note is the service makes no distinction between SROC and PRESROC. However, some linked entities and fields are only used when a transaction is one or the other.


Cruikshanks:
Need #502 merged before this can be merged.

Cruikshanks:
SonarCloud is complaining about duplication I _think_ (because it won't show where it thinks the duplication is!) with `app/services/billing/supplementary/fetch-previous-transactions.service.js`.

I'm fine with this as they are doing 2 separate things, and the duplication is with the query and not code (as such). We would overcomplicate things and/or impact performance by trying to write one query that would cover both scenarios just to make SonarCloud happy. Unfortunately, there isn't a way for me to tell SonarCloud to go away on this one 😁

test/services/bill-licences/fetch-bill-licence.service.test.js
```diff
@@ -0,0 +1,202 @@
...
+const { describe, it, beforeEach } = exports.lab = Lab.script()
+const { expect } = Code
+
+// Test helpers
+const BillHelper = require('../../support/helpers/water/bill.helper.js')
+const BillModel = require('../../../app/models/water/bill.model.js')
+const BillRunHelper = require('../../support/helpers/water/bill-run.helper.js')
+const BillRunModel = require('../../../app/models/water/bill-run.model.js')
+const BillLicenceHelper = require('../../support/helpers/water/bill-licence.helper.js')
+const BillLicenceModel = require('../../../app/models/water/bill-licence.model.js')
```
Beckyrose200, line 16: Think the bill-licence model and helper need to go above the bill-run model and helper

Cruikshanks: They do!

PR #502, opened 2023-11-06
Amend ChargeReferenceModel for PRESROC

Cruikshanks:

We are adding a [new fetch bill licence service](https://github.com/DEFRA/water-abstraction-system/pull/501) that will return data for both SROC and PRESROC transactions. Unfortunately, the previous team chose to store data related to the transactions differently. This means depending on what scheme a transaction is linked to you need to pull information from different places.

To support this we need to make a tweak to our `ChargeReferenceModel`. Under SROC a charge reference has multiple charge elements which in turn have multiple charge purposes. These purposes then link to a `PurposeModel`.

Under PRESROC the charge reference is _actually_ a charge element (hence the table name is `charge_elements`) and directly links to a `PurposeModel`.

So, this change updates the ChargeReferenceModel so that it _can_ pull through the linked purpose should one exist.

We also update the `PurposeModel` with the new relationship. Note, whilst doing this we spotted that the existing relationship to charge elements was wrong. It should be a `HasManyRelation` so we also use this opportunity to fix it.


Jozzey (approved):
👍🏼

PR #507, opened 2023-11-06
Add new Standard Charge Transaction presenter

Cruikshanks:

This adds a presenter that given a `TransactionModel` instance whose charge type is `standard` will return an object containing formatted details about the transaction.

The presenter handles both PRESROC and SROC transactions, altering the format of the output by what needs to be displayed on the page when the transaction is displayed.


Beckyrose200 (approved):
Looks good!

PR #520, opened 2023-11-14
Fix Airbrake when behind a proxy

Cruikshanks:
https://github.com/DEFRA/water-abstraction-team/issues/104

We copied the airbrake plugin we have from [sroc-charging-module-api](https://github.com/DEFRA/sroc-charging-module-api) where it works great. It also works great locally and we just assumed **water-abstraction-system** wasn't firing any errors in production. But we've realised no errors are getting through in our AWS environments because the apps sit behind a proxy and we haven't configured Airbrake to handle this!

In this change we follow the Airbrake docs and properly configure Airbrake (and yes - we have tested it this time!)

---

According to the [Airbrake docs](https://github.com/airbrake/airbrake-js/tree/master/packages/node#nodejs-request-and-proxy) the only way to do this is with the following.

> To use the [request](https://github.com/request/request) HTTP client, pass the request option which accepts a request wrapper:
>
> ```javascript
> const airbrake = new Notifier({
>   // ...
>   request: request.defaults({ proxy: 'http://localproxy.com' }),
> });
>

Unfortunately, Airbrake is relying on [request](https://github.com/request/request) as it has been deprecated since Feb 2020. But this directly impacts us because the only solution we've found is to also add **request** as a dependency. Normally, we've been able to `require()` sub-dependencies. But something about the way the Airbrake package is built means that's not the case here.

<details>
<summary>Screenshot of Errbit</summary>

Screenshot 2023-11-14 at 08 43 55

</details>

### DIFF
--- a/app/plugins/airbrake.plugin.js
+++ b/app/plugins/airbrake.plugin.js
@@ -6,7 +6,7 @@
  */
 
 /**
- * We use Airbrake to capture errors thrown within the service and send them tovan instance of Errbit we maintain in
+ * We use Airbrake to capture errors thrown within the service and send them to an instance of Errbit we maintain in
  * Defra.
  *
  * {@link https://hapi.dev/api/?v=20.0.0#-request-event}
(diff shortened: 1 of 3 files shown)


Cruikshanks:
SonarCloud is going to complain about test coverage, which is understandable as we don't yet have unit tests for the Airbrake plugin in _any_ repo. Unfortunately, it is a bugger to test so that situation ain't gonna change for the foreseeable future 😬

Cruikshanks:
Ah man! I hadn't expected Snyk to also highlight issues in **request**. 😩 😭 

_We_ are only using **request** to call `.defaults()` to get us the object that Airbrake needs to work behind a proxy.

The priority is to get error logging working so we're just going to have to accept this for now. But I'm going to leave [Airbrake isn't working](https://github.com/DEFRA/water-abstraction-team/issues/104) open until we have found a way around adding **request** as a dependency.

Jozzey (approved):

PR #521, opened 2023-11-14
Remove reporting from /health/info

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4169

Three charging forecast reports were built by the previous dev team. They are accessed from a link on the manage page in the internal UI. They have never been used because they were identified as returning incorrect values and have never been fixed for 'reasons'. They will never be fixed because all reporting work is being handled by our partnership with RDP and RSP.

This means [water-abstraction-reporting](https://github.com/DEFRA/water-abstraction-reporting) is a defunct app we can remove from the service. One of the first steps to allow this to happen is for the `/health/info` page which we use to check the health of the service to stop pinging reporting.

This change removes that check plus any associated config which was needed to support it.


Cruikshanks:
> While we are at it I think that we should also remove env var `REPORTING_URL` from the `.env.example` file.

Done!

Jozzey (approved):
👍🏼

PR #529, opened 2023-11-16
Refactor existing service to create a new one

Jozzey:

As part of the work we have been doing on two-part-tariff. We have found some code that we created for supplementary billing can be re-used for 2PT.

This PR will move the code that is responsible for determining the abstraction period from the `CalculateAuthorisedAndBillableDaysService`, and move it to a new service `DetermineAbstractionPeriodService`.

Once extracted out this function can then also be used by the 2PT process as well as supplementary.

### DIFF
--- a/app/services/bill-runs/determine-abstraction-periods.service.js
+++ b/app/services/bill-runs/determine-abstraction-periods.service.js
@@ -1,7 +1,7 @@
 'use strict'
 
 /**
- * Determine the abstraction period
+ * Determine real abstraction periods with years based on a reference period
  * @module DetermineAbstractionPeriodService
  */
 
--- a/app/services/bill-runs/determine-abstraction-periods.service.js
+++ b/app/services/bill-runs/determine-abstraction-periods.service.js
@@ -37,7 +37,6 @@
  * results.
  *
  * @param {Object} referencePeriod either the billing period or charge period
- * @param {module:ChargeElementModel} chargeElement holds the abstraction start and end day and month values
  * @param {Number} startDay the abstraction start day value
  * @param {Number} startMonth the abstraction start month value
  * @param {Number} endDay the abstraction end day value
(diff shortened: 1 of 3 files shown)


app/services/bill-runs/determine-abstraction-periods.service.js
```diff
@@ -0,0 +1,116 @@
+'use strict'
+
+/**
+ * Determine the abstraction period
```
Cruikshanks, line 4:
```suggestion
 * Determine real abstraction periods with years based on a reference period
```

app/services/bill-runs/determine-abstraction-periods.service.js
```diff
@@ -0,0 +1,116 @@
...
+ * @module DetermineAbstractionPeriodService
+ */
+
+/**
+ * Calculate from a charge element's abstraction data the relevant abstraction periods
+ *
+ * Before we can calculate the days and whether a period should be considered, we have to assign actual years to the
+ * charge element's abstraction start and end values.
```
Cruikshanks, line 12:
```suggestion
 * Determine real abstraction periods with years based on a reference period
 *
 * Abstraction periods on both charge elements and returns are held as just days and months, for example, '1st May to 31
 * Oct'. When we come to calculate billable days or attempt to match a return line to an element, we need to translate
 * them into real dates to do so.
 *
 * We use a reference period, typically a billing or charge period to determine what years to use. For example if the
 * `referencePeriod` is a billing period, it will be '1 April 2023 to 31 March 2024'. This service will then apply 2023
 * and 2024 to the provided abstraction period depending on whether it is an "in-year" or "out-year" abstraction period.
```

test/services/bill-runs/determine-abstraction-periods.service.test.js
```diff
@@ -0,0 +1,273 @@
...
+
+  describe('when the abstraction period is 01-JAN to 31-DEC (in-year)', () => {
+    beforeEach(() => {
+      startDay = 1
+      startMonth = 1
+      endDay = 31
+      endMonth = 12
+    })
+
+    describe('and the reference period is 01-APR-2022 to 31-MAR-2022 (starts first year, ends second year)', () => {
```
Cruikshanks, line 37:
```suggestion
    describe('and the reference period is 01-APR-2022 to 31-MAR-2023 (starts first year, ends second year)', () => {
```

Cruikshanks (changes requested):
Just a second opinion on the comments and a typo. Great stuff!

Cruikshanks (approved):

PR #533, opened 2023-11-20
Build models for `returns` based on new Views

Jozzey:
https://github.com/DEFRA/water-abstraction-team/issues/106

As part of the work we have been doing to get migrations working in [water-abstraction-system](https://github.com/DEFRA/water-abstraction-system) we are going to be creating all our new tables in a single schema `public`.

We have also decided that when there is a legacy table that we are still going to need for SROC. We are going to create a View of that table in the `public` schema and correct any issues with naming conventions, unused fields etc in the view. The first of these views that represent the old `returns` tables have been created in https://github.com/DEFRA/water-abstraction-system/pull/531

This PR will create the models, helpers and unit tests required to be able to start to use these new views.


Cruikshanks (changes requested):
One typo, but then a suggested change you wouldn't have to do if 'yours-truly' had done the job properly in the first place!

Cruikshanks (approved):

PR #539, opened 2023-11-22
Remove home.template

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4131

This was added way, way back when we were getting to grips with Nunjucks, vision and having Hapi render our pages. Should have been removed at the time but forgot and came across it again as part of building the new billing pages.


Jozzey (approved):
👍🏼

PR #542, opened 2023-11-22
Add new bill summaries presenter

Cruikshanks:

This presenter has been added to format the summarised data for each bill linked to a selected bill run. The data will be used in a new bill run page that will replace the existing one in the legacy service.

The page displays top-level details about the bill run (date created type, region etc) and then a summary for each bill linked to the bill run. This presenter handles the bill summary information returned by `FetchBillRunService`.

If a bill run contains bills linked to both water companies and other abstractors it will display them in 2 separate tables. If only one type of bill is present just a single table is shown. This is why the presenter takes the bills and groups them. The Nunjucks template will then decide whether to show just one table or both with captions based on how many groups this presenter returns.


Jozzey (dismissed):

Cruikshanks:
Don't merge until the refactoring work to use the new models is complete

PR #544, opened 2023-11-22
Add new view bill run endpoint and template

Cruikshanks:

We have been working on implementing a new version of the bill page. The current one tries to display all transaction details for all licences on a single page when the bill is selected.

In the larger annual bill runs this will equate to more than 150 licences and their transactions, which means the page is failing to load.

All that work has been done in this repo as part of our ongoing efforts to migrate away from the legacy code. It is in the final throws of testing and will soon be released.

To get to the new pages you have to pass through the existing bill run summary page. What our work has highlighted is this page leaves a lot to desire! Again, the larger the bill run the slower it is, noticeably so. Though it at least always loads 🧌.

It also is inconsistent. It has a different layout depending on the bill run type, and one of those depends on tabs, which is an incorrect use of the component (and 🤢).

So, we want to remove this remaining 'bump' in our billing user journey and replace it with a 'better' one. This is the final change in a series to support that, which adds the new template and endpoint.


Cruikshanks:
Work is still ongoing on this. We have found an issue with the counts on the page but the legacy view also seems to show confusing numbers!

So, we're just waiting on design to come up with an agreed solution which is going to mean changes are needed.

Cruikshanks:
> Work is still ongoing on this. We have found an issue with the counts on the page but the legacy view also seems to show confusing numbers!
> 
> So, we're just waiting on design to come up with an agreed solution which is going to mean changes are needed.

I'm going to implement the solution to the count issue in a separate PR.

PR #546, opened 2023-11-23
Fetch charge versions service

Beckyrose200:

This change is part of the work we have been doing on two-part-tariff. Introducing the FetchChargeVersionsService. This service will take a region code and a billing period and will return all the charge versions with there associated licences, charge references and elements. This is based on the charge version dates crossing over the billing period dates. The charge version being a status of current, the charge versions having an sroc scheme and it being in the region that was given to the service.


Beckyrose200:
Just a comment to add that I have taken out the ability to check an individual licence in this service.

Jozzey (changes requested):

Jozzey (dismissed):

Jozzey (approved):

PR #547, opened 2023-11-23
Create `fetch-returns-for-licence` service

Jozzey:

This PR will extract the `fetch-returns-for-licence` service out of our hacky branch https://github.com/DEFRA/water-abstraction-system/pull/458 and into its own PR along with the unit tests.


test/services/bill-runs/two-part-tariff/fetch-returns-for-licence.service.test.js
Cruikshanks, line 1: This is great! But you have done such a good job covering the possible scenarios I think it makes clear there is one missing; when there are no returns with a matching licence ref.

What do you think? Should we add one more test to confirm the results are empty?

Cruikshanks (changes requested):
I think we could do with one more test. But happy to chat about it.

test/services/bill-runs/two-part-tariff/fetch-returns-for-licence.service.test.js
Jozzey: You're right, totally overlooked that one. I'll pop one in.

Cruikshanks (approved):

PR #548, opened 2023-11-23
Adding a periods overlap helper

Beckyrose200:

This PR is focused on adding a helper method for the two-part-tariff work. The helper method is used to determine if two dates it has been given overlap or not.


app/lib/general.lib.js
```diff
@@ -27,6 +27,30 @@ function generateUUID () {
   return randomUUID({ disableEntropyCache: true })
 }
 
+/**
+ * Checks if any of the periods in checkPeriods overlap with any period in referencePeriods
+ * @param {*} referencePeriods Array of objects representing periods with startDate and endDate properties
+ * @param {*} checkPeriods Array of objects representing periods to check with startDate and endDate properties
+ * @returns Returns true if there is an overlap, otherwise false
```
Cruikshanks, line 34: I think there are a few things we can tidy up in the comments

```suggestion
 * Tests if one set of periods (represented by a start and end date) overlaps with another
 *
 * Added as part of two-part tariff and the need to match returns and lines to charge elements. A common complication in
 * WRLS is the need to convert an abstract period, for example '1 Nov to 31 Mar' to a concrete period (2023-11-01 to
 * 2024-03-31). It gets even more complex when the period crosses over another, for example the start or end of a
 * billing period. Then the only way to represent the abstract period in a usable way is as 2 separate periods. Hence
 * this service deals with arrays of periods.
 * 
 * > See the comments for `DetermineAbstractionPeriodService` to better understand the complexity of going from
 * > abstract to concrete periods
 * 
 * Then there are times we need to test if the periods of one thing overlap with another. In two-part tariff that's the
 * abstraction periods of a charge element with those of a return. If _any_ of the periods overlap then the return is
 * 'matched' and can be allocated to the charge element.
 * 
 * This method iterates through the `referencePeriods`. It then filters the `checkPeriods` by testing if any of them
 * overlap with the `referencePeriod`. If any do `checkPeriods.filter()` will return a non-empty array at which point
 * `periodsOverlap()` will return `true`.
 * 
 * Else, having compared all the `checkPeriods` against each `referencePeriod` and finding no overlaps the function will
 * return false.
 *
 * @param {Object[]} referencePeriods Each period is an object containing a `startDate` and `endDate` property
 * @param {Object[]} checkPeriods Each period is an object containing a `startDate` and `endDate` property. These
 * periods will be checked against the `referencePeriods for any overlaps
 *
 * @returns {boolean} Returns true if there _any_ check period overlaps with a reference period, else false
```

test/lib/general.lib.test.js
```diff
@@ -48,4 +48,49 @@ describe('RequestLib', () => {
       expect(result).to.equal('2015-10-21T20:31:57.000Z')
     })
   })
+
+  describe('#periodsOverlap', () => {
+    let referencePeriod
+    let checkPeriod
+
+    describe('when given dates that do not overlap', () => {
```
Cruikshanks, line 56: Because we name the params `fooPeriods` I'd suggest tweaking the `describe`

```suggestion
    describe('when given periods that do not overlap', () => {
```

test/lib/general.lib.test.js
```diff
@@ -48,4 +48,49 @@ describe('RequestLib', () => {
...
+      })
+
+      it('returns false', () => {
+        const result = GeneralLib.periodsOverlap(referencePeriod, checkPeriod)
+
+        expect(result).to.equal(false)
+      })
+    })
+
+    describe('when given dates that do overlap', () => {
```
Cruikshanks, line 76: Because we name the params `fooPeriods` I'd suggest tweaking the `describe`

```suggestion
    describe('when given periods that do overlap', () => {
```

test/lib/general.lib.test.js
```diff
@@ -48,4 +48,49 @@ describe('RequestLib', () => {
       expect(result).to.equal('2015-10-21T20:31:57.000Z')
     })
   })
+
+  describe('#periodsOverlap', () => {
```
Cruikshanks, line 52: I think to cover all bases there are a few more scenarios we could add that would hopefully answer any questions someone had about what this function does in certain scenarios.

These are my suggestions for the scenarios to cover

- when there is no overlap (you've already got this)
  - `referencePeriod = [{ startDate: new Date('2023-02-01'), endDate: new Date('2023-02-28') }]`
  - `checkPeriod = [{ startDate: new Date('2023-01-01'), endDate: new Date('2023-01-31') }]`
- when a check period overlaps the start of a reference period
  - `referencePeriod = [{ startDate: new Date('2023-02-01'), endDate: new Date('2023-02-28') }]`
  - `checkPeriod = [{ startDate: new Date('2023-01-15'), endDate: new Date('2023-02-15') }]`
- when a check period overlaps the end of a reference period
  - `referencePeriod = [{ startDate: new Date('2023-01-01'), endDate: new Date('2023-01-31') }]`
  - `checkPeriod = [{ startDate: new Date('2023-01-15'), endDate: new Date('2023-02-15') }]`
- when a reference period is _completely_ inside a check period
  - `referencePeriod = [{ startDate: new Date('2023-02-01'), endDate: new Date('2023-02-15') }]`
  - `checkPeriod = [{ startDate: new Date('2023-01-01'), endDate: new Date('2023-02-28') }]`
- when a check period is _completely_ inside a reference period
  - `referencePeriod = [{ startDate: new Date('2023-01-01'), endDate: new Date('2023-02-28') }]`
  - `checkPeriod = [{ startDate: new Date('2023-02-01'), endDate: new Date('2023-02-15') }]`
- when the periods are the same
  - `referencePeriod = [{ startDate: new Date('2023-02-01'), endDate: new Date('2023-02-28') }]`
  - `checkPeriod = [{ startDate: new Date('2023-02-01'), endDate: new Date('2023-02-28') }]`

Cruikshanks (changes requested):
I think we can expand the comments a bit more and fill in the details about the params (see [JSDoc comments](https://github.com/DEFRA/water-abstraction-team/blob/main/coding_conventions.md#jsdoc-comments) for notes/tips).

And though I know (or I think I do) what the function would do given certain scenarios I think it'd help to have them explicitly set out in the unit tests. I've provided the ones I can think of along with some suggested period dates to use.

Jozzey (approved):

Cruikshanks (approved):

PR #551, opened 2023-11-27
Create `water` schema views

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4188

As part of the work we have been doing on two-part tariff we are going to be creating all our new tables in a single schema `public`.

We have also decided that when there is a legacy table that we are still going to need for SROC. We are going to create a View of that table in the `public` schema and correct any issues with naming conventions, unused fields etc in the view. The first of these views that represents the old `returns` tables have been created in https://github.com/DEFRA/water-abstraction-system/pull/531

In this PR we will build the migration scripts to create the Views we need for SROC in the `water` schema.

PR #560, opened 2023-12-01
Create `water` schema models & helpers

Jozzey:

As part of the work we have been doing on two-part tariff we are going to be creating all our new tables in the default `public` schema.

We have also decided that when there is a legacy table that we are still going to need we will create a [View](https://www.postgresql.org/docs/current/sql-createview.html) of it in the `public` schema. This allows us to correct any issues with naming conventions, strip out unused fields, and join entities that are currently sat in different schemas. The first example of this approach was done in PR #531 .

This change adds the models and helpers for the views in the `water` schema that were created in PR https://github.com/DEFRA/water-abstraction-system/pull/551

> The final step will then be to refactor the existing code to use the new models and delete the old legacy ones.


Jozzey:
SonarCloud is complaining about duplicated lines. This is because until we switch to using our new models based on the views we effectively have 2 versions of each model, one based on the views and the other on the tables. This will be resolved once we have updated our code to use the new view-based models and removed the old ones.

PR #561, opened 2023-12-01
Create `crm_v2` schema models and helpers

Cruikshanks:

As part of the work we have been doing on two-part tariff we are going to be creating all our new tables in the default `public` schema.

We have also decided that when there is a legacy table that we are still going to need we will create a [View](https://www.postgresql.org/docs/current/sql-createview.html) of it in the `public` schema. This allows us to correct any issues with naming conventions, strip out unused fields, and join entities that are currently sat in different schemas. The first example of this approach was done in PR #531 .

This change adds the models and helpers for the views in the `crm_v2` schema that were created in PR https://github.com/DEFRA/water-abstraction-system/pull/556

> The final step will then be to refactor the existing code to use the new models and delete the old legacy ones.

### DIFF
--- a/app/models/return-log.model.js
+++ b/app/models/return-log.model.js
@@ -1,7 +1,7 @@
 'use strict'
 
 /**
- * Model for return_logs
+ * Model for return_logs (returns.returns)
  * @module ReturnLogModel
  */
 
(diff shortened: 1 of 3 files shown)


Cruikshanks:
SonarCloud is complaining about the duplication. But that is only to be expected until we can remove the legacy models.

Jozzey (dismissed):

PR #566, opened 2023-12-04
Refactor to use new view based CRM_V2 models

Cruikshanks:

As part of the work we have been doing on two-part tariff, we will be creating all our new tables in the default `public` schema.

We have also decided that when there is a legacy table that we are still going to need we will create a [View](https://www.postgresql.org/docs/current/sql-createview.html) of it in the `public` schema. This allows us to correct any issues with naming conventions, strip out unused fields, and join entities currently sat in different schemas. The first example of this approach was done in PR #531 .

We created the new views in [Create CRM_V2 schema views](https://github.com/DEFRA/water-abstraction-system/pull/556). Then in [Create CRM_V2 schema models and helpers](https://github.com/DEFRA/water-abstraction-system/pull/561) we added the new models and helpers that use them.

This is the final step in the process, we are refactoring any use of the legacy models to use the new ones and deleting all the legacy-based code.


Jozzey (changes requested):
Just the one typo

Jozzey (approved):

PR #573, opened 2023-12-07
Add no returns required page (return requirements)

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4248

This is the very first page in the return requirements setup journey when 'no returns required' is selected.


Demwunz (approved):
👍🏽 looks good to me

PR #574, opened 2023-12-07
Add Select the start date (return requirements)

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4249

This is the second  page in the return requirements setup journey when 'no returns required' is selected.


app/controllers/licences.controller.js
```diff
@@ -0,0 +1,19 @@
+'use strict'
+
```
Cruikshanks, line 2: It's become a convention for each module we create that we declare some JSDoc about it at the top. I'd suggest

```suggestion

/**
 * Controller for /licences endpoints
 * @module LicencesController
 */
 
```

Cruikshanks (changes requested):
I've just got one small request about sticking a module declaration at the top of the file. Who knows, one day it might be useful! 😁 

Other than that this is bang on. Welcome to the team! 👥

Cruikshanks (approved):

PR #577, opened 2023-12-11
Create `fetch-licences` service

Jozzey:

This PR will extract the `fetch-licences` service out of our hacky branch https://github.com/DEFRA/water-abstraction-system/pull/458 and into its own PR along with the unit tests.

### DIFF
--- a/app/services/bill-runs/two-part-tariff/fetch-licences.service.js
+++ b/app/services/bill-runs/two-part-tariff/fetch-licences.service.js
@@ -1,0 +1,64 @@
+'use strict'
+
+/**
+ * Fetches 2PT Licences for the matching region and billing period
+ * @module FetchLicencesService
+ */
+
+const FetchChargeVersionsService = require('./fetch-charge-versions.service.js')
+
+/**
+ *
+ * @param {*} regionId
+ * @param {*} billingPeriod
+ *
+ * @returns
+ */
+async function go (regionId, billingPeriod) {
+  const chargeVersions = await FetchChargeVersionsService.go(regionId, billingPeriod)
+
+  const uniqueLicenceIds = _extractUniqueLicenceIds(chargeVersions)
+
+  return _groupByLicence(chargeVersions, uniqueLicenceIds)
+}
+
+function _extractUniqueLicenceIds (chargeVersions) {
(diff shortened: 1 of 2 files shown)


app/services/bill-runs/two-part-tariff/fetch-licences.service.js
```diff
@@ -0,0 +1,66 @@
...
+const FetchChargeVersionsService = require('./fetch-charge-versions.service.js')
+
+/**
+ * Fetches 2PT Licences for the matching region and billing period plus the charge versions associated with them grouped
+ * by licence
+ *
+ * @param {*} regionId UUID of the region being billed that the licences must be linked to
+ * @param {*} billingPeriod Object with a `startDate` and `endDate` property representing the period being billed
+ *
+ * @returns {Object} Contains an array of unique licence IDs and array of charge versions to be processed
```
Cruikshanks, line 17: We can be explicit with the types in the JSDoc. Plus I'm not sure the return comment is correct anymore.

```suggestion
 * @param {String} regionId UUID of the region being billed that the licences must be linked to
 * @param {Object} billingPeriod Object with a `startDate` and `endDate` property representing the period being billed
 *
 * @returns {Object[]} the licences to be matched, each containing an array of charge versions applicable for two-part tariff
```

app/services/bill-runs/two-part-tariff/fetch-licences.service.js
```diff
@@ -0,0 +1,66 @@
...
+  for (let i = 0; i < uniqueLicenceIds.length; i++) {
+    const licenceId = uniqueLicenceIds[i]
+    const matchedChargeVersions = chargeVersions.filter((chargeVersion) => {
+      return chargeVersion.licence.id === licenceId
+    })
+
+    const { licenceRef, startDate, expiredDate, lapsedDate, revokedDate } = matchedChargeVersions[0].licence
+
+    licences[i] = {
+      licenceId,
```
Cruikshanks, line 51: I don't think we need to use the old ID name anymore

```suggestion
      id: licenceId,
```

Cruikshanks (changes requested):
Just some corrections that come from this code being around so long and having gone through so many iterations.

Cruikshanks (approved):

PR #581, opened 2023-12-12
Return `aggregate` as a float

Jozzey:

During QA it has been spotted that the value for the `aggregate` is being returned as a string rather than a float, as this is how it is stored in the JSONB field in the database.

In this PR the `aggregate` value will be converted to a float in the `fetch-charge-versions` service by using `castFloat` in the Objection query.


Beckyrose200 (approved):
👍🏻

PR #582, opened 2023-12-12
Update 2PT check endpoint with alternate changes

Cruikshanks:

Way, way back we built our very first iteration of a two-part tariff 'check' endpoint. The intent was to have some way for our QA team to begin verifying how we were matching and allocating returns correctly. Along the journey, we realised we needed to make a major shift in how things were working which has since become known as the [alternate engine](https://github.com/DEFRA/water-abstraction-system/pull/458).

This started as a 'spike' (does _this_ look right?), then was deployed on the branch to environments for double checking, was tweaked, tweaked a bit more, tweaked again, new requests catered for etc. _Now_ we're trying to figure out how we migrate from our hack to 'production-ready' unit-tested code, still in the same branch.

Recently, we lost the original alternate engine changes we made to the 'check' endpoint on that branch when we merged in `main`. That became the trigger to move out of the [alternate-2pt-engine](https://github.com/DEFRA/water-abstraction-system/tree/alternate-2pt-engine) branch those changes which are just related to the check endpoint and what the QA team need to keep an eye on things.

This should mean we won't lose those changes again. Also, the team focused on refactoring can differentiate their changes from stuff we've been doing to keep the QA's happy! 😁

So, this change updates the 'check' endpoint to be the version we've been using and testing against. This includes what was needed to get it to work with our recent model changes.

> When two-part tariff is completed, hopefully soon, we will delete all of this! 😈


Cruikshanks:
I need #586 to be approved and merged for the unit tests in this change to pass 😁

PR #586, opened 2023-12-13
Include volume in FetchChargeVersionsService

Cruikshanks:

This is related to our work building a new SROC two-part tariff matching and allocating engine.

We've been asked to make a change to the engine. Now when it allocates returns to charge elements it should take into account the authorised volume of the charge reference the charge element is under.

For example, if we had the following scenario

- `CR01 - authorised 100`
  - `CE01 - authorised 75`
  - `CE02 - authorised 75`
- `RT01 - submitted 150`

The original understanding was that we would be able to allocate all 150ML across the 2 charge elements. Now, when the total allocation across all charge elements within a charge reference reaches the charge references authorised we have to stop. So, the result would be

- `CR01 - authorised 100`
  - `CE01 - authorised 75 / allocated 75`
  - `CE02 - authorised 75 / allocated 25`
- `RT01 - submitted 150 / allocated 100`

To be able to do this we need to know what the authorised volume of the charge reference is. So, this change adds to what we get back when calling `FetchChargeVersionsService`.

### DIFF
--- a/test/services/bill-runs/two-part-tariff/fetch-charge-versions.service.test.js
+++ b/test/services/bill-runs/two-part-tariff/fetch-charge-versions.service.test.js
@@ -25,7 +25,6 @@ describe('Fetch Charge Versions service', () => {
     startDate: new Date('2022-04-01'),
     endDate: new Date('2023-03-31')
   }
-  const regionCode = 5
 
   let chargeCategoryId
   let licence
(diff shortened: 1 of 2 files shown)


test/services/bill-runs/two-part-tariff/fetch-charge-versions.service.test.js
```diff
@@ -15,403 +15,263 @@ const ChargeVersionHelper = require('../../../support/helpers/charge-version.hel
...
-          )
-
-          chargeCategory = ChargeCategoryHelper.add()
-
-          notInDateChargeReference = await ChargeReferenceHelper.add({
-            chargeVersionId: notInDateChargeVersion.id,
-            billingChargeCategoryId: chargeCategory.id,
-            adjustments: { s127: true, aggregate: 0.562114443 }
-          })
+    it('returns the charge versions order by licence reference', async () => {
```
Jozzey, line 160:
```suggestion
    it('returns the charge versions ordered by licence reference', async () => {
```

Jozzey (changes requested):
Just the one typo

Jozzey (approved):

PR #589, opened 2023-12-14
Prepare licence for allocation service

Beckyrose200:

This PR focuses on creating the PrepareLicenceForAllocationService along with its unit tests. This is a service within our two-part tariff sroc bill runs. This focuses on preparing the licence with its charge references for matching.


test/services/bill-runs/two-part-tariff/prepare-licences-for-allocation.service.test.js
Cruikshanks, line 1: So, we've got some things to cover in the unit tests. I think the problems come down to

- structure
- documentation VS testing

## Structure

First things first, I'd move all the test data to a function (`_testLicences()` ??) that sits at the bottom of the file. That way you can bypass the lengthy setup and get to the tests quickly. Take a look at `test/presenters/bills/view-bill.presenter.test.js` for an example of this.

Next, I'd move _all_ your test data into what that function returns. The test `it('sorts the charge references by subsistence charge')` adds a second charge reference that isn't going to change the behaviour of the service that the other tests are trying to cover.

That test also highlights a break with our convention. Where a test requires setup (Arrange) we do that in `before()/beforeEach()` blocks. The aim is that the first thing that happens in the test is calling our service (Act) followed by checking the results (Assert).

```javascript
  describe('when the music is valid', () => {
    let genre

    // Arrange
    beforeEach(() => {
      genre = 'synthwave'
    })

    it("gets a 'thumbs up'", () => {
      // Act
      const result = WillAlanPlayItService.go(genre)

      // Assert
      expect(result).to.equal('thumbs up')
    })
  })
````

> [The three A's of Unit Testing](https://dev.to/coderjay06/the-three-a-s-of-unit-testing-b22) is a quick read about Arrange, Act and Assert in unit tests

If you've been forced to put the arrangement inside your test it's a clue that something is out of whack with the current structure.

The next big clue is the output. This is the current output from the tests.

```
Prepare Licences For Allocation Service
  when given multiple licences and a billing period
    ✔ 447) sorts the charge references by subsistence charge (1 ms)
    ✔ 448) determines the charge period for each charge version (0 ms)
    fetches the relevant returns for that licence
      and prepares the returns for that licence
        ✔ 449) adds the abstraction period to the individual return (1 ms)
        ✔ 450) flags the return if abstraction has happened outside the abstraction period (0 ms)
        ✔ 451) flags if the return is a nil-return (0 ms)
        ✔ 452) sets the allocated quantity on the return to 0 (0 ms)
        ✔ 453) sets `matched` on the return to false (0 ms)
        ✔ 454) sets the quantity on the return to the sum of the returns lines (0 ms)
    prepares all the charge elements for that licence
      ✔ 455) does not prepare the charge element if it has a null startDate (0 ms)
      ✔ 456) sets the charge elements return logs to an empty array (1 ms)
      ✔ 457) sets the chare elements allocated quantity to 0 (0 ms)
      ✔ 458) sets the charge elements abstraction periods (1 ms)
  when given no licences and a billing period
    ✔ 459) does not throw an error (0 ms)
```

When you put the parts together they should be readable as a sentence/statement of what behaviour the test is checking. However, when I did that with this output a number of the statements no longer made sense.

- when given multiple licences and a billing period **it** sorts the charge references by subsistence charge
- when given multiple licences and a billing period **it** determines the charge period for each charge version
- when given multiple licences and a billing period fetches the relevant returns for that licence and prepares the returns for that licence **it** adds the abstraction period to the individual return
- when given multiple licences and a billing period fetches the relevant returns for that licence and prepares the returns for that licence **it** flags the return if abstraction has happened outside the abstraction period
- when given multiple licences and a billing period fetches the relevant returns for that licence and prepares the returns for that licence **it** flags if the return is a nil-return
- when given multiple licences and a billing period fetches the relevant returns for that licence and prepares the returns for that licence **it** sets the allocated quantity on the return to 0
- when given multiple licences and a billing period fetches the relevant returns for that licence and prepares the returns for that licence **it** sets `matched` on the return to false
- when given multiple licences and a billing period fetches the relevant returns for that licence and prepares the returns for that licence **it** sets the quantity on the return to the sum of the returns lines
- when given multiple licences and a billing period prepares all the charge elements for that licence **it** does not prepare the charge element if it has a null startDate
- when given multiple licences and a billing period prepares all the charge elements for that licence **it** sets the charge elements return logs to an empty array
- when given multiple licences and a billing period prepares all the charge elements for that licence **it** sets the chare elements allocated quantity to 0
- when given multiple licences and a billing period prepares all the charge elements for that licence **it** sets the charge elements abstraction periods
- when given no licences and a billing period **it** does not throw an error

## Documentation VS testing

I've confused things banging on about tests having multiple roles, one of which being documentation that is often overlooked or not considered. There are times when a test exists more to document a behaviour rather than to check it acts as expected. So for now, forget I ever mentioned documentation and focus on just testing the behaviour you can see in the thing under test. My starter for ten would be

- for each licence, when matching returns exist, that have submission lines inside the abstraction period, **it** preps the return correctly
- for each licence, when matching returns exist, that have submission lines outside the abstraction period, **it** preps the return and flags it as outside the abstraction period
- for each licence, when matching returns exist, that is a nil return, **it** preps the return and flags it as a nil return
- for each licence, **it** sorts the charge references by the subsistence charge
- for each licence, **it** preps the charge elements correctly

I can think of more, plus this is only my opinion on how to word/structure them. But you'll note I've not included tests that specify that the abstraction period has been set, or the allocation quantity. These are applied to all licences so they can be covered in other checks. And what the abstraction periods get set to is determined by other services that already have their own unit tests. This is that grey area of documentation VS test. Is there value in demonstrating how the billing period and the end dates on the licence affect the abstraction period that gets set? Possibly, but it also starts to confuse things. Hence for now I'd suggest sticking to black & white!

**Review by Cruikshanks — CHANGES_REQUESTED** [review]
More times than not it's the unit tests that prove harder than the actual code.

Rather than zero in on specifics, I've tried to give an expanded guide on structuring the unit tests and what to cover in them.

It'll be waffly (it is me 😁!) but hopefully it will help nail how to pull things together for this and future ones.

**Comment by Beckyrose200** [comment]
Sonar Cloud is complaining on 39% of the code being duplicated. This is a false positive and is due to the fact we are hacking away at the two-part tariff work, taking parts out and merging them in separate PR's.

File: app/services/bill-runs/two-part-tariff/prepare-licences-for-allocation.service.js [hunk 1]
```diff
@@ -0,0 +1,142 @@
...
+ * @module PrepareLicencesForAllocationService
+ */
+
+const DetermineAbstractionPeriodService = require('../determine-abstraction-periods.service.js')
+const DetermineChargePeriodService = require('../determine-charge-period.service.js')
+const FetchReturnLogsForLicenceService = require('./fetch-return-logs-for-licence.service.js')
+const { periodsOverlap } = require('../../../lib/general.lib.js')
+
+/**
+ * Prepares return logs and charge versions for matching and allocation based on provided licences and billing period
```
Cruikshanks, line 14:
```suggestion
 * For each licence finds returns for the billing period and prepares them and the licence's charge elements
 * ready to be matched and allocated to one another
 * 
 * > Rather than create a copy of each licence and amend it we amend the licences in place. Hence this service
 * > has no return value
```

app/services/bill-runs/two-part-tariff/prepare-licences-for-allocation.service.js
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * @param {Object[]} licences - The licences to prepare
+ * @param {Object[]} billingPeriod - The billing period
```
Cruikshanks, line 17:
```suggestion
 * @param {Object[]} billingPeriod - The period a bill run is being calculated for. Currently, this always equates to a
 * financial year, for example, 2022-04-01 to 2023-03-31
```

app/services/bill-runs/two-part-tariff/prepare-licences-for-allocation.service.js
```diff
@@ -0,0 +1,142 @@
...
+ */
+async function go (licences, billingPeriod) {
+  for (const licence of licences) {
+    await _prepareReturnLogs(licence, billingPeriod)
+    _prepareChargeVersions(licence, billingPeriod)
+  }
+}
+
+/**
+ * Checks if the abstraction period is outside the given return line's period
```
Cruikshanks, line 27:
```suggestion
 * Checks if a return has evidence of abstraction outside the return's abstraction periods
 * 
 * Each line has a start and end date. If that period does not overlap one of the abstraction periods assigned to the
 * licence the return needs to be flagged as having water been abstracted outside the agreed abstraction period.
```

app/services/bill-runs/two-part-tariff/prepare-licences-for-allocation.service.js
```diff
@@ -0,0 +1,142 @@
...
+    returnLog.allocatedQuantity = 0
+    returnLog.abstractionPeriods = abstractionPeriods
+    returnLog.abstractionOutsidePeriod = abstractionOutsidePeriod
+    returnLog.matched = false
+  })
+}
+
+/**
+ * Sorts charge references by subsistence charge in descending order
+ * @param {Object[]} chargeReferences - Charge references to sort
```
Cruikshanks, line 119:
```suggestion
 * Sorts charge references by their subsistence charge in descending order
 * 
 * The rules for matching returns to charge elements require us to consider those charge elements assigned to the
 * charge reference with the highest subsistence charge first.
 * 
 * Normally, we'd deal with the ordering of things in the original fetch service. But to do that in this case would have
 * meant either creating a complex Knex-based query instead of just using Objection.js or, looping through the results
 * both in `FetchChargeVersionsService` and here.
```

Cruikshanks (changes requested):
Assuming you are happy with my suggested change to the unit tests, I'd also suggest some enhancements to the comments to help us remember why we did things in 6 months-time 🕐 😁

Cruikshanks (approved):

PR #590, opened 2023-12-14
Returns requirement - Select reason for return requirements page

Demwunz:
https://eaflood.atlassian.net/browse/WATER-4252

This is the first page in the returns requirements journey.

### DIFF
--- a/app/views/return-requirements/reason.njk
+++ b/app/views/return-requirements/reason.njk
@@ -3,7 +3,7 @@
 {% from "govuk/components/button/macro.njk" import govukButton %}
 
 {% set title = "Select the reason for the return requirement" %}
-{% set rootLink = "/licences/" + licenceId %}
+{% set rootLink = "/system/licences/" + licenceId %}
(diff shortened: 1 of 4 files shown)


app/views/return-requirements/reason.njk
```diff
@@ -0,0 +1,26 @@
+{% extends 'layout.njk' %}
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+{% set title = "Select the reason for the return requirement" %}
+{% set rootLink = "/licences/" + licenceId %}
```
robertparkinson, line 6: you need the `/system/` at the beginning of the url

robertparkinson (changes requested):

app/views/return-requirements/reason.njk
Demwunz: thanks, missed that

PR #591, opened 2023-12-14
Returns requirement - How do you want setup returns required journey

Demwunz:
https://eaflood.atlassian.net/browse/WATER-4254

This is the third page in the returns requirements journey.

### DIFF
--- a/app/views/return-requirements/returns-how-do-you-want.njk
+++ b/app/views/return-requirements/returns-how-do-you-want.njk
@@ -2,7 +2,7 @@
 {% from "govuk/components/back-link/macro.njk" import govukBackLink %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
 
-{% set rootLink = "/licences/" + licenceId %}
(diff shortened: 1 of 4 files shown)


app/views/return-requirements/returns-how-do-you-want.njk
```diff
@@ -0,0 +1,25 @@
+{% extends 'layout.njk' %}
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+{% set rootLink = "/licences/" + licenceId %}
```
robertparkinson, line 5: you need to have `/system/` at the beginning of the url

robertparkinson (changes requested):

app/views/return-requirements/returns-how-do-you-want.njk
Demwunz: thanks, updated

PR #594, opened 2023-12-15
Set up session table (return requirements)

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4257

We need to store user journeys as people progress through the screens and we need a table to store these in.


test/models/session.model.test.js
```diff
@@ -0,0 +1,37 @@
...
+const { expect } = Code
+
+// Test helpers
+const DatabaseHelper = require('../support/helpers/database.helper.js')
+const SessionHelper = require('../support/helpers/session.helper.js')
+
+// Thing under test
+const SessionModel = require('../../app/models/session.model.js')
+
+describe('Sessions model', () => {
```
Cruikshanks, line 17:
```suggestion
describe('Session model', () => {
```

test/support/helpers/session.helper.js
```diff
@@ -0,0 +1,53 @@
...
+ * It will override or append to them any data provided. Mainly used by the `add()` method, we make it available
+ * for use in tests to avoid having to duplicate values.
+ *
+ * @param {Object} [data] Any data you want to use instead of the defaults used here in the database
+ */
+function defaults (data = {}) {
+  const defaults = {
+    id: generateUUID(),
+    data: {}
+  }
```
Cruikshanks, line 42: Where a table automatically generates an ID we don't tend to set one in the helper. If you've seen us do this it should normally come with a note saying the previous team failed to implement an automatic UUID on insert in the underlying table, hence the need to do it.

So, we just focus on setting meaningful defaults to reflect what an instantiated record would typically look like.

In this particular case `data: {}` could be what we often find. So, I'd suggest one of the following, I'll leave you to decide which one to go with!

```javascript
  const defaults = {
    data: { licenceId: '01/128' }
  }
```

> Or whatever pretend test values you'd like to set in `data:`

or 


```javascript
  const defaults = {}
```

Cruikshanks (changes requested):
Just the tweak-y-est of tweaks! 😁

PR #597, opened 2023-12-15
Move rtn req. routes and session handling

robertparkinson:
Having discussed and agreed a way forward for handling temporary sessions in our code base this demonstrates how we can create the initial session, then retrieve it in subsequent requests.

The other thing it does is break the dependence on the return requirement journey sitting under `/licences`. `/licences` is still the gateway to the journey, creating the temporary session for tracking the journey. But it then redirects away so `/return-requirements` can take over.

The benefit of this is without any additional controls, we have a way of 'clearing' the session. Essentially, the user simply has to leave the journey and start again and the session is cleared. But whilst in the journey the user can go forwards and backwards throughout and each page can validly assume `sessionId` refers to an existing record.


Demwunz (approved):
👍🏽 

seems like there is one thing that needs to be fixed. - https://sonarcloud.io/code?id=DEFRA_water-abstraction-system&branch=lick-it-and-stick-it&selected=DEFRA_water-abstraction-system%3Aapp%2Fcontrollers%2Freturn-requirements.controller.js&line=138

app/controllers/return-requirements.controller.js
```diff
@@ -0,0 +1,162 @@
...
+
+  const session = await SessionModel.query().findById(sessionId)
+
+  return h.view('return-requirements/returns-how-do-you-want.njk', {
+    activeNavBar: 'search',
+    ...session
+  })
+}
+
+async function saveReturnsHowDoYouWant (request, h) {
```
Demwunz, line 59: sonarcloud is complaining about this function being the same as `saveNote`

Demwunz (changes requested):
https://sonarcloud.io/code?id=DEFRA_water-abstraction-system&branch=lick-it-and-stick-it&selected=DEFRA_water-abstraction-system%3Aapp%2Fcontrollers%2Freturn-requirements.controller.js&line=138

Cruikshanks (approved):
I'm not worried about SonarCloud at the moment as a lot of this code will change once we start adding in validation. So...

app/controllers/return-requirements.controller.js
robertparkinson: I put in a small change to make it pass.

Demwunz (approved):

app/plugins/auth.plugin.js
```diff
@@ -47,6 +47,7 @@ const AuthPlugin = {
       },
       redirectTo: '/signin',
       validate: async (_request, session) => {
+        console.log('🚀 ~ file: auth.plugin.js:50 ~ validate: ~ session:', session)
```
Cruikshanks, line 50: Doh! Something I should have spotted the first time around.

```suggestion
```

Cruikshanks (changes requested):
Something I spotted on the first pass I should have flagged but forgot. 😬

Cruikshanks (approved):

app/plugins/auth.plugin.js
robertparkinson: weird that it's showing up on my PR.. it's not something I added. but i've removed it now

PR #600, opened 2023-12-15
Expand DB rollback script options in package.json

Cruikshanks:
When you run `npm run migrate` you are calling [knex migrate:latest](https://knexjs.org/guide/migrations.html#migration-cli). **Knex** will look to see which migrations have not been done and run them as a batch. Instinctively then you might assume running `npm run rollback` would roll back the last batch.

But because we copied it from the legacy code without thinking, what gets called is `knex migrate:rollback --all`. This will rollback _all_ migrations.

There are times we still will want to do this so this change doesn't remove the option. But to better match expectations we update the scripts so that `npm run rollback` calls `knex migrate:rollback`, and a new `npm run rollback:all` calls `knex migrate:rollback --all`.


Cruikshanks:
Doh! Doh! Doh!

This is what happens when you spend a weekend crunching in too many branches and not taking enough due care.

These changes got inadvertently merged in as part of [Refactor to use new view based WATER models](https://github.com/DEFRA/water-abstraction-system/pull/569). This is why you should do as the tech lead says and not as they do, because what they do is naughty 😈

PR #610, opened 2023-12-19
Enter abstraction period dates for return set up

Demwunz:
https://eaflood.atlassian.net/browse/WATER-4274

This is a basic page with a form to enter abstraction period dates for a return set up.

This is a stub page.


app/views/return-requirements/abstraction-period.njk
```diff
@@ -0,0 +1,30 @@
+{% extends 'layout.njk' %}
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+{% set title = "Enter the abstraction period for the return requirement
```
robertparkinson, line 5: I don't mind doing it this way but we need to be consistent. some of the pages have the text in the H1 element. so we either need to change the others to this way or make this the same as those.

robertparkinson (changes requested):

app/views/return-requirements/abstraction-period.njk
Demwunz: @robertparkinson I've created another branch and moved all title strings in other files to use the set.
I'm lifting this from the prototype, so it's less friction when copying across nunjucks code.

robertparkinson (approved):

PR #616, opened 2023-12-20
Persisting the results data from the two-part tariff match and allocate service

Beckyrose200:

As part of the work we have been doing, we've been hacking away at a branch in a mob to build the two-part-tariff service. As a result of this, we have some refactoring to do to tidy up the work we have done in our mob. We are keeping track of the refactoring to do in this issue DEFRA/water-abstraction-team#105. This PR will extract the `persistAllocatedLicencesToResultsService `out of our hacky branch and into its own PR along with the unit tests.

### DIFF
--- a/app/views/bill-runs/view.njk
+++ b/app/views/bill-runs/view.njk
@@ -145,7 +145,7 @@
     govukButton({
       classes: "govuk-button--secondary",
       text: "Download this bill run",
-      href: cancelBillRunLink
+      href: downloadBillRunLink
     })
   }}
 


app/services/bill-runs/two-part-tariff/persist-allocated-licences-to-results.service.js
```diff
@@ -0,0 +1,120 @@
...
+const ReviewChargeElementResultModel = require('../../../models/review-charge-element-result.model.js')
+const ReviewReturnResultModel = require('../../../models/review-return-result.model.js')
+const ReviewResultModel = require('../../../models/review-result.model.js')
+
+/**
+ * Persists the returnLogs and chargeElements processed from the `allocateReturnsToLicenceService`
+ *
+ * @param {String} billRunId UUID of the bill run this bill will be linked to
+ * @param {Object[]} licences licences with returns data to persist
+ *
```
Cruikshanks, line 17: A suggestion to expand the comments to explain a little more about what is going on.

```suggestion
 * Persists results of matching and allocating return logs to licence charge elements for a two-part tariff bill run
 *
 * Each licence will have a `returnLogs` property that contains all return logs linked to the licence for the billing
 * period of the bill run. It will also have a `chargeVersions` property, which within each one it will have details of
 * which charge elements matched to which return logs and whether anything was allocated to the charge element from
 * them.
 *
 * We need to persist all this information ready for use in the review screens that our users use to asses if the
 * matching and allocating looks correct or if any issues need resolving first.
 *
 * @param {String} billRunId - the ID of the two-part tariff bill run being generated
 * @param {Object[]} licences - the two-part tariff licences included in the bill run, along with their match and
 *  allocation results
```

Cruikshanks (changes requested):
I have some suggestions around the comments. We've already discussed moving where the `reviewReturnResultId gets set so I haven't bothered to flag it here.

Cruikshanks (commented):
One more suggested extra comment to include if you think it makes sense. I'll be all done after that - promise!

Cruikshanks (approved):

PR #630, opened 2024-01-03
Add LicenceDocumentRoleModel from document_roles

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4292

> This is part of a series of changes to add the ability to identify the licence holder for a licence

This change adds the migration, model, test helper and unit tests for a new `LicenceDocumentRoleModel`.

We added [LicenceDocumentModel](https://github.com/DEFRA/water-abstraction-system/pull/618). Though pointless, it's the only thing that points us to who the licence holder is. In the `crm_v2` schema it does this in `crm_v2.document_roles` which is the join table between `crm_v2.documents` and `crm_v2.roles`. `roles` is a lookup table of the 'role' a `crm_v2.contact` or `crm_c2.company` has, for example, licence holder! We added that as [LicenceRoleModel](https://github.com/DEFRA/water-abstraction-system/pull/629).

We are now on the final step which is adding the model that allows us to identify and extract the `crm_v2.document_roles` record that identifies _who_ is the current licence holder for a licence.

### DIFF
--- a/app/models/address.model.js
+++ b/app/models/address.model.js
@@ -23,6 +23,14 @@ class AddressModel extends BaseModel {
           from: 'addresses.id',
           to: 'billingAccountAddresses.addressId'
         }
+      },
+      licenceDocumentRoles: {
+        relation: Model.HasManyRelation,
+        modelClass: 'licence-document-role.model',
+        join: {
+          from: 'addresses.id',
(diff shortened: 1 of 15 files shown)


Cruikshanks:
Normally this would be done. But now we have this model we need to also add the relationships to the previous models we created.

PR #631, opened 2024-01-03
Licence Summary Page - Summary Tab

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-431


app/presenters/view-licences/view-licence.presenter.js
```diff
@@ -0,0 +1,38 @@
...
+  const endDate = _endDate(licenceData)
+
+  return {
+    id: licenceData.licence.id,
+    licenceRef: licenceData.licence.licenceRef,
+    region: licenceData.licence.region.displayName,
+    startDate: format(new Date(licenceData.licence.startDate), dataFormat),
+    ...(endDate && { endDate })
+  }
+}
```
Cruikshanks, line 34: Just re-reading [function naming conventions for services](https://github.com/DEFRA/water-abstraction-team/blob/main/coding_conventions.md#function-naming-conventions-for-services) we don't explicitly state that public methods go before private, and _then_ in alphabetical order! 🤦 

But that's the main change needed here.

Next up is the comments. We try and give public methods (which is just `go()` 99% of the time) the full [JSDoc treatment](https://jsdoc.app/). On private methods it is 'dealers choice'. If something is complex or exceptional we'll often add a comment. But we don't worry about following JSDoc completely. The suggestion provides an example of how this might look in this case.

> The suggestion also assumes the comment about just returning the `LicenceModel` instance rather than wrapping it as an object in `FetchLicenceService` has been applied.

The dates on the licence have been wrapped in `new Date()` calls. Only because of the [rabbit hole I fell down I believe this is not needed](https://github.com/DEFRA/water-abstraction-system/pull/655#issue-2076278837). So, I've amended that in my suggestion but you may have encountered something different. 🤷 

```suggestion
/**
 * Formats data for the view `/licences/{id}/` page
 *
 * @param {module:LicenceModel} licence - The licence where the data will be extracted for from
 *
 * @returns {Object} The data formatted for the view template
 */
function go (licence) {
  const { expiredDate, id, licenceRef, region, startDate } = licence
  const endDate = _endDate(licence)

  return {
    id,
    endDate: _endDate(expiredDate),
    licenceRef,
    region: region.displayName,
    startDate: formatLongDate(startDate)
  }
}

function _endDate (expiredDate) {
  if (!expiredDate) {
    return null
  }
  
  return formatLongDate(expiredDate)
}
```

app/routes/licence.routes.js
```diff
@@ -4,6 +4,18 @@ const LicencesController = require('../controllers/licences.controller.js')
...
+    path: '/licences/{id}',
+    handler: LicencesController.summary,
+    options: {
+      auth: {
+        access: {
+          scope: ['billing']
+        }
+      },
+      description: 'Licence summary tab'
+    }
```
Cruikshanks, line 17: Based on the comment about changing the handler it would need to be changed here as well.

```suggestion
    method: 'GET',
    path: '/licences/{id}',
    handler: LicencesController.view,
    options: {
      auth: {
        access: {
          scope: ['billing']
        }
      },
      description: 'View a licence page'
    }
```

app/services/licences/fetch-licence.service.js
```diff
@@ -0,0 +1,51 @@
+'use strict'
+
+/**
+ * Fetches data needed for the bill run page which includes a summary for each bill linked to the bill run
```
Cruikshanks, line 4: Damn you copy & paste!

```suggestion
 * Fetches data needed for the view '/licences/{id}` page
```

app/services/licences/fetch-licence.service.js
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Fetch the licence data from the water schema
+ *
+ * Was built to provide the data needed for the '/licences/{id}' page
+ *
+ * @param {string} id The UUID for the licence to fetch
+ *
+ * @returns {Promise<Object>} the data needed to populate the summary page
+ */
```
Cruikshanks, line 18: Good or bad (we're still deciding!) but the intent of adding views that we then base our models on was to hide the confusion that all the schemas add and make it seem like we're dealing with just one DB/schema.

So, I'd suggest not worrying about referencing that here.

```suggestion
/**
 * Fetch the matching licence and return data needed for the view licence page
 *
 * @param {string} id The UUID for the licence to fetch
 *
 * @returns {Promise<Object>} the data needed to populate the view licence page and some elements of the summary tab
 */
```

app/services/licences/fetch-licence.service.js
```diff
@@ -0,0 +1,51 @@
...
+async function go (id) {
+  const licence = await _fetchLicence(id)
+
+  return {
+    licence
+  }
```
Cruikshanks, line 24: It might be not seeing the benefit (so feel free to push back!) But normally we just return the instance directly rather than wrapping it in a new object. Is there any reason not to do this?

```suggestion
  return licence
```

app/views/view-licences/licence.njk
Cruikshanks, line 1: This is linked to my opinion/convention on the 'root' requests like `GET /thing/{id}`. For those the handler will be `view()` and we like to try and keep the view template name consistent with the handler to make it easier to connect the two.

So, in this case I'd suggest renaming both the folder and file to be on this path; `app/views/licences/view.njk`.

app/views/view-licences/licence.njk
```diff
@@ -0,0 +1,52 @@
...
+      href: "/licences"
+    })
+  }}
+{% endblock %}
+
+{% block content %}
+  <h1 class="govuk-heading-l">Licence Number {{ licenceRef }}</h1>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+        <div class="govuk-tabs" data-module="govuk-tabs">
```
Cruikshanks, line 17: This is more just a pointer to have a discussion. I just assumed I'd see the Nunjucks version of the [GOV.UK tabs implementation](https://design-system.service.gov.uk/components/tabs/).

I can see this is the HTML version. We've currently preferred going with Nunjucks over HTML. So, I just wanted to check there wasn't something I was missing or a blocker to doing so in this case.

Cruikshanks (changes requested):
Woohoo! We are finally going to take on the mess that is the legacy view licence page! Fantastic start.

There are some changes I'd like to make (and more reasons to get off my 🍑 and update our conventions 🤦😩!) There are also some things I haven't thought of that are worth having a discussion about.

But there is an overall thought I think I have about the change. Currently, the service and presenter built treat the licence page and summary tab as one-and-the-same.

But when we come to the other tabs they will share some things currently being used, but not all.

So, is it worth at this stage splitting the work needed to fetch and present data for the licence page from that needed for the summary tab?

app/presenters/licences/view.presenter.js
Cruikshanks, line 1: As the presenter is the `ViewLicencePresenter` we'd expect the file name to be `view-licence.presenter.js`.

This made me stop and think about whether it should match the name of the handler, route and template as you have it. But I _think_ we've been more verbose with the file names as you can still make the connection whilst making it clear what the presenter is for.

app/presenters/licences/view.presenter.js
```diff
@@ -0,0 +1,41 @@
...
+ * @module ViewLicencePresenter
+ */
+
+const { formatLongDate } = require('../base.presenter.js')
+
+/**
+ * Formats data for the `/licences/{id}/` page
+ * @param {module:LicenceModel} licence - The licence where the data will be extracted for from
+ * @module ViewLicencePresenter
+ */
```
Cruikshanks, line 14:
```suggestion
/**
 * Formats data for the `/licences/{id}/` page
 *
 * @param {module:LicenceModel} licence - The licence where the data will be extracted from
 *
 * @returns {Object} The data formatted for the view template
 */
```

app/presenters/licences/view.presenter.js
```diff
@@ -0,0 +1,41 @@
...
+    licenceRef,
+    region: region.displayName,
+    startDate: formatLongDate(startDate)
+  }
+}
+
+/**
+ * Formats the end date of the licence
+ * @module ViewLicencePresenter
+ */
```
Cruikshanks, line 30:
```suggestion
/**
 * Formats the expired date of the licence as the end date for the view
 */
```

app/presenters/licences/view.presenter.js
```diff
@@ -0,0 +1,41 @@
...
+function _endDate (expiredDate) {
+  if (!expiredDate || new Date(expiredDate) < Date.now()) {
```
Cruikshanks, line 32: Just want to double-check if we need to wrap `expiredDate` in a `new Date()` call. Again, I might have it wrong but my understanding is it is already a Date value.

package.json
```diff
@@ -38,6 +38,7 @@
     "@smithy/node-http-handler": "^2.0.4",
     "bcryptjs": "^2.4.3",
     "blipp": "^4.0.2",
+    "date-fns": "^3.2.0",
```
Cruikshanks, line 41: I believe we can revert this change to add the package now we're not using it.

test/controllers/licences.controller.test.js
```diff
@@ -148,4 +150,42 @@ describe('Licences controller', () => {
...
+      })
+    })
+
+    function _viewLicenceSummary () {
+      return {
+        id: '7861814c-ca19-43f2-be11-3c612f0d744b',
+        licenceRef: '01/130/R01',
+        region: 'Southern',
+        startDate: formatLongDate(new Date('2022-11-01')),
+        endDate: formatLongDate(new Date('2032-11-01'))
```
Cruikshanks, line 187: I think if we are hard-coding the dates we're creating (which I 💯 agree with in unit tests) we could skip a step and just set the values that will be returned.

This saves bringing in the `BasePresenter` at the top of the tests as well.

```suggestion
        startDate: '1 November 2022',
        endDate: '1 November 2032'
```

Cruikshanks (changes requested):
Just a few minor tweaks and queries from the second pass.

But the key one is we need to have a discussion about using Nunjucks vs HTML from the design system before this gets merged.

Cruikshanks (approved):

PR #635, opened 2024-01-04
Fix missing bill in view bill run screen

Cruikshanks:

We are replacing the legacy bill run view with one we have developed in this project. During testing, it was found that there are examples of bill runs where a bill is missing from the view. When we checked the legacy version the bill was present so this is an issue just in our version of the page.

This change fixes the feature to ensure all bills are displayed in the view as expected.


Cruikshanks:
Closing as we'll be dealing with this as a data fix. See https://github.com/DEFRA/water-abstraction-service/pull/2381

PR #636, opened 2024-01-04
Why no returns required? page iteration 2

Demwunz:
https://eaflood.atlassian.net/jira/software/c/projects/WATER/boards/96?selectedIssue=WATER-4262

This page has the fields required for the first part of the journey.

The fields are required and should return an error when not chosen.

The validation should be reusable and work again on check-answers page.

<details>
<summary>page screenshot</summary>
image
</details>

<details>
<summary>page screenshot with errors</summary>
image
</details>


Demwunz:
> It looks like a lot of requests for changes. But this always happens as folks get used to my 'tidy-freak' expectations about conventions and patterns. 😁
> 
> As we discussed though, once this is in you then have a blue-print for all the other pages which should make this painless in future.

That's fine, there's a particular way of doing things which I appreciate. I'll be sure to keep your feedback in mind from hereon out.

Cruikshanks (approved):

PR #639, opened 2024-01-05
Add licence ref. and holder to rtn. req. session

Cruikshanks:

In WATER-4292 we added a much of models (`LicenceDocument`, `LicenceRole` and `LicenceDocumentRole`) in preparation for identifying the current licence holder for a licence. We need the licence reference on every page of the return requirement setup journey because the page design has it as a caption alongside the page title.

Rather than the number on each page, it would be better to just grab the information once, and add it to the 'session' record the journey is using. But as we also need the licence holder ready for the check answers page we may as well grab that at the same time.

So, this change refactors the `LicencesController` to use a new `InitiateReturnRequirementSessionService` that will handle fetching this licence data and instantiating the initial  `SessionModel` instance. Subsequent controller handlers will then be able to access this data when they fetch the session record.

### DIFF
--- a/app/controllers/licences.controller.js
+++ b/app/controllers/licences.controller.js
@@ -10,9 +10,9 @@ const InitiateReturnRequirementSessionService = require('../services/return-requ
 async function noReturnsRequired (request, h) {
   const { id } = request.params
 
-  const sessionId = await InitiateReturnRequirementSessionService.go(id)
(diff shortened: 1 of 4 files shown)


Cruikshanks:
Do not merge this until it has been refactored to use #640 instead of the models it is currently using.

robertparkinson (approved):

PR #640, opened 2024-01-07
Add LicenceDocumentHeaderModel

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4322

> This is part of a series of changes to allow us to replace the legacy view licence page

When viewing a licence using the [legacy water-abstraction-ui](https://github.com/DEFRA/water-abstraction-ui) the summary tab features several links to other pages, some of which give expanded details for the item shown.

Those links rely on having the `crm.document_header` ID for the licence currently being viewed. Our work on WATER-4292 (identifying the licence holder) has shown that much of the data in the table has already been imported elsewhere.

But without the ID those links and pages won't work and they are beyond the scope of what we are doing with the view licence page.

So, to allow us to quickly get the related `crm.document_header` (that we have chosen to call `LicenceDocumentHeader`) for a licence this change adds the model to the project.


robertparkinson (approved):

Cruikshanks:
Thank you for reviewing this @robertparkinson . But further testing has shown that even though the legacy code doesn't usethe `LicenceDocument` based models, they _do_ point to the correct result.

As the `LicenceDocument` based models hold transformed data, i.e. the NALD data hasn't just been dumped in a JSONB field I'm going to stick with them. Which unfortunately makes this PR redundant. I'm sorry for wasting your time with this. If it helps I wasted 10x as much of my own time falling down the wrong hole. 😬

Cruikshanks:
I think we are going to need this anyway. We are currently working on rebuilding the view licence page and replacing the legacy UI version (WATER-4322 ).

A number of the links in the summary tab depend on having the document header ID. So, even if we do nothing else with this model we're going to need it to grab the correct ID for the links.

robertparkinson (approved):

PR #644, opened 2024-01-09
Alternate 2pt matching engine version 2

Beckyrose200:

We had a previous hacky branch for the two part tariff engine however the branch got to over 100 commits and made it difficult to keep track of the version control. This new branch is focussing on a second version of the engine. In this version rather then going through every service for ALL the licences we are focusing on running them one at a time. What this means is the engine will take fetch all the licences and return logs as it has previously done, but rather then processing everything at once will work to process them one at a time.


Cruikshanks:
All the changes needed from this PR have now been brought into `main`. We can close this spike PR down now.

PR #647, opened 2024-01-10
Housekeeping of return requirement routes

Cruikshanks:

This is some housekeeping of the return requirements setup routes before things get too baked in.

Primarily it started out as sorting the order of the handlers, routes and tests to be alphabetical. However, once we dug in we spotted a number of other things that needed sorting.

- make the endpoint paths consistent
- make handler names consistent with path names
- add missing `POST` endpoints
- add missing `POST` handlers
- redirect instead of return in all `POST` handlers
- rename `POST` handlers from `save` to `submit` (reason being they are handling the form _submission_, and if there is a validation issue no save will take place)
- ensure licence ID gets through to the `/approve` page
- make route descriptions consistent
- refactor tests to reuse `options`
- add back links to `check-your-answer` pages

So, it's a bigger change than we'd normally do. But a number are pretty intrinsically linked so we're bundling them into this single housekeeping change.


Demwunz (dismissed):
👍🏽

app/views/return-requirements/check-your-answers.njk
```diff
@@ -1,7 +1,20 @@
 {% extends 'layout.njk' %}
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
 
 {% set title = "Check your answers" %}
+{% set rootLink = "/system/return-requirements/" + id %}
```
robertparkinson, line 6: this page did not have a back button on it in the story and shabbir questioned it so I removed it. I think the point being you should be using the change links on the page to go back to anything you want to change? https://eaflood.atlassian.net/browse/WATER-4255

robertparkinson (dismissed):

app/views/return-requirements/check-your-answers.njk
Cruikshanks: Okey-dokey. That's valid and I don't see them in the prototype. I'll whip them out then merge this in when I get back from school run!

PR #651, opened 2024-01-10
Fix missing titles in returns reqs. set up pages

Cruikshanks:

Whilst working on some [housekeeping](https://github.com/DEFRA/water-abstraction-system/pull/647) and [tweaks](https://github.com/DEFRA/water-abstraction-system/pull/648) we spotted that the titles for all the new returns requirements set up pages were missing.

A quick check identified that all the pages are using our standard `layout.njk` (😁) but it expects a `pageTitle` value to be passed in the context and none of the pages are doing this (😱).

This change fixes the issue and adds the missing titles.

---

We also ended up doing a bit more housekeeping.

- remove the duplicate `/no-returns-check-your-answers`
- fix approve button on `check-your-answers` not redirecting properly
- remove Nunjucks `set title=` and just use `{{ pageTitle }}` in all templates
- fix 'Return to licence setup' link on approve page


Cruikshanks:
SonarCloud is right about the missing coverage. But that is because we're tinkering or adding logic that will be superseded shortly. So, we're just going to ignore it for now 😁

app/views/return-requirements/abstraction-period.njk
```diff
@@ -19,7 +17,7 @@
 {% block content %}
   {# Main heading #}
   <div class="govuk-body">
-    <h1 class="govuk-heading-xl govuk-!-margin-bottom-3">{{ title }}</h1>
+    <h1 class="govuk-heading-xl govuk-!-margin-bottom-3">Enter the abstraction period for the return requirement</h1>
```
robertparkinson, line 20: if pageTitle will always match the H1 question we should just reuse the pageTitle here rather than duplicating the text

robertparkinson (changes requested):

app/views/return-requirements/abstraction-period.njk
Cruikshanks: Great minds and all that!

sleep-blocker

robertparkinson (approved):

PR #656, opened 2024-01-11
Refactor Prepare Licence For Allocation Service

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4319

This pull request refactors the 'PrepareLicenceForAllocationService' as part of our ongoing effort to enhance the two-part-tariff service. Initially we began by creating a 'hack' branch to build and iterate on the two-part-tariff functionality. In the process of working on these branches, we identified a cleaner and more performant approach to implement the overall service.  We realised that the PrepareLicenceForAllocationService could be split into two smaller services, making it easier in the code to follow the two-part-tariff process and making it easier for testing.


Cruikshanks (commented):
I pushed some corrections. If you are ok with them I'll approve and we can get this merged.

Beckyrose200:
Looks all good to me :)

Cruikshanks (approved):
The only thing I would ask before merging is that the squashed commit & PR description are updated.

The 'hack' branches once 2PT is complete will be lost. So, we can reference things we're doing on them because there will be nothing to look up.

So, we need to explain in this commit/description that we've realised we can simplify the two-part tariff process and services involved by splitting the old `PrepareLicence` into 2 separate services.

PR #670, opened 2024-01-18
Add warning text to view licence page

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4325

As part of the work to migrate the summary page from the old UI to the new one we need to display a warning when a licence has been revoke, expired or lapsed.

### DIFF
--- a/app/presenters/licences/view-licence.presenter.js
+++ b/app/presenters/licences/view-licence.presenter.js
@@ -15,28 +15,110 @@ const { formatLongDate } = require('../base.presenter.js')
  * @returns {Object} The data formatted for the view template
  */
 function go (licence) {
-  const { expiredDate, id, licenceRef, region, startDate } = licence
+  const { expiredDate, id, lapsedDate, licenceRef, region, revokedDate, startDate } = licence
+  const warning = _generateWarningMessage(expiredDate, lapsedDate, revokedDate)
 
   return {
     id,
     endDate: _endDate(expiredDate),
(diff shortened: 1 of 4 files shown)


app/presenters/licences/view-licence.presenter.js
```diff
@@ -15,22 +15,46 @@ const { formatLongDate } = require('../base.presenter.js')
...
+  }
+  return secondEndDate
+}
+
+/**
+  * Formats the expired date of the licence as the end date for the view
+  *
+  * @param {date} expiredDate - The expired date to be formatted if it is in the past
+  *
+  * @module ViewLicencePresenter
```
Cruikshanks, line 56: I might be misunderstanding something because I keep flagging these `@module` tags in JSDoc comments for removal!

According to [JSDocs](https://jsdoc.app/howto-commonjs-modules)

> In most cases, your CommonJS or Node.js module should include a standalone JSDoc comment that contains a [@module tag](https://jsdoc.app/tags-module).

This is why we normally start with one of these at the top of our files. We then don't include them inside function comments.

But I might have gotten it wrong so should stop trying to remove them from your comments. 🤷 

> Also, same comment about whether you need a JSDoc comment for this private function applies here as well!

Cruikshanks (changes requested):

Cruikshanks (approved):
Guess you didn't like my reworded comment 😢

PR #679, opened 2024-01-22
Add Licence holder name to view licence page

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4328

This ticket is for adding the Lice holder name to the summary section of the page.


Cruikshanks (changes requested):
Absolutely nothing wrong with the code! 😂 

But would you mind updating the PR title and description to reflect these are the changes for adding the licence holder to the summary tab; WATER-4328.

Then I can relax that we don't squash this in with the wrong details 😰😁

PR #680, opened 2024-01-22
Update 2PT /check endpoint with latest services

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4339

This follows on from [Update 2PT check endpoint with alternate changes](https://github.com/DEFRA/water-abstraction-system/pull/582) where we applied the services we refactored out of our 2PT spike branch to the `/check` endpoint.

The final set of services has now been refactored and added to `main`, so we can update the stand-ins service.

We can also update the `TwoPartService` to use the new way of working, iterating all licences in one place and passing each licence to the new services.

---

We can drop the `PrepareLicences` from stand-ins and use the real one. However, when verifying the results we find it useful to see which lines were allocated to which elements. This is not something we care about when generating the review results for real.

The problem is this can only happen deep inside the allocation process. We don't want to add toggles to the production service just for this. We're on the cusp of having the pages ready to go at which point we expect the `/check` endpoint to become defunct.

So, for now, we just copy the real `AllocateReturnsToChargeElementService` into our stand-ins and add the tweak which captures which lines were allocated to it.

### DIFF
--- a/app/services/check/scenario-formatter.service.js
+++ b/app/services/check/scenario-formatter.service.js
@@ -92,9 +92,10 @@ function _formatChargeElementMatchedReturns (matchedReturns, formattedReturns) {
   return matchedReturns.map((matchedReturn) => {
     const {
       id,
-      allocatedQuantity,
-      lines
+      allocatedQuantity
     } = matchedReturn
+
+    const lines = matchedReturn.lines ? matchedReturn.lines : []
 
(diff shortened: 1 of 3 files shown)


Cruikshanks:
SonarCloud rightly points out that the stand-ins and scenario formatted services could 'be improved'!

But these are temporary modules only there until we have working review pages and the need for the `/check` endpoint is no more.

I could overrule SonarCloud to get all green. But these issues should continue to be flagged until we remove these modules so I won't.

PR #646, opened 2024-01-10
Returns required journey - Select start date page iteration 2 (1 of 7)

Demwunz:
https://eaflood.atlassian.net/browse/WATER-4266

This page contains date controls for a start date for the return requirement.

The page will have specific validation rules for the date which adhere to business requirements.

This is step 1 of a 7 step journey through return requirements.

PR #681, opened 2024-01-23
Move licence end date logic to model

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4325
https://eaflood.atlassian.net/browse/WATER-4263

We have had two tickets come through that both need to determine if and when a licence 'ends'.

A licence can 'end' for 3 reasons:

- because it is _revoked_
- because it is _lapsed_
- because it is _expired_

The previous delivery team chose to encode these as 3 separate date fields on the licence record. So, if a field is populated it means the licence 'ends' for that reason on that day. But it's not that simple! 😁

More than one of these fields may be populated. For example, a licence was due to expire on 2023-08-10 but was then revoked on 2022-04-27. So, to determine the reason you need to select the _earliest_ date.

But there's more! 😲

We have examples where 2 of the fields might be populated with the same date (and 1 licence where all 3 have the same date!) So, how do you determine the 'end' reason then? If more than one date field is populated and they hold the earliest date value then we select based on priority; revoked -> lapsed -> expired.

Both tickets are connected to PRs that have to handle all or some of this logic.

- [Returns required journey - Select start date page iteration 2 (1 of 7)](https://github.com/DEFRA/water-abstraction-system/pull/646)
- [Add warning text to view licence page](https://github.com/DEFRA/water-abstraction-system/pull/670)

Both PRs will also have to write numerous tests to check the right date is being used, and we have a strong suspicion we are already doing something similar in supplementary billing and that we'll need the 'end' in future tickets.

So, enough is enough! This refactoring change moves the logic to the model, specifically, as a [custom Objection.js instance method](https://vincit.github.io/objection.js/api/model/instance-methods.html). This is something we've done in a [previous project](https://github.com/DEFRA/sroc-charging-module-api) and in this project to determine a `ContactModel`s name.

We can then remove the duplication both in logic and unit tests.

This change updates the model, adds the new tests, and then refactors existing usages to use the model's `$ends()` method.

### DIFF
--- a/app/models/licence.model.js
+++ b/app/models/licence.model.js
@@ -74,6 +74,60 @@ class LicenceModel extends BaseModel {
       }
     }
   }
+
+  /**
+   * Determine the 'end' date for the licence
+   *
+   * A licence can 'end' for 3 reasons:
+   *
+   * - because it is _revoked_
+   * - because it is _lapsed_
+   * - because it is _expired_
+   *
+   * The previous delivery team chose to encode these as 3 separate date fields on the licence record. So, if a field is
+   * populated it means the licence 'ends' for that reason on that day.
+   *
+   * More than one of these fields may be populated. For example, a licence was due to expire on 2023-08-10 but was then
+   * revoked on 2022-04-27. So, to determine the reason you need to select the _earliest_ date.
+   *
+   * But are examples where 2 of the fields might be populated with the same date (and 1 licence where all 3 have the
+   * same date!) If more than one date field is populated and they hold the earliest date value then we select based on
+   * priority; _revoked_ -> _lapsed_ -> _expired_.
+   *
+   * @returns `null` if no 'end' dates are set else an object containing the date, priority and reason for either the
+   * earliest or highest priority end date
+   */
+  $ends () {
+    const endDates = [
+      { date: this.revokedDate, priority: 1, reason: 'revoked' },
+      { date: this.lapsedDate, priority: 2, reason: 'lapsed' },
+      { date: this.expiredDate, priority: 3, reason: 'expired' }
+    ]
+
+    const filteredDates = endDates.filter((endDate) => endDate.date)
+
+    if (filteredDates.length === 0) {
+      return null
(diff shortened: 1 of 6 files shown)


Cruikshanks:
Ignore SonarCloud! Fake news, fake news! (I'm dealing with it 😩 😬 )

app/models/licence.model.js
```diff
@@ -74,6 +74,60 @@ class LicenceModel extends BaseModel {
...
+    ]
+
+    const filteredDates = endDates.filter((endDate) => endDate.date)
+
+    if (filteredDates.length === 0) {
+      return null
+    }
+
+    filteredDates.sort((firstDate, secondDate) => {
+      if (firstDate.date !== secondDate.date) {
```
robertparkinson, line 114: weird that this is working for you. when I implemented it I found when doing !== or === you need to use date.getTime() as can be explained here: https://stackoverflow.com/questions/492994/compare-two-dates-with-javascript
Or perhaps it is failing and just defaulting to the priority check below? which is still giving you the same result but possibly not for the reason you think? 🤔

robertparkinson (changes requested):

app/models/licence.model.js
Cruikshanks: Spot on insight! Should have known not to trust JavaScript :trollface: 

I've implemented your suggestion, with a couple of commits along the way to confirm how I managed to get away with it!

app/services/return-requirements/initiate-return-requirement-session.service.js
```diff
@@ -45,10 +45,12 @@ async function _createSession (data) {
 
 function _data (licence, journey) {
   const { id, licenceDocument, licenceRef, licenceVersions } = licence
+  const ends = licence.$ends()
 
   return {
     licence: {
       id,
+      endDate: ends ? ends.date : null,
```
Demwunz, line 53: 🎉

robertparkinson (approved):

PR #682, opened 2024-01-23
Display purposes in view licence page summary tab

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4326

This ticket is for adding the purposes to the licence summary tab


db/migrations/legacy/20240129145658_create-licence-version-purposes-table.js
```diff
@@ -0,0 +1,46 @@
...
+      table.date('time_limited_end_date')
+      table.text('notes')
+      table.decimal('annual_quantity')
+      table.dateTime('date_created').notNullable()
+      table.dateTime('date_updated').notNullable()
+      table.string('external_id')
+      table.boolean('is_test')
+
+      // Legacy timestamps
+      // NOTE: They are not automatically set and there are large numbers of records where these fields are null!
```
Cruikshanks, line 32: The comment isn't correct (now doubt just a copy and paste thing)

The fields are `notNullable()` so have to be populated.

```suggestion
      // NOTE: They are not automatically set
```

app/models/licence-version.model.js
```diff
@@ -23,6 +23,18 @@ class LicenceVersionModel extends BaseModel {
...
+        relation: Model.ManyToManyRelation,
+        modelClass: 'purpose.model',
+        join: {
+          from: 'licenceVersions.id',
+          through: {
+            from: 'licenceVersionPurposes.licenceVersionId',
+            to: 'licenceVersionPurposes.purposeId'
+          },
+          to: 'purposes.id'
+        }
```
Cruikshanks, line 37: As a way of ensuring we don't break this relationship, it would be ideal if we also had a test in `test/models/licence-version.model.test.js` that covered this relationship.

app/presenters/licences/view-licence.presenter.js
Cruikshanks, line 1: _Is this_ the point where we break this presenter up? Or would you rather do it when the summary tab is complete?

It's just I'm conscious there are things we will always need to display in the view; page title, licence reference, warning message etc.

But then there are things like purposes which are only relevant if you are viewing the summary tab.

What do you think?

app/presenters/licences/view-licence.presenter.js
```diff
@@ -46,6 +57,21 @@ function _generateLicenceHolder (licenceHolder) {
...
 }
 
+function _generatePurposes (licenceVersions) {
+  if (!licenceVersions || licenceVersions.length === 0 || licenceVersions[0]?.purposes.length === 0) {
+    return null
+  }
+
+  const uniquePurposes = [...new Set(licenceVersions[0].purposes.map((item) => {
+    return item.description
+  }))]
```
Cruikshanks, line 67: Just for clarity in my opinion I'd split this into separate statements.

```suggestion
  const allPurposes = licenceVersions[0].purposes.map((item) => {
    return item.description
  })
  
  const uniquePurposes = [...new Set(allPurposeDescriptions)]
```

Cruikshanks (changes requested):
A request for an additional unit test, a correction in a comment and a suggestion to simplify something. All pretty minor hopefully.

The key feedback is a question; is it time to break up the presenter? 🤔

test/models/licence-version.model.test.js
```diff
@@ -64,5 +67,39 @@ describe('Licence Version model', () => {
         expect(result.licence).to.equal(testLicence)
       })
     })
+
+    describe('when linking to licence version purposes', () => {
```
Cruikshanks, line 71: I've taken the example set out in `user.model.test.js`

```suggestion
    describe('when linking through licence version purposes to purposes', () => {
```

test/models/licence-version.model.test.js
```diff
@@ -64,5 +67,39 @@ describe('Licence Version model', () => {
...
+      })
+
+      it('can successfully run a related query', async () => {
+        const query = await LicenceVersionModel.query()
+          .innerJoinRelated('purposes')
+
+        expect(query).to.exist()
+      })
+
+      it('can eager load the licence version', async () => {
```
Cruikshanks, line 92:
```suggestion
      it('can eager load the purposes', async () => {
```

Cruikshanks (changes requested):
Just a couple of changes in the unit test descriptions. Then all good!

Cruikshanks (approved):

PR #689, opened 2024-01-25
Create licence holder seeder

Jozzey:

Whilst carrying out testing there have been multiple instances where we need to create data for a licence holder. This isn't a simple process and involves creating records in 4 tables in the DB.

We have therefore created a "seeder" which when passed a licence reference will create the records required for a licence holder.


Jozzey:
My commit should have said update TEST not SERVICE

Cruikshanks (approved):
Woo-hoo!

Hopefully, this idea will help clean up the tests a bit!

PR #693, opened 2024-01-27
Add registered user modifier to LicenceModel

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4324

> This is part of a series of changes to allow us to replace the legacy view licence page

When viewing a licence using the [legacy water-abstraction-ui](https://github.com/DEFRA/water-abstraction-ui) the page will display information on who the registered user is.

The registered is an external user who has declared that they manage the licence via the external UI. They are not the licence holder, just the person declared to be responsible for managing it.

Unfortunately, this functionality dates from the very start of the service and the data is in one of the oldest legacy schemas; `crm`. It seems this was one of the things which the previous team never managed to migrate to `crm_v2`, their intended better CRM structure.

We have to go through a number of tables using a specific query to identify the registered user and whether they have set a custom name for the licence.

This change adds a [modifier](https://vincit.github.io/objection.js/recipes/modifiers.html#modifiers) to the `LicenceModel` so we only have to do this once!

### DIFF
--- a/app/models/licence.model.js
+++ b/app/models/licence.model.js
@@ -131,7 +131,7 @@ class LicenceModel extends BaseModel {
             ])
           })
       },
-      registeredUserAndLicenceName (query) {
+      registeredToAndLicenceName (query) {
         query
           .withGraphFetched('licenceDocumentHeader')
           .modifyGraph('licenceDocumentHeader', (builder) => {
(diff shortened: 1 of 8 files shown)


Cruikshanks:
Assigned this one just to you @robertparkinson for review as, hopefully, this will support you in your ticket to add 'registered to' details to the view licence page.

Within the commits I did you will see how to change `FetchLicenceService` to use this new modifier to grab the details you need.

robertparkinson (approved):

PR #702, opened 2024-01-30
Spike new annual bill run process

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4348

We have a working bill run engine for SROC supplementary in **water-abstraction-system**.

So we should have everything we need to create an annual bill run variant.

The problem is that the supplementary engine had the additional complexity of needing to recalculate charge versions over 6 years and then comparing the results to ascertain where charges were cancelled out. The result could be nothing to bill, a single zero-charge bill, or multiple bills.

The annual bill run engine doesn't need any of that complexity.

This 'spike' is to build a working annual bill run engine from the existing code.

Having done this we'll then know

- what existing code should be refactored to be reusable across the engines (reduce duplication and maintenance)
- what new code is needed
- how to break the work down into manageable, reviewable changes


Cruikshanks:
All the work from this has now been merged and we are in the process of validating the new engine. This PR can now be closed and the branch deleted.

PR #704, opened 2024-01-31
Create review an individual licence for 2PT bill run

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4191

As part of our two-part tariff bill run work, we've developed a review page displaying a summary of all licences in a bill run. This PR allows you to click on a specific licence from the review page and review its details within the bill run.


Jozzey (approved):
I think we may be able to pull the data out of the DB so that it's in a format which will require less processing to get it into a state ready to pass to the view.
However, as discussed. This PR only covers the top portion of the screen. The next ticket, which is for the lower portion of the screen will require the query that is fetching the data to be updated. So any changes we make now may no longer be relevant once we start on the next ticket.

Therefore we may as well get this merged in and tested while we work on the lower portion of the screen and see what refactoring we can do once we have a clearer picture of how the final page is going to look.

PR #705, opened 2024-01-31
Connect 2PT bill run request to match & allocate engine

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4343

At this point the the SROC 2PT bilI run is created with a status of `QUEUED` and then … nothing.

We need to complete the final step after creating the bill run. `water-abstraction-system` now needs to `PROCESS` the bill run, triggering the new match & allocate engine, and updating the bill run status to `REVIEW` when it's complete. It also needs to handle marking the bill run as `ERRORED` should one be thrown, or `EMPTY` should there be no licences to bill.


app/services/bill-runs/two-part-tariff/process-two-part-tariff-returns.service.js
```diff
@@ -0,0 +1,66 @@
+'use strict'
+
+/**
+ * Processes the licences for a two-part tariff bill run for the given billing periods
```
Cruikshanks, line 4:
```suggestion
 * Processes the two-part tariff match & allocate stage for the given bill run and billing period
```

app/services/bill-runs/two-part-tariff/process-two-part-tariff-returns.service.js
```diff
@@ -0,0 +1,66 @@
...
+
+function _logError (billRun, error) {
+  global.GlobalNotifier.omfg('Bill run process errored', { billRun }, error)
+}
+
+async function _updateStatus (billRunId, status) {
+  await BillRunModel.query()
+    .findById(billRunId)
+    .patch({ status })
+}
```
Cruikshanks, line 62: A few things are going on here. Some corrections, some suggestions and some come from have recently gone through the billing engine for a second time due to my work on annual.

I've made a bunch of suggested changes below. But I'll list the reasons why here

- Before sending the billing periods to the `MatchAndAllocateService`, we know we only care about `billingPeriods[0]`. Whether we do it here or in `StartBillRunProcessService` was a toss-up. I went for here because then the start service remains consistent across the bill runs and can remain ignorant; it is up to each bill run processing service to decide what to do with the billing periods generated.
- Avoiding `isThing`. IMHO I don't think booleans need to be prefixed with `is` which I know comes from my ruby days. The interweb comes down more on the side of it is fine. But I am 👑 here! (I'm also aware this was already here and I've let it get by me during the supplementary days 😬 )
- Creating too many methods. I do believe in a named method can help explain what a single line of code is doing. But on review, I think we took it too far during supplementary billing. Hence I've gotten rid of `_logError()`
- Cleaned up `_setBillRunStatus()` (I blame too much time with the legacy code during supplementary billing!)
- Remove unnecessary await in `_updateStatus()` (Again, old code we should have known better)
- Update JSDoc comments to match team conventions

```suggestion
/**
 * Matches and allocates licences to returns for a two-part tariff bill run
 *
 * The results of the matching process are then persisted to the database ready for the results to be reviewed. The bill
 * run status is also updated to 'review'.
 *
 * In the unlikely event of no licences match to returns it will set the status to 'empty'. It will also handle updating
 * the bill run if an error occurs during the process.
 *
 * @param {module:BillRunModel} billRun - The two-part tariff bill run being processed
 * @param {Object[]} billingPeriods - An array of billing periods each containing a `startDate` and `endDate`. For 2PT
 * this will only ever contain a single period
 */
async function go (billRun, billingPeriods) {
  const { id: billRunId } = billRun
  // NOTE: billingPeriods come from `DetermineBillingPeriodsService` which always returns an array because it is used by
  // all billing types. For two-part tariff we know it will only contain one because 2PT bill runs are only for a single
  // financial year
  const billingPeriod = billingPeriods[0]

  try {
    const startTime = currentTimeInNanoseconds()

    await _updateStatus(billRunId, 'processing')

    // `populated` will be set to true if `MatchAndAllocateService` processes at least one licence
    const populated = await MatchAndAllocateService.go(billRun, billingPeriod)

    await _setBillRunStatus(billRunId, populated)

    calculateAndLogTimeTaken(startTime, 'Process bill run complete', { billRunId, type: 'two_part_tariff' })
  } catch (error) {
    await HandleErroredBillRunService.go(billRunId)
    global.GlobalNotifier.omfg('Bill run process errored', { billRun }, error)
  }
}

async function _setBillRunStatus (billRunId, populated) {
  // It is highly unlikely no licences were matched to returns. So we default status to 'review'
  let status = 'review'

  // Just in case no licences were found to be matched to returns we set the status to 'empty'
  if (!populated) {
    status = 'empty'
  }

  // Update the bill run's status
  return _updateStatus(billRunId, status)
}

async function _updateStatus (billRunId, status) {
  return BillRunModel.query()
    .findById(billRunId)
    .patch({ status })
}
```

Jozzey:
Sonarcloud is complaining about the length of the `_process` function in the `MatchAndAllocateService`. We have already refactored out all of the actual processing that is going on in there. The function is now not much more than a series of nested `forEach` loops calling various services which I don't think can be refactored any further.

If we do any further refactoring of this function I think it will make it less clear what it is doing so I am going to override the SonarCloud issue for that particular service.

Cruikshanks (approved):

PR #725, opened 2024-02-09
Fix & tidy test DB legacy migrations

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4365

> For context this came out of us working on re-implementing the SROC annual bill run using what we've learnt and components from our supplementary billing engine.

Whilst spiking on getting a working SROC annual billing engine we found that some of the legacy tables the migrations were creating in the test DB didn't match with the real thing. No doubt a result of our original intent being legacy migrations for the test DB did the minimum needed to get tests passing.

We now can fix legacy migrations without borking everyone's local test DB thanks to [Add wipe test DB function to project](https://github.com/DEFRA/water-abstraction-system/pull/724).

So, in this change, we're not just going to fix the 2 issues we found during our annual billing spike. Instead, we've reviewed all the original legacy migrations and fixed any inconsistencies with the real tables we found.

We've then taken the opportunity to tidy them up; grouping them by schema and ordering them in a way that makes sense.

### DIFF
--- a/app/controllers/return-requirements.controller.js
+++ b/app/controllers/return-requirements.controller.js
@@ -73,6 +73,18 @@ async function checkYourAnswers (request, h) {
   })
 }
 
+async function existing (request, h) {
+  const { sessionId } = request.params
+
+  const session = await SessionModel.query().findById(sessionId)
+
+  return h.view('return-requirements/existing.njk', {
+    activeNavBar: 'search',
+    pageTitle: 'Select an existing return requirement from',
+    ...session
+  })
+}
+
 async function frequencyCollected (request, h) {
   const { sessionId } = request.params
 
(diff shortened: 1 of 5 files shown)


Cruikshanks:
This looks like a massive PR but the changes are relatively minor. I've just added missing details in the migrations and corrected a few others.

The extra constraints have meant we need to start populating certain fields hence a few helpers got touched. Also, a couple of tests that created more than one record needed fixing for the same reason.

What has _really_ made things look crazy is me taking the opportunity to group what we have by schema, which meant changing the file names.

The only person this is going to affect right now is @robertparkinson (more than happy if you want me to jump on your PR and fix what merging this will break Robert).

So feel free to 'pass' on reviewing this. Just let me know so I know I can go ahead with abusing my tech-lead privileges.

robertparkinson (approved):

Jozzey (approved):

PR #728, opened 2024-02-12
Create Landing page

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4359

As part of an internal project for a Junior Developer, Roble (me) has be assigned to create a online form to replace the old paper form for external users to sign up for a water abstraction licence.

This PR is to create a landing page for external users.


test/controllers/landing-page.controller.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+// For running our service
+const { init } = require('../../app/server.js')
+
+describe('Landing Page controller', () => {
+  let server
+
+  // Create server before each test
+  beforeEach(async () => {
+    server = await init()
```
Beckyrose200, line 18: Looking at the config for the controller tests in other files, it appears we need to add Sinon to stub the `server.logger ` and `server.app.airbrake`. By doing this we are silencing any notification that might get sent to Errbit.

test/controllers/landing-page.controller.test.js
```diff
@@ -0,0 +1,53 @@
...
+  })
+
+  describe('when a request is made to correct URL', () => {
```
Beckyrose200, line 21: This first describe block should state the path of the route so:

`describe('GET /landing-page', () => {
}`

test/controllers/landing-page.controller.test.js
```diff
@@ -0,0 +1,53 @@
...
+    it('displays the landing page', async () => {
+      const options = {
```
Beckyrose200, line 23: We always declare any variables we need for tests before the `it` block of the test.
So 
`      options = {
        method: 'GET',
        url: '/landing-page'
      }`
Should be declared under the second describe block.

test/controllers/landing-page.controller.test.js
```diff
@@ -0,0 +1,53 @@
...
+        method: 'GET',
+        url: '/landing-page'
+      }
+
+      const response = await server.inject(options)
+      const payload = response.payload
```
Beckyrose200, line 29: Rather than declaring `payload` as a separate variable, in the expect block just call the property on the response object.

`expect(response.payload).to.include('Landing Page)`

test/controllers/landing-page.controller.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+      expect(response.statusCode).to.equal(200)
+
+      expect(payload).to.include('Landing Page')
+      expect(payload).to.include('Lorem ipsum')
+    })
+  })
+
+  describe('when a request is made to an incorrect URL', () => {
```
Beckyrose200, line 38: At the moment there is no reason for the request to fail so we don't need the second test for 
`describe('when a request is made to an incorrect URL', () => {`

Beckyrose200 (changes requested):
A few changes to be made but a brilliant first attempt at the ticket, and considering you did all this without any help is great. If you have any questions on my feedback just give me a ping and we can go through it together :D

test/controllers/landing-page.controller.test.js
```diff
@@ -0,0 +1,44 @@
...
+
+    // We silence calls to server.logger.error in the plugin to keep test output as
+    // clean as possible
+    Sinon.stub(server.logger, 'error')
+
+    // We silence sending a notification to our Errbit instance using Airbrake
+    Sinon.stub(server.app.airbrake, 'notify').resolvesThis()
+  })
+
+  describe('GET /landing-page', () => {
```
Beckyrose200, line 29: Just some slight changes on the test still needed. To follow the conventions we have with all the controller test files we need to add an additional describe block for 'when the request succeeds' and I would also just change the wording for the it block to be more generic that it 'returns the page successfully' 

```
describe('GET /landing-page', () => {
    const options = {
      method: 'GET',
      url: '/landing-page'
    }
      it('returns the page successfully', async () => {
        const response = await server.inject(options)

        expect(response.statusCode).to.equal(200)

        expect(response.payload).to.include('Landing Page')
        expect(response.payload).to.include('Lorem ipsum')
      })
    })
  })
```

test/controllers/landing-page.controller.test.js
```diff
@@ -0,0 +1,44 @@
...
+  beforeEach(async () => {
+    // Create server before each test
+    server = await init()
+
+    // We silence calls to server.logger.error in the plugin to keep test output as
+    // clean as possible
+    Sinon.stub(server.logger, 'error')
+
+    // We silence sending a notification to our Errbit instance using Airbrake
+    Sinon.stub(server.app.airbrake, 'notify').resolvesThis()
```
Beckyrose200, line 26: Since we are using Sinon to stub the logger and airbrake we need an afterEach block for Sinon.restore()
Have a look at `test/controllers/bills.controller.test.js` to see how to do this

```
  afterEach(() => {
    Sinon.restore()
  })
```

Beckyrose200 (changes requested):

Jozzey (commented):
It's looking good 👍🏼  I'll leave it for @Beckyrose200 to approve once she is happy with it.

Beckyrose200 (approved):

PR #731, opened 2024-02-13
Add no returns to licence review page

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4191

Following on from testing for the above ticket it was discovered that it is indeed possible to have a licence in the two-part tariff bill run with no returns. Initially we didn't think this was a possibility but bad data does slip through often enough to warrant us to add in logic to handle it. This PR is for the licence review page to handle no returns, adding in additional text and allowing the licence to still be clicked into and not error.


app/services/bill-runs/two-part-tariff/fetch-review-licence-results.service.js
```diff
@@ -39,6 +41,12 @@ async function _fetchBillRun (billRunId) {
     })
 }
 
+async function _licenceRef (licenceId) {
+  return LicenceModel.query()
```
robertparkinson, line 45: would it be better to extract the licenceRef here in the function and return it rather than having `licenceRef.licenceRef` ?

robertparkinson (changes requested):

robertparkinson (approved):

PR #732, opened 2024-02-13
Display source of supply in view licence page summary tab

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4327

This PR adds the source of the supply onto the summary page


app/models/licence.model.js
```diff
@@ -9,6 +9,8 @@ const { Model } = require('objection')
 
 const BaseModel = require('./base.model.js')
 
+const licenceRef = 'licences.licenceRef'
```
Cruikshanks, line 12: I know SonarCloud made you do this. But there are occasions where we are happy to tell SonarCloud to go 'do one' and this is one of them.

If we do it for one we should do it for all, and we think it just makes the models more complex and less maintainable.

Would you mind reverting this change? Ping me and I can silence SonarCloud when it moans again.

db/migrations/legacy/20240219152047_create-permit-licence-table.js
Cruikshanks, line 1: You've been caught out by my decision to completely re-jig things in the db/migrations/legacy folder.

Would you mind changing the file name to `db/migrations/legacy/20221108005001-permit-licence.js`. It will then match the new convention.

test/services/licences/view-licence.service.test.js
```diff
@@ -173,6 +174,130 @@ describe('View Licence service', () => {
...
+          purposes: [{
+            purposePoints: [{
+              point_source: {}
+            }]
+          }]
+        }
+        Sinon.stub(FetchLicenceService, 'go').resolves(fetchLicenceResult)
+      })
+
+      it('will return the source of supply for use in the licence summary page', async () => {
```
Cruikshanks, line 213: Just checking this, and the others like it underneath are correct?

It is returning a property but it is `null`. And we are not distinguishing the message from the test where a value is returned. Should it be something like

```suggestion
      it('will return null for the source of supply', async () => {
```

test/support/helpers/permit-licence.helper.js
```diff
@@ -0,0 +1,76 @@
...
+    licenceDataValue: null,
+    licenceSearchKey: null,
+    licenceStatusId: 1,
+    licenceTypeId: 8,
+    licenceRef: generateLicenceRef(),
+    licenceRegimeId: 1,
+    licenceSummary: null,
+    metadata: null,
+    publicDomain: null,
+    startDate: null
```
Cruikshanks, line 64: Again just checking. Normally if a field is nullable we don't bother setting a default. We just have these for fields which don't allow nulls but don't have a default, or where in 100% cases we expect something to be there.

If we strip the defaults that are set to null, do the tests still pass?

Cruikshanks (changes requested):
A couple of tweaks and then just some questions. Feel free to grab me if you want to get this over the line pronto!

Cruikshanks (approved):

PR #746, opened 2024-02-20
Returns required -  Select an existing return requirement from basic page

Demwunz:
This is a basic page layout for step 4 of 7 in the return requirements journey.

Validation will follow in a later ticket.

#WATER-4267 depends on this page.


app/routes/return-requirement.routes.js
```diff
@@ -120,6 +120,32 @@ const routes = [
...
+    method: 'POST',
+    path: '/return-requirements/{sessionId}/existing',
+    handler: ReturnRequirementsController.submitExisting,
+    options: {
+      auth: {
+        access: {
+          scope: ['billing']
+        }
+      },
+      description: 'Select an existing return requirement from'
```
Cruikshanks, line 146:
```suggestion
      description: 'Submit an existing return requirement from'
```

Cruikshanks (changes requested):
Just a NIT for the route description. But other than that it's all gravy!

Cruikshanks (approved):

PR #759, opened 2024-02-23
Update allocation engine to handle overdue returns

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4377

Update the match & allocation engine to ‘allocate’ overdue returns.

The complication is the engine is built on matching return submission lines to charge elements by comparing the line’s start and end dates to a charge element’s calculated abstraction dates (amongst a host of other parameters).

So where there is a licence with an overdue return that has been matched to an element. The engine needs to ensure that the customer is charged the authorised amount for that element.

The engine is to be updated so that after all completed returns have had their volumes allocated. If there are any overdue returns, the engine will attempt to match those to a charge element. If matched the engine will then bump the volume allocated to that element up to its authorised volume.

### DIFF
--- a/app/services/bill-runs/two-part-tariff/allocate-returns-to-charge-element.service.js
+++ b/app/services/bill-runs/two-part-tariff/allocate-returns-to-charge-element.service.js
@@ -58,7 +58,6 @@ function _allocateReturns (chargeElement, matchedReturn, chargePeriod, chargeRef
 
       chargeElement.allocatedQuantity += qtyToAllocate
       chargeElement.returnLogs[i].allocatedQuantity += qtyToAllocate
-
       matchedLine.unallocated -= qtyToAllocate
       matchedReturn.allocatedQuantity += qtyToAllocate
       chargeReference.allocatedQuantity += qtyToAllocate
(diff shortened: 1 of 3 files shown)


app/services/bill-runs/two-part-tariff/allocate-returns-to-charge-element.service.js
```diff
@@ -10,43 +10,32 @@ const { periodsOverlap } = require('../../../lib/general.lib.js')
 /**
  * For a chargeElement with matching returns any abstracted volume recorded on the return log will be allocated to the
  * charge element up to a maximum of the charge elements authorised volume, or the remaining authorised volume on the
- * charge reference, whichever is lower.
+ * charge reference, whichever is lower. If any "due" returns are are matched to a charge element, that element will
```
Cruikshanks, line 13: Just a typo (great explanation BTW)

```suggestion
 * charge reference, whichever is lower. If any "due" returns are matched to a charge element, that element will
```

Cruikshanks (changes requested):
I'll be honest, I didn't work through every single unit test. But this is fantastic.

The smallest of reasons to request a change!

app/services/bill-runs/two-part-tariff/allocate-returns-to-charge-element.service.js
Jozzey: Thanks. I'll sort the typo.

Cruikshanks (approved):

PR #770, opened 2024-02-28
View licence summary points of abstraction bug

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4330

View licence summary points of abstraction
Add braces around local name and use correct array for summary

### DIFF
--- a/app/presenters/licences/view-licence.presenter.js
+++ b/app/presenters/licences/view-licence.presenter.js
@@ -180,9 +180,6 @@ function _parseAbstractionsAndSourceOfSupply (permitLicence) {
 
   const abstractionPointsCaption = uniqueAbstractionPoints.length > 1 ? 'Points of abstraction' : 'Point of abstraction'
 
-  console.log(abstractionPoints)
-  console.log(uniqueAbstractionPoints)
-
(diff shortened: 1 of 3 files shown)


app/presenters/licences/view-licence.presenter.js
```diff
@@ -180,6 +180,9 @@ function _parseAbstractionsAndSourceOfSupply (permitLicence) {
 
   const abstractionPointsCaption = uniqueAbstractionPoints.length > 1 ? 'Points of abstraction' : 'Point of abstraction'
 
+  console.log(abstractionPoints)
+  console.log(uniqueAbstractionPoints)
```
Cruikshanks, line 184: If we do need to log stuff we have a different function for that. But I'm guessing these are just left over from debugging! 🐛🐞

Cruikshanks (changes requested):

app/presenters/licences/view-licence.presenter.js
robertparkinson: ah sorry about that! 😅

Cruikshanks (approved):

PR #771, opened 2024-02-28
Migrate legacy send bill run functionality

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4390

We have been working on replacing the legacy SROC annual billing engine. It is slow and prone to error leading to the team having to devote resources every year to helping the billing & data team get the annual bill runs through.

Like with our supplementary billing engine, we focused on the first part of the process; creating the bill run record and then generating the transactions to send to the [charging-module-api (CHA)](https://github.com/DEFRA/sroc-charging-module-api). With that done we then pass control back to the legacy code to refresh the data and mark the bill run as **READY**.

But our testing demonstrated that even with this part of the process replaced annual bill runs can still prove to be problematic due to the shoddy way the previous team implemented them. For example, when we send the POST request to the legacy `/refresh` endpoint it kicks off a process to request the 'generated' billing information from the CHA. This entails sending a request for every bill in the bill run and then updating our local data with the result.

You can't **SEND** a bill run until this step is complete. When you view the bill run all the values will be £0.00 and the download CSV will be working. Yet the legacy code still marks the bill run as **READY**. Users familiar with the service may know they need to leave it a while for the numbers to refresh. What they probably don't know is the CHA is getting hammered and the service is busy with this process.

Ideally, they should wait for this process to complete before starting a new bill run. But there is no way to know that. So, the next bill run will take longer to complete and you increase the risk of an error.

We felt this situation was ridiculous so opted to amend the [water-abstraction-service](https://github.com/DEFRA/water-abstraction-service) to solve this. (See [Stop marking bill runs as ready when they are not](https://github.com/DEFRA/water-abstraction-service/pull/2442)).

It worked but inadvertently broke the legacy 'send a bill run' process (the final step in the bill run journey). When you confirm and send a bill run it tells the CHA to send the bill run to SOP. The CHA does a range of things behind the scenes one of which is to generate a transaction reference for each invoice. Again we need that information on our side but only that.

This is where things go 🦇💩! The legacy service does this by calling the refresh invoices job again!! So, we're going through the whole process of sending a request for every invoice, updating values we already have including dealing with updating the transactions all over again. _All of this_ for something we can get making ***one*** request to the CHA (a request the legacy process already does before going through all the invoices again).

We didn't know the legacy code reused this job. So, when we amended it to set the bill run status to **READY** at the end of the process we broke sending because in that scenario we are setting the status back from **SENT** to **READY**.

We've implemented a 'hack' to get around this. But enough is enough! With this change, we take over the 'send bill run process' and resolve this properly.


Cruikshanks:
The SonarCloud issues are all to do with line length (and we're only a few lines over the 20 limit).

The key thing is this is now a working replacement that is _sooo much_ faster than the legacy version.

However, a key change from the legacy is we don't have the final **Bill run sent** page.

We've opted to not include it because it enforces the need for the 'spinner' page, something we are trying to quash with a burning passion!

We understand this is the end of the 'journey' for a bill run but we think we can make the argument to drop it and just redirect the user back to 'Bill Runs' which is where every other journey now finishes.

What'll help though is if we can get the bill runs page auto-refreshing. So, we are parking this change until we can get an opportunity to look into that.

Cruikshanks:
> [..] However, a key change from the legacy is we don't have the final **Bill run sent** page.
> 
> We've opted to not include it because it enforces the need for the 'spinner' page, something we are trying to quash with a burning passion!
> 
> We understand this is the end of the 'journey' for a bill run but we think we can make the argument to drop it and just redirect the user back to 'Bill Runs' which is where every other journey now finishes.
> 
> What'll help though is if we can get the bill runs page auto-refreshing. So, we are parking this change until we can get an opportunity to look into that.

We've changed our mind on this plan. We've got more pressing priorities than migrating the bill runs page. We don't want to leave to improvement festering so we are opting to use the same solution we were required to follow in [Migrate confirm remove licence from bill run page](https://github.com/DEFRA/water-abstraction-system/pull/828).

We'll redirect to the legacy processing page (the spinner!) and allow it to do what it would normally do once it sees the status change to `sent`.

PR #779, opened 2024-03-01
Add Charging Module Delete Bill Run Service

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4387

We need to add this to support our work to improve how bill runs are cancelled. Basically, replace the crummy legacy one!

When you cancel a bill run you need to send a [DELETE request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE) to the [Charging Module API](https://github.com/DEFRA/sroc-charging-module-api) to tell it to delete its copy.

Following our existing pattern we wrap interactions with the Charging Module in specific services. That way any calls to the ChargingModule can be centrally managed. It also makes it easier to stub the CHA in our tests.

This change adds a new `ChargingModuleDeleteBillRunService`.

### DIFF
--- a/app/lib/charging-module-request.lib.js
+++ b/app/lib/charging-module-request.lib.js
@@ -11,6 +11,23 @@ const requestConfig = require('../../config/request.config.js')
 const servicesConfig = require('../../config/services.config.js')
 
 /**
+ * Sends a DELETE request to the Charging Module for the provided path
+ *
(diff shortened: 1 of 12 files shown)


Cruikshanks:
Waiting on https://github.com/DEFRA/water-abstraction-system/pull/778 to be merged

robertparkinson (approved):

PR #790, opened 2024-03-05
Trigger two-part tariff annual bill run spike

Cruikshanks:

In the legacy service, there is no clear separation between two-part tariff annual billing and two-part tariff supplementary. The previous team, for the right reasons, tried to cover it all in one. Hindsight has shown it would be better to handle them separately. This is a wish from users as well as those of us who have to build it!

This means we are going to need to _update_ the 'Create a bill run' journey to handle the changes. And if you haven't guessed from how we do things here, what this _really_ means is we get to replace the whole thing! 😁

So, we intend to use this branch to spike out a working version. We'll then break up the changes into smaller PRs for review once we know how we want to do things.

### DIFF
--- a/app/controllers/bill-runs-create.controller.js
+++ b/app/controllers/bill-runs-create.controller.js
@@ -5,8 +5,11 @@
  * @module BillRunsCreateController
  */
 
+const Boom = require('@hapi/boom')
+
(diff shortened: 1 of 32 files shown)


Cruikshanks:
All the changes have been done and implemented. We can now discard the spike.

PR #802, opened 2024-03-11
Add bill runs setup type page to journey

Cruikshanks:

Having [Add first route to bill run setup journey](https://github.com/DEFRA/water-abstraction-system/pull/801) we can now add our first page which is where the user selects what bill run type they wish to create.

It is planned in the future that there will be 4; annual, supplementary, two-part tariff and two-part tariff supplementary. Until we have implemented two-part tariff supplementary though it is just the first 3 options.

Screenshot 2024-03-11 at 09 21 40

### DIFF
--- a/app/validators/bill-runs/setup/type.validator.js
+++ b/app/validators/bill-runs/setup/type.validator.js
@@ -10,8 +10,7 @@ const Joi = require('joi')
 const VALID_VALUES = [
   'annual',
   'supplementary',
-  'two_part_tariff',
-  'two_part_tariff_supplementary'
+  'two_part_tariff'
 ]
 
 /**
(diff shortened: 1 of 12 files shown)


app/validators/bill-runs/setup/type.validator.js
```diff
@@ -0,0 +1,42 @@
...
+ * @module TypeValidator
+ */
+
+const Joi = require('joi')
+
+const VALID_VALUES = [
+  'annual',
+  'supplementary',
+  'two_part_tariff',
+  'two_part_tariff_supplementary'
```
robertparkinson, line 14: This is not in the list of options on the page. is it going to be added later?

Cruikshanks: Good point. It is coming. But I couldn't be 100% certain that is what the name of the bill run type will be when we do have to implement it.

So, on reflection, it would be better that we add this as a valid bill run type when we make the changes to support it.

PR #805, opened 2024-03-12
Add bill runs setup financial year page to journey

Cruikshanks:

This adds the third page to the setup bill run journey; select the financial year. This like [Add bill runs setup region page to journey](https://github.com/DEFRA/water-abstraction-system/pull/804) has the complexity of having to direct the user to different pages depending on the year chosen.

The page only applies to two-part tariff bill runs and is only intended to be temporary. Two-part tariff bill runs were held during COVID then the SROC ones have been delayed whilst we build an SROC 2PT billing engine. This means we have a backlog so users need to be able to select the year they wish to create the bill run for. Normally, the current financial year would automatically be determined in the same way as annual bill runs. The years shown are determined by those with outstanding two-part tariff bill runs.

Once the backlog of bill runs is cleared this page can be dropped from the journey.

### DIFF
--- a/app/services/bill-runs/setup/submit-year.service.js
+++ b/app/services/bill-runs/setup/submit-year.service.js
@@ -1,0 +1,75 @@
+'use strict'
+
+/**
+ * Handles the user submission for the `/bill-runs/setup/{sessionId}/year` page
+ * @module SubmitYearService
+ */
+
+const SessionModel = require('../../../models/session.model.js')
+const YearPresenter = require('../../../presenters/bill-runs/setup/year.presenter.js')
+const YearValidator = require('../../../validators/bill-runs/setup/year.validator.js')
+
+/**
+ * Handles the user submission for the `/bill-runs/setup/{sessionId}/year` page
+ *
+ * It first retrieves the session instance for the setup bill run journey in progress. It then validates the payload of
+ * the submitted request.
+ *
+ * If there is no validation error it will save the selected value to the session then return an object with a
+ * `setupComplete:` property. If the year is not PRESROC this will be true. Else it will be false. The fact the property
+ * exists will tell the controller the submission was successful. Whether the property `setupComplete:` is true or false
+ * will be used to determine which page to direct the user to next.
+ *
+ * If there is a validation error it is combined with the output of the presenter to generate the page data needed to
+ * re-render the view with an error message.
+ *
+ * @param {string} sessionId - The UUID of the current session
+ * @param {Object} payload - The submitted form data
+ *
+ * @returns {Promise<Object>} An object with a `setupComplete:` property if there are no errors else the page data for
+ * the year page including the validation error details
+ */
+async function go (sessionId, payload) {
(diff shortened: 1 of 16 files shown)


app/services/bill-runs/setup/submit-year.service.js
```diff
@@ -0,0 +1,75 @@
...
+ */
+async function go (sessionId, payload) {
+  const session = await SessionModel.query().findById(sessionId)
+
+  const validationResult = _validate(payload)
+
+  if (!validationResult) {
+    await _save(session, payload)
+
+    return { setupComplete: ['2024', '2023'].includes(session.data.year) }
```
robertparkinson, line 40: will this only ever be for these years? or is this supposed to be the current and previous year so should be computed?

app/validators/bill-runs/setup/year.validator.js
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Validates data submitted for the `/bill-runs/setup/{sessionId}/year` page
+ * @module YearValidator
+ */
+
+const Joi = require('joi')
+
+const VALID_VALUES = [
+  '2024',
```
robertparkinson, line 11: will this require a code change every year to add in new options? should this list be computed?

app/views/bill-runs/setup/year.njk
```diff
@@ -0,0 +1,72 @@
...
+        fieldset: {
+          legend: {
+            text: pageTitle,
+            isPageHeading: true,
+            classes: 'govuk-fieldset__legend--l govuk-!-margin-bottom-6'
+          }
+        },
+        items: [
+          {
+            text: '2023 to 2024',
```
robertparkinson, line 47: should we be hard coding these? what will happen next year?

robertparkinson (changes requested):

app/services/bill-runs/setup/submit-year.service.js
Cruikshanks: From the PR description

> The page only applies to two-part tariff bill runs and is only intended to be temporary. Two-part tariff bill runs were held during COVID then the SROC ones have been delayed whilst we build an SROC 2PT billing engine. This means we have a backlog so users need to be able to select the year they wish to create the bill run for. Normally, the current financial year would automatically be determined in the same way as annual bill runs. The years shown are determined by those with outstanding two-part tariff bill runs.

app/validators/bill-runs/setup/year.validator.js
Cruikshanks: See the PR description! 😁

app/views/bill-runs/setup/year.njk
Cruikshanks: See the PR description! 😁

app/services/bill-runs/setup/submit-year.service.js
robertparkinson: ah sorry.. went straight to the code 😳

robertparkinson (approved):

PR #811, opened 2024-03-12
Submit Check your answers page

Demwunz:
https://eaflood.atlassian.net/browse/WATER-4264

This page allows the user to submit their answers.

The "check your answers" page will submit the data to the check your answers service which in turn will check the licence.

If the licence is valid, the user will be directed to the approved page.

If the licence is invalid (expired, lapsed or revoked), the service will throw an error.


app/services/return-requirements/check-your-answers.service.js
```diff
@@ -0,0 +1,28 @@
...
+const CheckYourAnswersPresenter = require('../../presenters/return-requirements/check-your-answers.presenter.js')
+const SessionModel = require('../../models/session.model.js')
+
+async function go (sessionId) {
+  const session = await SessionModel.query().findById(sessionId)
+  const formattedData = CheckYourAnswersPresenter.go(session)
+
+  return {
+    activeNavBar: 'search',
+    licence_id: session?.data?.licence?.licenceRef,
```
Cruikshanks, line 20: There might be a valid reason. But could this not also just be `licenceRef`?

```suggestion
    licenceRef: session?.data?.licence?.licenceRef,
```

app/services/return-requirements/check-your-answers.service.js
Cruikshanks, line 1: I'd avoid using the [Optional Chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) (sometimes known as the safe navigation operator) unless we have a situation where we know a property may not be populated.

In the case of this service, the session should be correctly populated. If it isn't it's because someone has purposely tried to access the page with a valid session ID but an invalid session. So, it's own them!

app/services/return-requirements/fetch-licence.service.js
Cruikshanks, line 1: Looking at `SubmitCheckYourAnswers` the only thing we need from this service is  `$ends`. In fact, we know `SubmitCheckYourAnswers` is going to get busier in the near future as it will be doing the work of persisting the return requirements.

So, it might be prudent to combine this service and `_checkLicenceEnded()` from `SubmitCheckYourAnswers` into a `CheckLicenceEndedService` that just returns a `true/false`. `SubmitCheckYourAnswers` can then use this result to decide whether to throw the error.

What do you think?

Cruikshanks (changes requested):
I've suggested merging `FetchLicenceService` and part of `SubmitCheckYourAnswers` into a new `CheckLicenceEnded` service that is also going to affect tests.

Happy to pair on this or jump on to get this over the line or if it's not clear what I'm suggesting 😜

app/services/return-requirements/check-your-answers.service.js
Demwunz: @Cruikshanks ok, that makes sense. I'll remove it and replace with a direct reference.

app/services/return-requirements/fetch-licence.service.js
Demwunz: @Cruikshanks It does make sense to youse the `$ends` property. I didn't want to jump the gun given the previous structure of JIRA tickets. I can make the above change you've suggested to save time later.

app/services/return-requirements/check-licence-ended.service.js
```diff
@@ -0,0 +1,62 @@
...
+  if (!ends) {
+    return false
+  }
+
+  const endDate = new Date(ends.date)
+  const today = new Date()
+  today.setHours(0, 0, 0, 0)
+
+  return endDate <= today
+}
```
Cruikshanks, line 58: As `$ends()` is a function rather than a property we wouldn't normally use destructuring to invoke the method.

And the object it returns (when there is an end date) contains a date. Did some testing and we shouldn't need to create a new date instance.

So, I believe after these changes you'll get the same result.

```suggestion
  const ends = licence.$ends()

  if (!ends) {
    return false
  }

  const today = new Date()
  today.setHours(0, 0, 0, 0)

  return ends.date <= today
}
```

app/services/return-requirements/check-your-answers.service.js
```diff
@@ -0,0 +1,28 @@
...
+ * @module CheckYourAnswersService
+ *
+ * @param {string} id - The UUID for return requirement setup session record
+ *
+ * @returns {Promise<Object>} page data needed by the view template
+ */
+const CheckYourAnswersPresenter = require('../../presenters/return-requirements/check-your-answers.presenter.js')
+const SessionModel = require('../../models/session.model.js')
+
+async function go (sessionId) {
```
Cruikshanks, line 14: Just some tidying up at the top of the module

```suggestion
'use strict'

/**
 * Orchestrates fetching and presenting the data for `/return-requirements/{sessionId}/check-your-answers` page
 * @module CheckYourAnswersService
 */

const CheckYourAnswersPresenter = require('../../presenters/return-requirements/check-your-answers.presenter.js')
const SessionModel = require('../../models/session.model.js')

/**
 * Orchestrates fetching and presenting the data for `/return-requirements/{sessionId}/check-your-answers` page
 *
 * @param {string} sessionId - The UUID for return requirement setup session record
 *
 * @returns {Promise<Object>} page data needed by the view template
 */
async function go (sessionId) {
```

app/services/return-requirements/submit-check-your-answers.service.js
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @param {string} sessionId - The id of the current session
+ *
+ * @returns {string} The licence id
+ */
+const CheckLicenceEndedService = require('./check-licence-ended.service.js')
+const SessionModel = require('../../models/session.model.js')
+const ExpandedError = require('../../errors/expanded.error.js')
+
+async function go (sessionId) {
```
Cruikshanks, line 18: Just some tidying up at the top of the module

```suggestion
'use strict'
/**
 * Manages converting the session data to return requirement records when check your answers is confirmed
 * @module SubmitCheckYoursAnswersService
 */

const CheckLicenceEndedService = require('./check-licence-ended.service.js')
const ExpandedError = require('../../errors/expanded.error.js')
const SessionModel = require('../../models/session.model.js')

/**
 * Manages converting the session data to return requirement records when check your answers is confirmed
 * 
 * > This service is work in progress. Some of the functionality described is yet to be implemented
 * 
 * After fetching the session instance for the returns requirements journey in progress it validates that what the user
 * has setup can be persisted for the licence.
 * 
 * If valid it converts the session data to return requirements records then deletes the session record.
 *
 * @param {string} sessionId - The UUID for return requirement setup session record
 *
 * @returns {string} The licence ID
 */
async function go (sessionId) {
```

Cruikshanks (changes requested):
I've not gone through the tests or view yet because I think some things need tidying up in the services first.

Let's deal with those as they'll impact the tests then ping me and I'll check the rest.

Cruikshanks (approved):

PR #818, opened 2024-03-14
Fix matching logic in bill run setup

Cruikshanks:

Having completed the migration of the setup bill run journey we deployed to our test environment and did some initial checking.

In that environment **Wales** had a two-part tariff bill run in a state of `READY` but nothing else. So, when trying to create a supplementary bill run we expected to see both an SROC and PRESROC bill run created.

However, only the SROC one was created. Debugging confirmed we were sending the request to the legacy service but it was returning a `409 - conflict`. Further checking confirmed that it was seeing the two-part tariff as an existing bill run that should block the supplementary from being processed.

"Well legacy is wrong!" was our immediate response. But on reflection, it makes perfect sense. The idea of a supplementary is to recalculate charges to a customer because a change has been made. But it can only do that if the existing bill runs it is looking at are also not still being processed.

Double-checking the legacy code we confirmed they take a simple approach to the matching. They look for _any_ live bill runs for the specific region and year, and then for any exact matches.

Irrespective of the bill run being created, if there are any live bill runs it will block the creation of the new one. If there are no live matches, supplementary can proceed. In the case of annual and two-part tariff if there are no exact matches as well, it can proceed.

So, in this change, we are adding the ability to look for live bill runs as well as matching bill runs. We're then incorporating that into our `ExistsService` to determine which match to show, if any.

### DIFF
--- a/app/services/bill-runs/setup/bill-run-match.service.js
+++ b/app/services/bill-runs/setup/bill-run-match.service.js
@@ -1,0 +1,114 @@
+'use strict'
+
+/**
+ * Determines if an existing bill run matches the one a user is trying to setup
+ * @module ExistsService
+ */
+
+const FetchLiveBillRunsService = require('./fetch-live-bill-runs.service.js')
+const FetchMatchingBillRunService = require('./fetch-matching-bill-run.service.js')
+
+async function go (regionId, batchType, financialEndYear, season = null) {
+  const supplementary = batchType === 'supplementary'
+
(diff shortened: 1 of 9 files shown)


Cruikshanks:
I've properly jumped feet-first into a quagmire of weird rules! 🥾😱

To try and explain it I've probably gone OTT with the comments. So, if this is too big to review or you need a walk-through first do give me a ping.

robertparkinson (approved):

PR #823, opened 2024-03-15
Fix std charge transaction presenter for S130

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4409

Our users reported that for licences where the Canal and Rivers Trust agreement (S130) has been applied, it isn't displayed against the transaction in the new bill view.

We assumed this was just another thing we missed when building the new views. So, we were surprised when we found there is code to handle displaying the Canal and rivers trust agreement.

After doing some digging we found out why 🙄🤦

water-4409_cause

We'd already clocked that for some reason the previous team had made this a text field rather than a boolean, even though we are using it to store true or false. We hadn't clocked was that when the legacy code sets the field it includes a space at the end. Computers being computers means 'true' !== 'true '. 😡 

This change updates the code to handle this.


app/presenters/bill-licences/view-standard-charge-transaction.presenter.js
```diff
@@ -174,6 +176,7 @@ function _presrocContent (transaction) {
 }
 
 function _srocContent (transaction) {
+  console.log('🚀 ~ _srocContent ~ transaction:', transaction)
```
Jozzey, line 179: Think you forgot to remove your `console.log`

Jozzey (changes requested):
Just a `console.log` to remove

app/presenters/bill-licences/view-standard-charge-transaction.presenter.js
Cruikshanks: Doh! Good spot!

robertparkinson (approved):

Jozzey (approved):

PR #828, opened 2024-03-16
Migrate confirm remove licence from bill run page

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4410

> Part of a series of changes to add support for removing a bill licence back into the billing views

We're not ready to implement removing a bill licence from a bill run as it depends on sending requests to the [Charging Module API](https://github.com/DEFRA/sroc-charging-module-api) and running the legacy refresh job to update everything on our side.

But we can migrate the confirmation page a user sees when removing a licence. Not only is that one less thing in the legacy code but we can also update the view to be consistent with others we've done like

- [Migrate view errored bill run page](https://github.com/DEFRA/water-abstraction-system/pull/785)
- [Migrate view empty bill run page](https://github.com/DEFRA/water-abstraction-system/pull/783)
- [Add cancel bill run page](https://github.com/DEFRA/water-abstraction-system/pull/780)

With this in place, we can then fix our billing views and make the [Remove licence button](https://github.com/DEFRA/water-abstraction-system/pull/824) visible again!

Screenshot 2024-03-16 at 14 37 10

### DIFF
--- a/app/controllers/bill-licences.controller.js
+++ b/app/controllers/bill-licences.controller.js
@@ -5,7 +5,19 @@
  * @module BillLicencesController
  */
 
+const RemoveBillLicenceService = require('../services/bill-licences/remove-bill-licence.service.js')
 const ViewBillLicenceService = require('../services/bill-licences/view-bill-licence.service.js')
+
+async function remove (request, h) {
+  const { id } = request.params
+
+  const pageData = await RemoveBillLicenceService.go(id)
+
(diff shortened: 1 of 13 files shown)


Cruikshanks:
SonarCloud's duplication flagging is valid. We are now running the same query to support determining the billing account name in a few places.

However, we intend to tackle this refactoring in a future PR.

PR #833, opened 2024-03-18
Select the points for the requirements for returns page

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4295

This is PR is focused on adding functionality to page 2 of 11 in the return requirement pages.

The pages were previously setup as part of a separate PR to create the user journey.


app/services/return-requirements/submit-points.service.js
```diff
@@ -0,0 +1,58 @@
...
+const SessionModel = require('../../models/session.model.js')
+
+/**
+ * Orchestrates validating the data for `/return-requirements/{sessionId}/points` page
+ *
+ * It first retrieves the session instance for the returns requirements journey in progress.
+ *
+ * The user input is then validated and the result is then combined with the output of the presenter to generate the page data needed by the view.
+ * If there was a validation error the controller will re-render the page so needs this information. If all is well the
+ * controller will redirect to the next page in the journey.
```
Cruikshanks, line 20: Looks like some of the lines have gone over our 120 char line limit. Could also do with a bit of a tidy! 😁 

```suggestion
 * Orchestrates validating the data for `/return-requirements/{sessionId}/points` page
 *
 * It first retrieves the session instance for the returns requirements journey in progress.
 *
 * The user input is then validated and the result is then combined with the output of the presenter to generate the
 * page data needed by the view. If there was a validation error the controller will re-render the page so needs this
 * information. If all is well the controller will redirect to the next page in the journey.
```

I highly recommend the VSCode extension [Rewrap](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) to keep a handle on comments.

app/services/return-requirements/submit-points.service.js
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * @param {string} sessionId - The id of the current session
```
Cruikshanks, line 22:
```suggestion
 * @param {string} sessionId - The UUID of the current session
```

app/validators/return-requirements/point.validator.js
Cruikshanks, line 1: I believe the file name should be `points.validator.js`.

app/validators/return-requirements/point.validator.js
```diff
@@ -0,0 +1,45 @@
...
+ * Validates data submitted for the `/return-requirements/{sessionId}/points` page
+ * @module PointsValidator
+ */
+
+const Joi = require('joi')
+
+/**
+ * Validates data submitted for the `/return-requirements/{sessionId}/points` page
+ *
+ * When setting up a requirement users must specify a points for the return requirement.
```
Cruikshanks, line 13: Typo!

```suggestion
 * When setting up a requirement users must specify points for the return requirement.
```

app/validators/return-requirements/point.validator.js
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @returns {Object} The result from calling Joi's schema.validate(). If any errors are found the
+ * `error:` property will also exist detailing what the issue is.
+ */
+function go (payload) {
+  let points = payload.points
+
+  if (!Array.isArray(points)) {
+    points = [points]
+  }
```
Cruikshanks, line 27: If we are always validating the payload from the request, when would points not be an array?

I guess I'm asking when the page is submitted, what is the scenario where this is needed? And if I'm asking it (and it is perfectly valid) then it might be worth a `// NOTE: `` to explain why this bit of code exists.

app/validators/return-requirements/point.validator.js
```diff
@@ -0,0 +1,45 @@
...
+
+  const errorMessage = 'Select any points for the return requirement'
```
Cruikshanks, line 29: I like this!

Should probably be doing it in our other validators if only to get SonarCloud off our backs!

app/views/return-requirements/points.njk
```diff
@@ -15,14 +18,51 @@
...
+      ]
+    }) }}
+  {%endif%}
+
   {# Main heading #}
-  <div class="govuk-body">
+  <div>
+    <span class="govuk-caption-l"> Licence {{ licenceRef }} </span>
     <h1 class="govuk-heading-xl govuk-!-margin-bottom-3">{{ pageTitle }}</h1>
+
```
Cruikshanks, line 38: Probably one too many new lines here I thinks!
```suggestion
```

app/views/return-requirements/points.njk
```diff
@@ -15,14 +18,51 @@
...
+      {% endfor %}
+        {{ govukCheckboxes({
+          name: "points",
+          errorMessage: error,
+          items: checkBoxItems
+        }) }}
+
+    {{ govukButton({ text: "Continue" }) }}
   </form>
+</div>
```
Cruikshanks, line 65: It feels like the alignment has got a little out of whack here. The `<div>` isn't inline with the section above and the elements inside the form appear to be double indented.

I'd suggest giving it another look.

app/views/return-requirements/points.njk
```diff
@@ -15,14 +18,51 @@
...
+
+
```
Cruikshanks, line 67: I don't think we need any empty lines at the end of the section here. If this was code StandardJS would shout at us!
```suggestion
```

test/controllers/return-requirements.controller.test.js
```diff
@@ -151,12 +152,17 @@ describe('Return requirements controller', () => {
   })
 
   describe('GET /return-requirements/{sessionId}/points', () => {
+    beforeEach(async () => {
+      Sinon.stub(SelectPointsService, 'go').resolves({
+        id: '8702b98f-ae51-475d-8fcc-e049af8b8d38', pageTitle: 'Select the points for the requirements for returns'
+      })
+    })
```
Cruikshanks, line 159: Having removed lines above, here we need one!
```suggestion
    })

```

test/services/return-requirements/fetch-points.service.test.js
```diff
@@ -0,0 +1,83 @@
...
+      NGR1_NORTH: '50394',
+      NGR1_SHEET: 'TQ',
+      NGR2_NORTH: 'null',
+      NGR2_SHEET: 'null',
+      NGR3_NORTH: 'null',
+      NGR3_SHEET: 'null',
+      NGR4_NORTH: 'null',
+      NGR4_SHEET: 'null'
+    }])
+  })
```
Cruikshanks, line 82: I know there isn't much to these services. However, we do try to make the output more readable for documentation purposes. So, I'd suggest looking at some of the other fetch services. For example, `test/services/bill-runs/setup/fetch-regions.service.test.js`

```javascript
describe('Bill Runs Setup Fetch Regions service', () => {
  describe('when called', () => {
    it('returns the ID and display name for each region ordered by display name', async () => {
    })
  })
})
```

test/services/return-requirements/submit-points.service.test.js
```diff
@@ -0,0 +1,185 @@
...
+            LOCAL_NAME: 'BEWL WATER RESERVOIR',
+            NGR1_NORTH: '33604',
+            NGR1_SHEET: 'TQ',
+            NGR2_NORTH: 'null',
+            NGR2_SHEET: 'null',
+            NGR3_NORTH: 'null',
+            NGR3_SHEET: 'null',
+            NGR4_NORTH: 'null',
+            NGR4_SHEET: 'null'
+          }
```
Cruikshanks, line 87: It looks like this data is used in a couple of places. And because of its size it tends to push the tests down when that is what we're trying to work with.

If you scout our code base in cases like this, you'll find we often put a function at the end of the tests to return test data. That way we remove the duplication and convert things like setting the resolver in a stub to be a one-liner.

I'd suggest doing that here.

Cruikshanks (changes requested):
Awesome work! If we've been doing our job properly the whole `Get-view-Submit-Validate` pattern in our code should start to become clear and you can see how other pages are doing the same thing.

A few things to clear up, and don't be concerned about pining me if all I've done is confuse things or more likely, got it completely wrong!

app/services/return-requirements/submit-points.service.js
```diff
@@ -6,7 +6,7 @@
  */
 
 const FetchPointsService = require('../../services/return-requirements/fetch-points.service.js')
-const PointValidation = require('../../validators/return-requirements/point.validator.js')
+const PointsValidation = require('../../validators/return-requirements/points.validator.js')
```
Cruikshanks, line 9: NIT

```suggestion
const PointsValidator = require('../../validators/return-requirements/points.validator.js')
```

test/services/return-requirements/submit-points.service.test.js
```diff
@@ -14,7 +14,7 @@ const SessionHelper = require('../../support/helpers/session.helper.js')
 
 // Things we need to stub
 const FetchPointsService = require('../../../app/services/return-requirements/fetch-points.service.js')
-const PointsValidation = require('../../../app/validators/return-requirements/point.validator.js')
+const PointsValidation = require('../../../app/validators/return-requirements/points.validator.js')
```
Cruikshanks, line 17: NIT

```suggestion
const PointsValidator = require('../../../app/validators/return-requirements/points.validator.js')
```

Cruikshanks (changes requested):
Just some _really_ minor nit-picks left!

Cruikshanks (approved):

PR #835, opened 2024-03-18
Filter licences by status on the 2PT Review Licences Page

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4217

Add a filter to the two-part tariff Review Licences page to allow the user to filter by license status.

### DIFF
--- a/app/services/bill-runs/two-part-tariff/fetch-bill-run-licences.service.js
+++ b/app/services/bill-runs/two-part-tariff/fetch-bill-run-licences.service.js
@@ -23,11 +23,14 @@ const ReviewLicenceModel = require('../../../models/review-licence.model.js')
  */
 async function go (id, payload) {
   const filterLicenceHolder = payload?.filterLicenceHolder
+  const filterLicenceStatus = payload?.filterLicenceStatus
+
+  const filterData = { filterLicenceHolder, filterLicenceStatus }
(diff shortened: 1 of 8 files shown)


app/services/bill-runs/two-part-tariff/fetch-bill-run-licences.service.js
Cruikshanks, line 1: The one change I'd suggest making is where `filterData` gets put together.

We're passing in the data for it from `review-bill-run.service.js`. I reckon that's a job that is better suited to being _in_ `review-bill-run.service.js`.

If then the function signature for `fetch-bill-run-licences.service.js` is updated to `(id, licenceHolder = null, status = null)` this service becomes a bit cleaner and what the args are is clearer.

We also avoid passing the payload on further than we need to. I hope that makes sense?

Jozzey: Agreed, that would make it clearer. I'll make that change.

Jozzey: Made some changes and renamed a load of stuff. Hopefully a bit nicer now :grin:

Cruikshanks (approved):

PR #837, opened 2024-03-19
Migrate confirm remove bill from bill run page

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4416

We'd recently spotted that the new billing views we'd developed had dropped a 'Remove licence' button. So, we used that opportunity to [Migrate confirm remove licence from bill run page](https://github.com/DEFRA/water-abstraction-system/pull/828) and [added the button back in](https://github.com/DEFRA/water-abstraction-system/pull/824).

But then our conscientious QA team started testing it and found it didn't do what it claimed to do. That was to flag the removed licence for the next supplementary bill run. We display the same message on the confirm remove bill from bill run so now doubtful they tested that as well with the same result.

We've [fixed the remove licence issue](https://github.com/DEFRA/water-abstraction-service/pull/2464). But whilst we fix the legacy code for remove bill run this is a great opportunity to also migrate the confirm remove bill from bill run page. Not only is it one less thing in the legacy code but we can also update the view to be consistent with others we've done like

- [Migrate confirm remove licence from bill run page](https://github.com/DEFRA/water-abstraction-system/pull/828)
- [Migrate view errored bill run page](https://github.com/DEFRA/water-abstraction-system/pull/785)
- [Migrate view empty bill run page](https://github.com/DEFRA/water-abstraction-system/pull/783)
- [Add cancel bill run page](https://github.com/DEFRA/water-abstraction-system/pull/780)

Screenshot 2024-03-19 at 22 40 16

### DIFF
--- a/app/controllers/bills.controller.js
+++ b/app/controllers/bills.controller.js
@@ -5,7 +5,34 @@
  * @module BillsController
  */
 
+const Boom = require('@hapi/boom')
+
+const RemoveBillService = require('../services/bills/remove-bill.service.js')
+const SubmitRemoveBillService = require('../services/bills/submit-remove-bill.service.js')
 const ViewBillService = require('../services/bills/view-bill.service.js')
+
+async function remove (request, h) {
+  const { id } = request.params
+
+  const pageData = await RemoveBillService.go(id)
+
+  return h.view('bills/remove.njk', {
+    activeNavBar: 'bill-runs',
+    ...pageData
+  })
+}
+
+async function submitRemove (request, h) {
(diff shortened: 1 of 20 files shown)


Cruikshanks:
I'm not going to close the issues SonarCloud has found this time. It is correct on the duplication. We need to remove it but we intend to do that in a separate PR (this is related to a fix we need to get shipped!)

On the function length, we will need to tell SonarCloud to accept it. But it would be better to do that with our refactored reusable fetch services and presenters rather than in this specific case.

PR #848, opened 2024-03-21
View licence summary abstraction conditions bugs

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4332

As part of the testing of this QA pointed out that the more complicated licences did not match the legacy system. After more investigating it was discovered that the legacy code is using a part of the permit licence JSON blob to determine unique abstraction conditions. I found that this can be replicated using the purposeUseId in the `licence_version_purposes` table


app/presenters/licences/view-licence.presenter.js
```diff
@@ -92,11 +92,12 @@ function _generateAbstractionConditions (conditionsData) {
...
-      : 'View details of the abstraction condition',
-    conditions: conditionsData.abstractionConditions
+    caption: conditionsData.numberOfAbstractionConditions === 1 ? 'Abstraction condition' : 'Abstraction conditions',
+    conditions: conditionsData.uniqueAbstractionConditions,
+    linkText: conditionsData.numberOfAbstractionConditions === 1
+      ? 'View details of the abstraction condition'
+      : 'View details of the abstraction conditions',
+    numberOfAbstractionConditions: conditionsData.numberOfAbstractionConditions
   }
 }
```
Cruikshanks, line 102: Would this simplify things a little in the return statement?

```javascript

  const conditionText = conditionsData.numberOfAbstractionConditions === 1 ? 'condition' : 'conditions'

  return {
    caption: `Abstraction ${conditionText}`,
    conditions: conditionsData.uniqueAbstractionConditions,
    linkText: `View details of the abstraction ${conditionText}`,
    numberOfAbstractionConditions: conditionsData.numberOfAbstractionConditions
  }
```

app/services/licences/fetch-licence-version-purpose-condition.service.js
```diff
@@ -10,45 +10,74 @@ const LicenceVersionPurposeCondtionModel = require('../../models/licence-version
...
+    numberOfAbstractionConditions: abstractionConditions.length,
+    uniqueAbstractionConditions: [...new Set(abstractionConditionDisplayTitles)]
   }
 }
 
-async function _fetchLicenceVersionPurposeCondition (id) {
+async function _fetchLicenceVersionPurposeCondition (ids) {
   const result = await LicenceVersionPurposeCondtionModel.query()
-    .where('licenceVersionPurposeId', id)
+    .whereIn('licenceVersionPurposeId', ids)
```
Cruikshanks, line 77: I know it's not in this PR but just spotted there is a typo in the model name `..Condtion...` not `...Condition...`.

Worth fixing as we are amending the file in this PR.

app/services/licences/fetch-licence-version-purpose-condition.service.js
```diff
@@ -10,45 +10,74 @@ const LicenceVersionPurposeCondtionModel = require('../../models/licence-version
...
   }
-  const LicenceVersionPurposeCondition = await _fetchLicenceVersionPurposeCondition(id)
-  const data = await _data(LicenceVersionPurposeCondition)
+
+  const licenceVersionIds = []
+
+  ids.forEach((item) => { licenceVersionIds.push(item.licenceVersionPurposeId) })
+
+  const LicenceVersionPurposeCondition = await _fetchLicenceVersionPurposeCondition(licenceVersionIds)
+  const data = await _data(LicenceVersionPurposeCondition, ids)
```
Cruikshanks, line 30: Should have spotted this in previous PRs.

We don't normally capitilise variable names is the first thing. But also as we don't have a `limt(1).first()` in the query in fetch we'll be getting an array back.

The function we're calling has it right!

```suggestion
  const licenceVersionPurposeConditions = await _fetchLicenceVersionPurposeCondition(licenceVersionIds)
  const data = await _data(licenceVersionPurposeConditions, ids)
```

app/services/licences/fetch-licence-version-purpose-condition.service.js
```diff
@@ -10,45 +10,74 @@ const LicenceVersionPurposeCondtionModel = require('../../models/licence-version
...
 
   return data
 }
 
-async function _data (licenceVersionPurposeConditions) {
+async function _data (licenceVersionPurposeConditions, licenceVersionPuroseAndPurposeUseIds) {
```
Cruikshanks, line 35: Typo!

```suggestion
async function _data (licenceVersionPurposeConditions, licenceVersionPurposeAndPurposeUseIds) {
```

But I also think this function has grown too large and is doing too much. I would suggest moving the logic inside the loop into its own function that we call from here. That will also allow us to name the function something meaningful that might help explain what is going on (my little brain gets lost easily with what we're doing here!)

app/services/licences/fetch-licence-version-purpose-condition.service.js
```diff
@@ -10,45 +10,74 @@ const LicenceVersionPurposeCondtionModel = require('../../models/licence-version
...
  * Fetch the matching licence version purpose conditions and return data needed for the view licence page
  *
- * @param {string} id The UUID for the licence_version_purpose_id to fetch
+ * @param {object} ids The object contains the UUIDs for the licence_version_purpose_id and the purpose_id
  *
  * @returns {Promise<Object>} the data needed to populate the view licence page and some elements of the summary tab
  */
-async function go (id) {
-  if (!id) {
+async function go (ids) {
```
Cruikshanks, line 17: I think if this was just an array of UUIDs i.e. strings then ids would be ok (I'd probably still try and come up with some more descriptive 😛)

But we give it a more descriptive name when it finally gets to the function that uses it; `licenceVersionPurposeAndPurposeUseIds`. So, I'd suggest renaming the param to that 🤷

app/services/licences/fetch-licence-version-purpose-condition.service.js
```diff
@@ -10,45 +10,74 @@ const LicenceVersionPurposeCondtionModel = require('../../models/licence-version
 /**
  * Fetch the matching licence version purpose conditions and return data needed for the view licence page
  *
- * @param {string} id The UUID for the licence_version_purpose_id to fetch
+ * @param {object} ids The object contains the UUIDs for the licence_version_purpose_id and the purpose_id
```
Cruikshanks, line 13: If it's an array of something you can do this with the JSDoc

```suggestion
 * @param {Object[]} ids - The object contains the UUIDs for the `licence_version_purpose_id` and the `purpose_id`
```

app/services/licences/fetch-licence-version-purpose-condition.service.js
```diff
@@ -10,45 +10,74 @@ const LicenceVersionPurposeCondtionModel = require('../../models/licence-version
...
+
   return {
-    abstractionConditions: [...new Set(abstractionConditions)]
+    numberOfAbstractionConditions: abstractionConditions.length,
+    uniqueAbstractionConditions: [...new Set(abstractionConditionDisplayTitles)]
   }
 }
 
-async function _fetchLicenceVersionPurposeCondition (id) {
+async function _fetchLicenceVersionPurposeCondition (ids) {
```
Cruikshanks, line 75: As this returns an array rather than a single object we should probably pluralise the method name.

```suggestion
async function _fetchLicenceVersionPurposeConditions (ids) {
```

app/services/licences/fetch-licence-version-purpose-condition.service.js
Cruikshanks, line 77: I know the Objection docs have a habit of putting the 'where' clauses first. But we prefer to put them after the `select()` as this matches what you would do if it was SQL.

app/services/licences/fetch-licence-version-purpose-condition.service.js
```diff
@@ -10,45 +10,74 @@ const LicenceVersionPurposeCondtionModel = require('../../models/licence-version
...
+    }
   }
 
-  licenceVersionPurposeConditions.forEach((purposeConditions) => {
-    purposeConditions.licenceVersionPurposeConditionTypes.forEach((licenceVersionPurposeConditionType) => {
-      abstractionConditions.push(licenceVersionPurposeConditionType.displayTitle)
+  licenceVersionPurposeConditions.forEach((purposeCondition) => {
+    // Get the purposeUseId from the LicenceVersionPurpose table using the licenceVersionPurposeId of the current purposeCondition
+    const { purposeId: purposeUseId } = licenceVersionPuroseAndPurposeUseIds.find((item) =>
+      item.licenceVersionPurposeId === purposeCondition.licenceVersionPurposeId)
```
Cruikshanks, line 48: We have a team convention that states we always use block body over concise for [arrow functions](https://github.com/DEFRA/water-abstraction-team/blob/main/coding_conventions.md#arrow-functions) even if it is a one-liner.
```suggestion
    const { purposeId: purposeUseId } = licenceVersionPurposeAndPurposeUseIds.find((item) => {
      return item.licenceVersionPurposeId === purposeCondition.licenceVersionPurposeId
    })
```

app/services/licences/view-licence.service.js
```diff
@@ -26,8 +26,14 @@ async function go (id) {
...
-    const licenceVersionPurposeId = licenceData?.licenceVersions[0]?.licenceVersionPurposes[0]?.id
-    licenceVersionPurposeConditionData = await FetchLicenceVersionPurposeConditionService.go(licenceVersionPurposeId)
+    const licenceVersionPurposeIds = []
+    licenceData?.licenceVersions[0]?.licenceVersionPurposes.forEach((licenceVersionPurpose) => {
+      licenceVersionPurposeIds.push({
+        licenceVersionPurposeId: licenceVersionPurpose.id,
+        purposeId: licenceVersionPurpose.purposeId
+      })
+    })
+    licenceVersionPurposeConditionData = await FetchLicenceVersionPurposeConditionService.go(licenceVersionPurposeIds)
```
Cruikshanks, line 36: I think this addition means it is time to move preparing the `licenceVersionPurposeConditionData` into its own private function within the service rather than doing it directly in `go()`.

Cruikshanks (changes requested):
I haven't gotten into the tests because I've spotted a few things like typos but also have made some suggestions.

If it's okay I'll focus on those after we've decided whats changes we'll implement as they are sure to be edited as well!

PR #853, opened 2024-03-23
Fix for view licence abstraction conditions V2

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4332

> @Cruikshanks stab at the problem!

As part of the testing for the new view licence summary page, QA pointed out that the more complicated licences did not match the legacy system. We would see different results for the 'Abstraction conditions' section.

The first thing we'd overlooked was the need to ensure we were only looking at the 'current' licence version. After more investigation, we then discovered that the permit licence JSON blob the legacy code is using is replicated in tables that already exist in the `water` schema. We found we could access the same data using these tables to get a matching result to the legacy view.

This change adds a new service to fetch the abstraction data we need and updates the presenter to generate the distinct list of conditions to display.


app/services/licences/fetch-licence-abstraction-conditions.service.js
```diff
@@ -0,0 +1,68 @@
...
+
+const LicenceVersionPurposeModel = require('../../models/licence-version-purpose.model.js')
+
+/**
+ * Fetches a licence's abstraction conditions needed for the summary tab on the view '/licences/{id}` page
+ *
+ * Each time a licence is changed in the upstream NALD system a new 'version is generated, the latest of which becomes
+ * the 'current' version.
+ *
+ * Linked each version could be multiple purposes for the abstraction, for example, spray irrigation and mineral
```
robertparkinson, line 16: NIT but I think you're missing a 'to' here?

app/services/licences/fetch-licence-abstraction-conditions.service.js
```diff
@@ -0,0 +1,68 @@
...
+ * washing.
+ *
+ * Then each purpose might have one or more conditions attached to it. Where things get complex is the same condition
+ * can be attached to more than one purpose and we're trying to show a distinct list of abstraction conditions for the
+ * licence. But we still need a count of them all!
+ *
+ * This service is able to determine the purposes for the current licence version and their conditions then extract the
+ * display title for each one. The total number of results gives us the abstraction condition count we need. As there
+ * may be duplicates in the titles we process the results to determine the distinct list needed. Finally, just in
+ * we also provide a distinct list of the purpose IDs :-)
```
robertparkinson, line 26: I think you're missing a 'case' from here?

robertparkinson (changes requested):

robertparkinson (approved):

PR #865, opened 2024-03-27
Exclude ended charge versions from 2PT billing

Cruikshanks:

In recent testing of triggering an SROC two-part tariff bill run, we've seen errors being thrown by the `DetermineAbstractionPeriodService`.

After investigating it is not the issue. It was being passed an empty charge period and this was because we're including charge versions that end before the billing period starts. This causes `DetermineChargePeriodService` to return an empty charge period.

This change updates `FetchChargeVersionsService` for two-part tariff to exclude charge versions that ended before the billing period starts.


Cruikshanks:
SonarCloud's issues with duplication are beyond the scope of this change.

Beckyrose200 (approved):

PR #867, opened 2024-03-27
Fix licence issues failing to insert for review

Cruikshanks:

In recent testing of triggering an SROC two-part tariff bill run, we've seen errors thrown by the `PersistAllocatedLicenceToResultsService`.

After investigating it is not the issue. It was being asked to persist every issue for every element and return found by `DetermineLicenceIssuesService` in its `issues` field.

No de-duping of the issues is taking place so it's blowing past the 255-char limit for the DB field for some licences.

This causes the bill run to fail and the system app to crash (but we'll deal with that in another fix!)

This change will de-dupe the issues found _before_ they are persisted to keep us under the 255 char limit.

### DIFF
--- a/app/services/bill-runs/two-part-tariff/determine-licence-issues.service.js
+++ b/app/services/bill-runs/two-part-tariff/determine-licence-issues.service.js
@@ -36,7 +36,7 @@ async function go (licence) {
   const allElementIssues = _determineChargeElementsIssues(chargeVersions, licenceReturnLogs)
 
   licence.status = _determineLicenceStatus(allElementIssues, allReturnIssues)
-  licence.issues = [...allElementIssues, ...allReturnIssues]
+  licence.issues = _licenceIssues(allElementIssues, allReturnIssues)
 }
 
 function _determineChargeElementsIssues (chargeVersions, licenceReturnLogs) {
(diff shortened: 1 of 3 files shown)


Beckyrose200 (approved):

Beckyrose200 (changes requested):
If the maximum the issues string can be in the DB is all the issues once then the string would look like this 

'Abstraction outside period, Aggregate, Checking query, No returns received, Over abstraction, Overlap of charge dates, Returns received but not processed, Returns received late, Return split over charge references, Some returns not received, Unable to match return'

This still surpasses the 255 character limit on the db, so I think we need to change that to be bigger or drop it completely?

Jozzey:
> If the maximum the issues string can be in the DB is all the issues once then the string would look like this
> 
> 'Abstraction outside period, Aggregate, Checking query, No returns received, Over abstraction, Overlap of charge dates, Returns received but not processed, Returns received late, Return split over charge references, Some returns not received, Unable to match return'
> 
> This still surpasses the 255 character limit on the db, so I think we need to change that to be bigger or drop it completely?

I think we should just drop the constraints on these `varchar` columns. They were added unintentionally and if you look at the rest of the DB there are no such constraints on any of the other `varchar` columns (not that I've checked them all).

Jozzey (approved):

Beckyrose200 (approved):

PR #871, opened 2024-03-28
View match details for an element during 2PT review

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4195

As part of our two-part tariff bill run work, we've developed a review page displaying a summary of all licences in a bill run. From there you can view the match details of a specific charge element to any returns it has matched or unmatched. This PR is the journey for the match details page.


app/services/bill-runs/two-part-tariff/fetch-view-match-details.service.js
Cruikshanks, line 1: I'd suggest a slight tweak to the file and service name; `fetch-match-details.service.js`.

We haven't referenced that we're fetching for a view vs general logic in other fetch services so we can save ourselves some typing here.

app/services/bill-runs/two-part-tariff/fetch-match-details.service.js
```diff
@@ -0,0 +1,70 @@
...
+const ReviewChargeElementModel = require('../../../models/review-charge-element.model.js')
+
+/**
+ * Fetches the match details for an individual charge element
+ *
+ * @param {String} billRunId - UUID of the bill run
+ * @param {String} licenceId - UUID of the individual licence to review
+ * @param {String} reviewChargeElementId - The UUID of the review charge element being viewed
+ *
+ * @returns {Promise<Object[]>} Contains an array of bill run data and review charge element data
```
Cruikshanks, line 18: Correction. We're not returning an array and neither value is an array.

```suggestion
 * @returns {Promise<Object>} An object containing the bill run and review charge element instances
```

app/views/bill-runs/match-details.njk
```diff
@@ -0,0 +1,208 @@
...
+      {% else %}
+        {% set colour = "govuk-tag--green" %}
+      {% endif %}
+
+      <p class="govuk-body">
+        {{ govukTag({
+          text: chargeElement.status,
+          classes: colour
+        }) }}
+      </p>
```
Cruikshanks, line 35: Could we use our `badge.njk` macro to handle this instead?

app/views/bill-runs/match-details.njk
```diff
@@ -0,0 +1,208 @@
...
+  </div>
+
+  {# ---------- Billable returns button ----------- #}
+  <div class="govuk-grid-row govuk-!-margin-bottom-6">
+    <div class="govuk-grid-column-full">
+      <button class="govuk-button govuk-button--secondary">
+        Edit the billable returns
+      </button>
+    </div>
+  </div>
```
Cruikshanks, line 109: Just a question. Is it not possible to use the [GOV.UK button component](https://design-system.service.gov.uk/components/button/) for this?

app/views/bill-runs/match-details.njk
```diff
@@ -0,0 +1,208 @@
...
+      {% if return.returnStatus == 'overdue' or return.returnStatus == 'query' %}
+        {% set colour = "govuk-tag--red"%}
+        {% set returnLink = "/return/internal?returnId=" + return.returnId %}
+      {% elif return.status == 'void' %}
+        {% set colour = "govuk-tag--grey" %}
+        {% set returnLink = "/return/internal?returnId=" + return.returnId %}
+      {% else %}
+        {% set colour = "govuk-tag--green" %}
+        {% set returnLink = "/returns/return?id=" + return.returnId %}
+      {% endif %}
```
Cruikshanks, line 126: We are already working out several other details for each matched return in the presenter. Could we not also determine the `returnLink` in the presenter?

And on determining the color, could that not be dealt with by just using the `badge.njk` macro?

Cruikshanks (approved):

PR #876, opened 2024-04-02
Select the returns cycle for the requirements for returns

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4297

This is PR is focused on adding functionality to page 4 of 11 in the return requirement pages.

The pages were previously setup as part of a separate PR to create the user journey.


test/services/return-requirements/submit-returns-cycle.service.test.js
```diff
@@ -0,0 +1,89 @@
...
+          startDate: '2022-04-01T00:00:00.000Z'
+        },
+        journey: 'returns-required'
+      }
+    })
+  })
+
+  describe('when called', () => {
+    describe('with a valid payload', () => {
+      const payload = {
```
Beckyrose200, line 40: Since you are using payload twice I would put payload as a let and then put them both in beforeEach blocks instead of const.

test/validators/no-returns-required/returns-cycle.validator.test.js
```diff
@@ -0,0 +1,47 @@
...
+
+const { describe, it } = exports.lab = Lab.script()
+const { expect } = Code
+
+// Thing under test
+const ReturnsCycleValidator = require('../../../app/validators/return-requirements/returns-cycle.validator.js')
+
+describe('Returns Cycle validator', () => {
+  describe('when valid data is provided', () => {
+    const payload = {
```
Beckyrose200, line 15: See comment above

Beckyrose200 (changes requested):
Looks really good just a few comments :)

app/views/return-requirements/returns-cycle.njk
```diff
@@ -15,12 +17,42 @@
...
 
 {% block content %}
+
+{# Error summary #}
+  {% if error %}
+    {{ govukErrorSummary({
+      titleText: "There is a problem",
+      errorList: [
+        {
+          text: "Select the returns cycle for the return requirement",
```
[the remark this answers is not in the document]
rvsiyad: Good spot! I was hard coding the value but now figured out I can get the error message by using this:
```suggestion
          text:  error.text,
```

app/validators/return-requirements/returns-cycle.validator.js
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * When setting up a requirement users must specify a returns cycle for the return requirement. Users must select one
+ * period for the returns cycle. If this requirement is not met the validation will return an error.
+ *
+ * @param {Object} payload - The payload from the request to be validated
+ *
+ * @returns {Object} The result from calling Joi's schema.validate(). The result from calling Joi's schema.validate().
+ * If any errors are found the `error:` property will also exist detailing what the issue is.
+ */
+
```
Beckyrose200, line 21: The space needs to be removed here

test/presenters/return-requirements/returns-cycle.presenter.test.js
```diff
@@ -0,0 +1,76 @@
...
+const Code = require('@hapi/code')
+
+const { describe, it, beforeEach } = exports.lab = Lab.script()
+const { expect } = Code
+
+// Thing under test
+const ReturnsCyclePresenter = require('../../../app/presenters/return-requirements/returns-cycle.presenter.js')
+
+describe('Returns Cycle presenter', () => {
+  let session
```
Beckyrose200, line 14: The let session block can go down a describe block. It's good practise to only declare the variables when you need them. Since your next describe block is 'when provided with a populated session' it would be good to declare the session there (since it is within the context).

Beckyrose200 (changes requested):
Just two more really small comments from me, but really good! Get those two done and I will approve :D

Beckyrose200 (approved):

PR #878, opened 2024-04-02
Edit billable volume on element review page

Beckyrose200:

As part of our two-part tariff bill run work, we've developed licence review pages where you can view the matching details of the two-part tariff engine. When viewing a charge elements matching details you have the option as a user to be able to change the billable volume on the element. This PR covers that functionality.


Beckyrose200:
SonarCloud is currently complaining that there is duplicate code coverage within the amend-billable-returns-presenter. This is valid as there is a lot of duplicated code in the presenters but can be ignored for the time being. The plan is once all the pages have been built, to then go back into all the presenters and services and see what can be de-duplicated.

test/services/bill-runs/two-part-tariff/submit-amended-billable-returns.service.test.js
```diff
@@ -0,0 +1,69 @@
...
+
+// Test helpers
+const DatabaseSupport = require('../../../support/database.js')
+const ReviewChargeElementHelper = require('../../../support/helpers/review-charge-element.helper.js')
+const ReviewChargeElementModel = require('../../../../app/models/review-charge-element.model.js')
+
+// Thing under test
+const SubmitAmendedBillableReturnsService = require('../../../../app/services/bill-runs/two-part-tariff/submit-amended-billable-returns.service.js')
+
+describe('Submit Amended Billable Returns Service', () => {
```
Jozzey, line 18: Need to add a test to check when the validation fails and there is a value in `validationResult`

Jozzey (changes requested):
Just a couple of comments

Jozzey (approved):

PR #879, opened 2024-04-02
Fix bump supp. end year if no annual bill run

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4403

In [Bump supplementary end year if no annual bill run](https://github.com/DEFRA/water-abstraction-system/pull/875) we added logic that would mean when creating a new supplementary bill run the engine could tell if an annual bill run existed for the same year and region.

If one doesn't it will make a mess of the billing for customers if the supplementary happened before the annual. But we don't want to block users from working. So, instead, we bump the financial end year to use back to the same as the last 'sent' annual bill run for that region.

We even [Refactor DetermineBillingPeriods to use type](https://github.com/DEFRA/water-abstraction-system/pull/864) so it would know to use the calculated year when determining the billing periods for the supplementary bill run.

What we forgot to do was remove a [temporary fix](https://github.com/DEFRA/water-abstraction-system/pull/859) we'd added until this work can be completed.

This means the year we've calculated is being dropped and not passed to `DetermineBillingPeriods`. This change fixes that!

### DIFF
--- a/app/services/bill-runs/setup/create.service.js
+++ b/app/services/bill-runs/setup/create.service.js
@@ -50,15 +50,7 @@ async function _triggerBillRun (regionId, batchType, user, year, existingBillRun
     return null
   }
 
-  // TODO: We do want to send the year through. However, the billing engine was written initially for supplementary
-  // then extended for annual and soon to be amended again for two-part tariff. 2PT has a short term requirement to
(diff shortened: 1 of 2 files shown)


Cruikshanks:
I knew I'd already done this! The change is in #864 but I'd overlooked that this hadn't been merged before moving the ticket to 'done'!

PR #882, opened 2024-04-03
Update badge macro to support 2pt

Beckyrose200:

As part of the ongoing work building the two-part tariff bill run pages, this PR is to adapt the badge macro to support the two-part tariff review statuses and update the pages that use badges.


Cruikshanks (commented):
I am _so_ sorry! 😢 

This change made me realise it's not just the colour that needs to change but, in some cases, the size of the tag.

In essence, the needs of the review pages are different to those of the bill run pages.

So, I believe a better approach is to have separate macros for each context rather than the workarounds you've been forced to implement.

If I'd have known better I could have given better direction.

So, I'm sorry for asking you to do this when I can see it is not the right direction. But so you're not still blocked I've cracked on and raised a PR for the changes I think we believe need to make.

https://github.com/DEFRA/water-abstraction-system/pull/885

Check it out and ping me if you need to discuss. But I think we might just need to close this PR.

PR #883, opened 2024-04-03
Add a remove duplicate licence feature

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4429

Usually once or twice a month we have to deal with a duplicate licence in the service. It is not a duplicate but an issue with NALD (the upstream service) that our [water-abstraction-import](https://github.com/DEFRA/water-abstraction-import) does not handle.

We believe when a change is made to a licence in NALD it requires users to enter the licence reference again. This creates a new record in NALD but our import service recognises it is for the same licence because the references match.

However, NALD is known for poor validation and doesn't trim data entered. So, often a user will accidentally add an extra space or newline character when entering the licence reference again. On import, computers being the way they are, will see this as a new licence so will create a new licence record with the same 'bad' reference.

This causes an error for anything that attempts to retrieve the licence by reference in the service. The most visible casualty is the search page.

We _should_ be fixing the import. But we're only just diving into that black box of code and are reticent to change things we don't understand. Plus the reason we are now looking at the import code is because of the strong likelihood we will have to build a new import service this year.

So, to avoid constantly having to create migration scripts and allow us to respond sooner when an issue is spotted, this change adds a tool that will allow us to delete any 'bad' licence records.


app/services/data/deduplicate/submit-deduplicate.service.js
```diff
@@ -0,0 +1,47 @@
...
+ * controller will redirect the user to the search page and have it search for the licence reference. If the tool has
+ * done its job the page should no longer error.
+ *
+ * @returns {Promise<Object>} an object containing a parsed version of the licence reference submitted else an error
+ * message if nothing was entered
+ */
+async function go (payload) {
+  const licenceRef = payload['licence-ref']
+
+  if (!licenceRef || licenceRef.trim() === '') {
```
robertparkinson, line 29: what if someone puts a space in the middle of the reference? will the `whereILike` handle that ok?

robertparkinson (approved):

Beckyrose200 (approved):
This is so much better than running those migration scripts!

app/services/data/deduplicate/submit-deduplicate.service.js
Cruikshanks: Pfft! Don't need your approval anymore! 😝

PR #885, opened 2024-04-03
Add review badge, rename existing and call it tag!

Cruikshanks:

A long time back we created a macro for displaying 'badges' for bill runs. These are [GOV.UK tag components](https://design-system.service.gov.uk/components/tag/) and indicate the status of a bill run.

We have to display them in so many places it is the only thing to justify having a macro so far.

In the work we are doing building the review two-part tariff matches and allocations views we again need badges for statuses.

Great! We have a 'badge' macro. Only there are differences. `ready` for bill runs should be displayed as blue. In the review screens it needs to be green. Also, the badges need to be different sizes for the two contexts.

We had it in mind to [expand the existing functionality](https://github.com/DEFRA/water-abstraction-system/pull/882) but that was what made it clear we needed two different badges for the 2 contexts.

This means the existing `badge.njk` macro should also be given a more explicit name.

Then the final change; stop calling these blooming things badges. They're tags!

### DIFF
--- a/app/views/bill-licences/remove.njk
+++ b/app/views/bill-licences/remove.njk
@@ -5,7 +5,7 @@
 {% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
 {% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
 
-{% from "macros/badge.njk" import statusBadge %}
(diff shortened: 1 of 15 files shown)


Beckyrose200 (approved):
This is much nicer and looks cleaner than what I had going on :)

PR #890, opened 2024-04-04
Move two-part tariff review routes to own folder

Cruikshanks:

Before a two-part tariff bill run is ready to be billed users first have to review the results of the matching and allocation. Because the review is connected to the bill run process we started by having these routes be part of `/bill-runs`. But this means we are adding more and more services into `app/services/bill-runs/two-part-tariff` that are solely to review the results. It is making it harder to see what is billing engine-related and what is just for the review views.

So, before this gets any more confusing we are going to move all two-part tariff review-related routes to a new `/review` root URL. We can then move all services, presenters etc to their own matching `/review` folder. This will then separate our review logic from the billing logic.


Cruikshanks:
Going to close this for now. We still fully intend to do this but taking a step back, there is still a lot of changes going on in this area and I think it would be prudent to wait for all the review pages to be complete before then looking at what we have and how it can be tidied up/refactored.

PR #900, opened 2024-04-11
Two-part tariff logging branch

Beckyrose200:
The purpose of this PR is to allow us to add additional logging information on our two-part tariff engine. This is so we can see why the engine is making certain decisions in regards to the matching and allocating. This change also means we can test individual licences rather then an entire region.


Cruikshanks:
No longer needed

PR #902, opened 2024-04-11
Fix review licence page

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4219

Following on from testing the above ticket, an issue with the page was discovered where a piece of text is missing at the top of the page. The text is dynamically shown when there are charge elements that have a status of 'review'. This PR is to add this text.

### DIFF
--- a/test/presenters/bill-runs/two-part-tariff/review-licence.presenter.test.js
+++ b/test/presenters/bill-runs/two-part-tariff/review-licence.presenter.test.js
@@ -184,34 +184,6 @@ describe('Review Licence presenter', () => {
         })
       })
     })
-
-    describe('when there is data to be presented and the user has clicked the "Mark progress" button', () => {
-      beforeEach(() => {
-        billRun = _billRun()
-        licence = _licenceData()
-        markProgress = 'mark'
-      })
-
-      it('correctly returns the text for the "Licence updated" notification banner', async () => {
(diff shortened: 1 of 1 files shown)


app/views/bill-runs/review-licence.njk
```diff
@@ -41,6 +42,13 @@
       </a></li>
   </ul>
 
+  {# ------------- Licence in review text --------------- #}
```
Jozzey, line 45: We shouldn't be adding all the dashes to the comments
```suggestion
  {# Licence in review text #}
```

app/presenters/bill-runs/two-part-tariff/review-licence.presenter.js
```diff
@@ -127,6 +128,18 @@ function _contactName (billingAccount) {
   return null
 }
 
+function _elementsInReview (licence) {
+  const hasReviewStatus = licence.reviewChargeVersions.some((chargeVersion) => {
```
Jozzey, line 132: Had a chat with Alan about this. We think it should be possible to calculate if an element is in review and return a true/false at the licence level by updating the SQL query. I can pair with you on it if you like when you get back.

Jozzey (changes requested):
Couple more changes following going through this with Alan

Jozzey (approved):

PR #906, opened 2024-04-12
Returns required journey - Notes page

Demwunz:
https://eaflood.atlassian.net/browse/WATER-4269

This branch will add a notes feature to add a note to the return requirement.

The page will have a text entry box for notes and a character count below it to display how many characters have been entered. Validation will ensure no more than 500 characters are entered.

The content will be saved in the session until the user submits the application.

<details>
<summary>Screenshots</summary>
Add Note page:

image

Too many characters inline GOVUK error:
image

Submitted with no text error:
image

Submitted with too many characters:
image

Check Your Answers page:
image
</details>


Cruikshanks (changes requested):
Accepting that it is highly unlikely to be a different user, I think it would still be more 'correct' if the user email for the note was captured _with_ the note.

So, I think it would be better to grab the user off of the request in `submitAddNote (request, h) {}` and pass it through to `SubmitAddNoteService`. I'd then save the email with the note value, for example

```javascript
session.note = {
  content: payload.note,
  userEmail: user.username
}
```

The `CheckYouAnswersPresenter` called in `CheckYourAnswersService` can then extract this information from the session rather than use the email of whoever made the request to `/check-your-answers`.

Would you have any issues with this?

Demwunz:
@Cruikshanks makes sense, I have no issue with this suggestion

app/presenters/return-requirements/add-note.presenter.js
```diff
@@ -0,0 +1,34 @@
...
+  if (!currentNote) {
+    return {
+      note: ''
+    }
+  }
+
+  return {
+    note: sessionData.note
+  }
+}
```
Cruikshanks, line 30: Surely this won't work any more because of the changes made to how the note is stored in the session. It's now `{ note: { content: 'foo bar', userEmail: 'jo@example.com' }` ??

Plus, I think we can make it a bit simpler.

```suggestion
function go (session) {
  const { id, data } = session

  const pageData = {
    id,
    licenceRef: data.licence.licenceRef,
    note: data.note ? data.note.content : ''
  }

  return pageData
}
```

test/presenters/return-requirements/add-note.presenter.test.js
```diff
@@ -0,0 +1,41 @@
...
+  beforeEach(() => {
+    session = {
+      id: 'f1288f6c-8503-4dc1-b114-75c408a14bd0',
+      data: {
+        licence: {
+          id: 'ea53bfc6-740d-46c5-9558-fc8cabfc6c1f',
+          licenceRef: '01/123',
+          licenceHolder: 'Jane Doe'
+        },
+        note: 'Note attached to return requirement'
```
Cruikshanks, line 25: With the changes made in `SubmitAddNoteService` this is no longer how the note will be stored on the session.

The test and the presenter need to be updated to expect

```suggestion
        note: {
          content:  'Note attached to return requirement',
          userEmail: 'jo@example.com'
        }
```

test/presenters/return-requirements/add-note.presenter.test.js
Cruikshanks, line 1: I think we should also have some coverage for when `session.data.note` is undefined. Typically where populating a property involves logic we'll have the happy path then call out those specific properties. Check out `describe("the 'accountName' property"` and others in `test/presenters/bills/view-bill.presenter.test.js` for an example of the pattern.

Cruikshanks (changes requested):
Still some corrections that need to be made for this to work I'm afraid.

app/presenters/return-requirements/add-note.presenter.js
Demwunz: This works because we return the object `{ content: 'foo bar', userEmail: 'jo@example.com' }` back to nunjucks template. We access it [here](https://github.com/DEFRA/water-abstraction-system/blob/f400bf60469e03b2df0d7b38833eac9f7cfeba41/app/views/return-requirements/check-your-answers.njk#L108)  using `note.content` and `note.userEmail` respectively. If you prefer `content` and `userEmail` to be sent as separate variables, we can do that instead.

I agree, using ternary operator to simplify the line makes does make it more succinct.

Demwunz: amended the test and presenter. Rather than ternary, we could simplify line 14 to `data.note.content || ''` which yields the same result. I've added it to the PR, happy to use ternary if you object.

app/presenters/return-requirements/add-note.presenter.js
```diff
@@ -0,0 +1,22 @@
...
+ * @module AddNotePresenter
+ */
+
+function go (session) {
+  const { id, data } = session
+
+  const pageData = {
+    id,
+    licenceRef: session.data.licence.licenceRef,
+    note: data.note.content || ''
```
Cruikshanks, line 14: 🍬 😮

app/views/return-requirements/add-note.njk
```diff
@@ -1,26 +1,55 @@
 {% extends 'layout.njk' %}
-{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from "govuk/components/character-count/macro.njk" import govukCharacterCount %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from 'govuk/components/error-summary/macro.njk' import govukErrorSummary %}
```
Cruikshanks, line 5: Just some alphabetty-spaghetti!

```suggestion
{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
{% from "govuk/components/button/macro.njk" import govukButton %}
{% from "govuk/components/character-count/macro.njk" import govukCharacterCount %}
{% from 'govuk/components/error-summary/macro.njk' import govukErrorSummary %}
```

test/presenters/return-requirements/add-note.presenter.test.js
Cruikshanks: I know we've simplified it further in the presenter. But I still think even just for documentation purposes it'd be good to call out the `note:` property with its own tests.

test/services/return-requirements/submit-add-note.service.test.js
```diff
@@ -0,0 +1,149 @@
...
+        licence: {
+          id: '8b7f78ba-f3ad-4cb6-a058-78abc4d1383d',
+          currentVersionStartDate: '2023-01-01T00:00:00.000Z',
+          endDate: null,
+          licenceRef: '01/ABC',
+          licenceHolder: 'Turbo Kid',
+          startDate: '2022-04-01T00:00:00.000Z'
+        },
+        journey: 'no-returns-required',
+        note: '',
```
Cruikshanks, line 37: I'm fine with either the property not existing or an existing note being on the session as part of the test. But it'd never be `note: ''` anymore so we need to correct it.

app/presenters/return-requirements/add-note.presenter.js
Demwunz: This would normally work, but it was causing nunjucks to have a wobbly, so I reverted to ternary :(

Cruikshanks (changes requested):
I think there are still a few NITS to be ironed out with this. Plus I was never sure about the error messages. Having checked [the design system](https://design-system.service.gov.uk/components/textarea/) they _really_ do not seem consistent with the guidance (though that is not a problem of our making).

But as this one has done a fair amount of ping-pong I reckon it is time to work together on this to get it over the line rather than highlight some more individual suggested changes.

Cruikshanks (approved):

PR #911, opened 2024-04-15
Adding CSRF form protection via Hapi Crumb plugin

rvsiyad:
https://github.com/DEFRA/water-abstraction-team/issues/114

[Crumb](https://hapi.dev/module/crumb/) is used to diminish CSRF attacks using a random unique token that is validated on the server side. **Crumb** may be used whenever you want to prevent malicious code from executing system commands that are performed by HTTP requests.

After this change, any view that has a form that uses `method="POST"`, will need to have a hidden input field.

```html
  <input type="hidden" name="wrlsCrumb" value="{{wrlsCrumb}}"/>
```

When the page is requested, the **Crumb** plugin will generate a token in the `onPreResponse` event that it will save to a cookie called 'wrlsCrumb'. It also makes it available in the view context, hence our views can reference `{{wrlsCrumb}}`.

When the user submits the form, the **Crumb** plugin jumps in again, this time on the `onPostAuth` event. It compares both values, and if they match, it 'authorises' the request.

If the payload is missing `{{wrlsCrumb}}`, or the value doesn't match that saved in the cookie (which is secure and unreadable to the client) then it rejects the request. In our service, the user will see a 404, as that is the default for an 'unauthorised' request.

**Crumb** will be enabled by default for all endpoints to avoid us forgetting to protect an HTML form. However, this means it performs the check for _all_ POST requests. Our service exposes API-only POST endpoints, which do not expect a payload. For this, the route config must include the following to disable **Crumb**.

```javascript
 options: {
   plugins: {
     crumb: false
   }
 }
```


Cruikshanks:
Just a reminder to my future self. Where we have come to a stop is unit testing. What solution can we implement that will allow us to test a POST request to an endpoint that has CSRF (crumb) enabled on it?

The POST request needs to be able to include a token that crumb will validate successfully. Digging round the future farming repos they solved this by sending a GET request first then extracting the token and adding it to the POST.

There is not much on the interwebs on crumb and testing. What we found suggests disabling crumb when unit testing, though no actual example on how to go about doing this.

There is also the feeling that as we know all the config and credentials used that we _should_ be able to send a POST request with a token that crumb will validate, without having to send a GET request first or disabling it.

robertparkinson (approved):

app/plugins/crumb.plugin.js
```diff
@@ -0,0 +1,50 @@
...
+ * `{{wrlsCrumb}}`.
+ *
+ * When the user submits the form the Crumb plugin jumps in again, this time on the `onPostAuth` event. It compares both
+ * values and if they match it 'authorises' the request.
+ *
+ * If the the payload is missing `{{wrlsCrumb}}`, or the value doesn't match that saved in the cookie (which is secure
+ * so unreadable to the client) than it rejects the request. In our service the user will see a 404 as that is the
+ * default for an 'unauthorised' request.
+ *
+ * We have Crumb enable by default for all endpoints to avoid us forgetting to protect a html form. However, this means
```
Jozzey, line 31: Just a typo
```suggestion
 * We have Crumb enabled by default for all endpoints to avoid us forgetting to protect a html form. However, this means
```

Jozzey (dismissed):
Looks good. Spotted a typo

Jozzey (approved):

PR #913, opened 2024-04-16
create returns from abstraction data

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4442

This story covers the creation of return requirements based from the abstraction data that is in the licence


app/lib/general.lib.js
```diff
@@ -98,6 +98,41 @@ function determineCurrentFinancialYear () {
...
 }
 
+/**
+ * Generate a string that represents an abstraction point
+ *
+ * When abstracting water the point at which this is done can be described in several ways depending on the number of
+ * Nation Grid References are saved against the abstraction point. This function checks for these references and builds
+ * a string that defines the details of the abstraction point.
+ *
+ * @returns {String} a description of the abstraction point
```
Cruikshanks, line 108: Just missing the param

```suggestion
 * a string that defines the details of the abstraction point.
 *
 * @param {Object} pointDetail - Object containing all the details for the point 
 *
 * @returns {String} a description of the abstraction point
```

app/services/return-requirements/check.service.js
```diff
@@ -19,6 +20,10 @@ const SessionModel = require('../../models/session.model.js')
 async function go (sessionId, yar) {
   const session = await SessionModel.query().findById(sessionId)
 
+  // if (session.setup === 'use-abstraction-data') {
+  //   const result = await FetchReturnRequirementsBasedOnAbstractionDataService.go(session.licence.id)
+  // }
+
```
Cruikshanks, line 26: If we don't need it we should delete it. Assuming this just got overlooked!

```suggestion
```

app/services/return-requirements/check.service.js
```diff
@@ -6,6 +6,7 @@
  */
 
 const CheckPresenter = require('../../presenters/return-requirements/check.presenter.js')
+// const FetchReturnRequirementsBasedOnAbstractionDataService = require('./fetch-requirements-for-returns-from-abstraction-data.service.js')
```
Cruikshanks, line 9:
```suggestion
```

app/views/return-requirements/check-your-answers.njk
Cruikshanks, line 1: If you changed something in this file as part of the PR it will explain why it is trying to add it. However, this was renamed as `check.njk` and updated in my big tidy up PR.

If you haven't changed anything in the file you can remove it. Else whatever change you made needs to be made to `check.njk`.

app/services/return-requirements/fetch-requirements-for-returns-from-abstraction-data.service.js
```diff
@@ -0,0 +1,103 @@
...
+  const purposeStartDate = new Date(epocStartDate, purpose.abstractionPeriodStartMonth - 1, purpose.abstractionPeriodStartDay)
+  const purposeEndDate = new Date(epocStartDate, purpose.abstractionPeriodEndMonth - 1, purpose.abstractionPeriodEndDay)
+
+  const summerStartDate = new Date('1970-04-01')
+  const summerEndDate = new Date('1970-10-31')
+
+  return purposeStartDate >= summerStartDate && purposeEndDate <= summerEndDate
+}
+
+function _calculateReportingCycle (licenceData, licenceVersionPurpose) {
```
Cruikshanks, line 89: I think the name should be changed to fit in with the step in the journey and the collection function.

```suggestion
function _calculateReportingFrequency (licenceData, licenceVersionPurpose) {
```

app/services/return-requirements/fetch-requirements-for-returns-from-abstraction-data.service.js
```diff
@@ -0,0 +1,103 @@
...
+  if (licenceData.waterUndertaker) {
+    return 'daily'
+  }
+
+  if (licenceVersionPurpose.dailyQuantity <= upperDailyQuantityThreshold) {
+    return 'monthly'
+  } else {
+    return 'weekly'
+  }
```
Cruikshanks, line 98: I encourage folks to use a [guard clause approach](https://en.wikipedia.org/wiki/Guard_(computer_science)) to `if` statements like this. IMHO they read clearer and ensure you always have a distinct return statement.

```suggestion
  if (licenceVersionPurpose.dailyQuantity <= upperDailyQuantityThreshold) {
    return 'monthly'
  }
  
  return 'weekly'
```

app/services/return-requirements/fetch-requirements-for-returns-from-abstraction-data.service.js
```diff
@@ -0,0 +1,103 @@
...
+    return 'daily'
+  }
+
+  if (licenceVersionPurpose.dailyQuantity < minimumDailyQuantityThreshold) {
+    return 'none'
+  } else if (licenceVersionPurpose.dailyQuantity <= upperDailyQuantityThreshold) {
+    return 'monthly'
+  } else {
+    return 'weekly'
+  }
```
Cruikshanks, line 75: Would also suggest a guard clause style approach here as well.

```suggestion
  if (licenceVersionPurpose.dailyQuantity < minimumDailyQuantityThreshold) {
    return 'none'
  }
  
  if (licenceVersionPurpose.dailyQuantity <= upperDailyQuantityThreshold) {
    return 'monthly'
  }
  
  return 'weekly'
```

Cruikshanks (changes requested):
I'm not finished! But my machine is forcing a restart. So, just to avoid loosing the suggestions I've already made completing this review for now. 😬

app/services/return-requirements/fetch-requirements-for-returns-from-abstraction-data.service.js
Cruikshanks, line 1: I love verbose names! But I reckon in this case we can simplify things a bit. How about `fetch-requirements-from-abstraction-data.service.js`?

app/services/licences/fetch-licence-summary.service.js
Cruikshanks, line 1: Do we need _everything_ this fetch service provides plus these additional 3 fields? Or is it the case this service provides some stuff we need apart from these 3 fields?

If it's the latter we will normally create a specific fetch service for the use case. We'll then look at what if anything can be refactored as shareable/reusable afterwards.

This might also answer my thoughts on whether we could simplify `fetch-requirements-for-returns-from-abstraction-data.service.js` by moving some of the work into the query.

Cruikshanks (changes requested):
A couple more points. But also a question about how the data is being fetched that would be worth answering before I go deeper.

I could also really do with the PR title and description being firmed up. What I read makes it sound like we are _creating_ the requirement you'll see on the `/check` page.

I could be wrong, but what I'm seeing is the things needed to fetch and format the data _in preparation_ of creating the session requirement.

Ping me if you want a chat or clearer guidance!

app/services/licences/fetch-licence-summary.service.js
robertparkinson: It would need the `permitLicence`, `licenceRef`, `licenceVersions`,  which are the big bits that are used in the licence summary page. I've added `waterUndertaker` which is not used in the licence summary page. 

I can create a separate fetch service for it. I just thought reusing the existing service would cut down on code duplication

Cruikshanks: I'm fine with a bit of duplication. I find that a clearer 'smell' of when something needs refactoring rather than preemptively doing it.

For example, I don't think we're using licence holder or the gauging stations from `FetchLicenceSummaryService`. So, having a dedicated service makes it clearer what we actually need.

Cruikshanks:
Superseded by https://github.com/DEFRA/water-abstraction-system/pull/1107

PR #916, opened 2024-04-17
https://eaflood.atlassian.net/browse/WATER-4268

Demwunz:
This page contains controls to allow the user to review and links to amend previous answers if necessary.

This page will have no validation.

The page is shared with two return-requirements journeys.

Some links need to be adjusted to redirect the user back to the appropriate page for the journey.


Cruikshanks (approved):
Happy to approve this but would you mind having another stab at the PR description, and also correct the title?

The current description makes it sound like we are adding a whole new page whereas we're just making the 'Change reason' link work for both returns required and no returns required journeys.

The corrected title and description can then be used for the commit message when squashing.

Demwunz:
@Cruikshanks updated title and description, I should have caught this.

PR #917, opened 2024-04-17
Fix 2PT volume allocation

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4377

It was found during testing that if the individual return line volumes are greater than the charge reference authorised volume, which in turn is greater than the element authorised volume. That the element was being over-allocated.

This PR fixes this issue and creates a new unit test to check for it as it wasn't being picked up previously.

### DIFF
--- a/test/services/bill-runs/two-part-tariff/allocate-returns-to-charge-element.service.test.js
+++ b/test/services/bill-runs/two-part-tariff/allocate-returns-to-charge-element.service.test.js
@@ -10,7 +10,7 @@ const { expect } = Code
 // Thing under test
 const AllocateReturnsToChargeElementService = require('../../../../app/services/bill-runs/two-part-tariff/allocate-returns-to-charge-element.service.js')
 
-describe('Allocate Returns to Charge Element Service', () => {
+describe.only('Allocate Returns to Charge Element Service', () => {
   describe('when there are records to process', () => {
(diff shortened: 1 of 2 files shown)


test/services/bill-runs/two-part-tariff/allocate-returns-to-charge-element.service.test.js
```diff
@@ -144,6 +144,53 @@ describe('Allocate Returns to Charge Element Service', () => {
...
+          const { chargeElement, chargeReference, matchingReturns } = testData
+
+          AllocateReturnsToChargeElementService.go(chargeElement, matchingReturns, chargePeriod, chargeReference)
+
+          expect(chargeElement.authorisedAnnualQuantity).to.equal(3)
+          expect(chargeElement.allocatedQuantity).to.equal(3)
+          expect(chargeElement.returnLogs[0].allocatedQuantity).to.equal(3)
+        })
+
+        it('allocates 10 from the return log', () => {
```
Beckyrose200, line 177: Think this is meant to be 'allocates 3 from the return log' not 10? :)

Jozzey: You are correct, I'll changer that

Beckyrose200 (approved):

PR #920, opened 2024-04-18
Fix fetching returns for two part tariff

Beckyrose200:

As part of the ongoing testing for the two-part tariff engine, it was found that there were too many returns matching a charge element. The returns were spread over two years (all within the financial year but starting and finishing outside). This wasn't a bug in the code but rather an A|C that wasn't fleshed out properly. The team don't want returns to be pulled in if the end date of the return is outside the financial year. This PR is to update the fetching query to adjust the start and end dates that the returns can be picked from.

### DIFF
--- a/test/services/bill-runs/two-part-tariff/fetch-return-logs-for-licence.service.test.js
+++ b/test/services/bill-runs/two-part-tariff/fetch-return-logs-for-licence.service.test.js
@@ -147,7 +147,7 @@ describe('Fetch Return Logs for Licence service', () => {
       })
     })
 
-    describe('where the return logs start date is outside the billing period', () => {
+    describe('where the return logs start date is outside the billing period but the end date is inside', () => {
       beforeEach(async () => {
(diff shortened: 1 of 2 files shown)


test/services/bill-runs/two-part-tariff/fetch-return-logs-for-licence.service.test.js
```diff
@@ -146,10 +146,31 @@ describe('Fetch Return Logs for Licence service', () => {
         expect(result[0].returnSubmissions[0].returnSubmissionLines).to.have.length(0)
       })
     })
+
+    describe('where the return logs start date is outside the billing period', () => {
```
Jozzey, line 150: Would be good to make it clearer that the end date is within the billing period as well as the start date being outside.

Jozzey (changes requested):
Just a couple of comments

Jozzey (approved):

PR #923, opened 2024-04-18
Check your answers page with notes for additions, changes and deletions

Demwunz:
https://eaflood.atlassian.net/browse/WATER-4291

This feature allows the user to see notifications on the page when a note is added, changed or deleted.

A `delete-note` service has been added which is accessed via a simple GET request. This will delete any note in the session attached to the requirement.

Notifications are shown using hapi/yar and are immediately removed upon first view on the page.

<details>
<summary>Screenshots</summary>
image
image
image

</details>


test/services/return-requirements/delete-note.service.test.js
```diff
@@ -0,0 +1,64 @@
+// write test in hapi/yar for delete-note.service.js
```
Cruikshanks, line 1: I'm guessing this was a note to yourself!

So, we need to drop it. Also, we always add `'use strict'` to the top of all our files.

```suggestion
'use strict'

```

test/services/return-requirements/delete-note.service.test.js
```diff
@@ -0,0 +1,64 @@
...
+// Test framework dependencies
+const Lab = require('@hapi/lab')
+const Code = require('@hapi/code')
+const Sinon = require('sinon')
+const { describe, it, beforeEach } = exports.lab = Lab.script()
```
Cruikshanks, line 6: NIT - adding some whitespace

```suggestion
const Sinon = require('sinon')

const { describe, it, beforeEach } = exports.lab = Lab.script()
```

test/services/return-requirements/delete-note.service.test.js
```diff
@@ -0,0 +1,64 @@
...
+    note: {
+      content: 'Note attached to requirement',
+      userEmail: 'carol.shaw@atari.com'
+    },
+    startDateOptions: 'licenceStartDate'
+  }
+}
+describe('Delete Note service', () => {
+  let session
+  let yarStub
```
Cruikshanks, line 37: Any values we need for the unit tests we normally define within the `describe()` block at the relevant scope. In this case I'm assuming that would be

```suggestion
describe('Delete Note service', () => {
  const sessionData = {
    data: {
      id: 'f1288f6c-8503-4dc1-b114-75c408a14bd0',
      checkYourAnswersVisited: true,
      licence: {
        endDate: null,
        licenceRef: '01/ABC',
        licenceHolder: 'Astro Boy',
        currentVersionStartDate: '2023-02-08T00:00:00.000Z'
      },
      reason: 'abstraction-below-100-cubic-metres-per-day',
      journey: 'no-returns-required',
      note: {
        content: 'Note attached to requirement',
        userEmail: 'carol.shaw@atari.com'
      },
      startDateOptions: 'licenceStartDate'
    }
  }

  let session
  let yarStub
```

test/services/return-requirements/check-your-answers.service.test.js
Cruikshanks, line 1: Wherever we use Sinon we need to add a `restore()` at the tope level in an `afterEach()` block.

```javascript
  afterEach(() => {
    Sinon.restore()
  })
```

test/services/return-requirements/delete-note.service.test.js
```diff
@@ -0,0 +1,64 @@
...
+    })
+
+    yarStub = {
+      flash: Sinon.stub()
+    }
+  })
+
+  it('deletes the note', async () => {
+    await DeleteNoteService.go(session.id, yarStub)
+    expect(session.note).to.be.undefined()
```
Cruikshanks, line 53: First, a NIT, adding some whitespace. But also this test is incorrect.

It is currently passing because we never set the property `session.note` but _do_ set the property `session.data.note`. If you update the test to check for that it will fail because `session.data.note` is defined.

You need to do this to fix the test and properly check the note has been deleted.

```suggestion
    await DeleteNoteService.go(session.id, yarStub)
    
    const refreshedSession = await session.$query()
    
    expect(refreshedSession.data.note).to.be.undefined()
```

However, this still won't pass unless the `DeleteNoteService` is updated to drop the `note` property instead of setting it to an empty string.

app/services/return-requirements/delete-note.service.js
```diff
@@ -0,0 +1,45 @@
...
+
+  yar.flash('notification', notification)
+
+  return notification
+}
+
+async function _save (session) {
+  const currentData = session.data
+
+  currentData.note = ''
```
Cruikshanks, line 38: If we want to delete the note we ideally want to put it back to the state it was before we added one in the first place. If that is the case I think you need to call this instead.

```suggestion
  delete currentData.note
```

test/services/return-requirements/submit-add-note.service.test.js
Cruikshanks, line 1: Wherever we use Sinon we need to add a `restore()` at the tope level in an `afterEach()` block.

```javascript
  afterEach(() => {
    Sinon.restore()
  })
```

Cruikshanks (changes requested):
A few changes needed and then this will be ship-shape and bristol fashion!

test/services/return-requirements/delete-note.service.test.js
Demwunz: makes sense, updated

test/services/return-requirements/submit-add-note.service.test.js
```diff
@@ -37,34 +39,75 @@ describe('Submit Add Note service', () => {
...
+        it("sets the notification message to 'Added' for a new note", async () => {
+          await SubmitAddNoteService.go(session.id, payload, user, yarStub)
+
+          const [flashType, notification] = yarStub.flash.args[0]
+
+          expect(flashType).to.equal('notification')
+          expect(notification).to.equal({ title: 'Added', text: 'Changes made' })
         })
       })
+      describe('with an updated note', () => {
```
Cruikshanks, line 88: NIT - whitespace

```suggestion

      describe('with an updated note', () => {
```

Cruikshanks (changes requested):

Cruikshanks (approved):

PR #924, opened 2024-04-18
Add paginator presenter

Cruikshanks:

> Part of a series of changes to migrate the legacy view bill runs page into this project

The legacy view bill runs page has a custom pagination control. We understand why, the [pagination component](https://design-system.service.gov.uk/components/pagination/) didn't exist in the design system at the time it was built. But now we have the component there is no need to migrate the custom one.

One of the key reasons to look at migrating the bill runs page is because we have other pages that need updating that also use a form of custom pagination. If we can crack how to implement the GOV.UK component we have a solution for all of them.

This `PaginatorPresenter` we're adding will build the 'object' that has to be passed to `govukPagination()` in the nunjucks view. There are several scenarios it has to consider, for example

- when there are not enough results to warrant pagination
- whether to show the previous or next controls
- when to show the ellipsis to highlight skipped pages

To do this all it needs are

- the total number of records for the thing being paginated
- how many results to show per page
- what page of results is selected

The presenter we've built is intended to be used by all pages that need the component, not just bill runs. Because of this, we've tried to ensure the testing and documentation cover all possible scenarios.


robertparkinson (dismissed):

app/presenters/paginator.presenter.js
```diff
@@ -0,0 +1,242 @@
...
+  }
+}
+
+function _component (selectedPageNumber, numberOfPages) {
+  const items = _items(selectedPageNumber, numberOfPages)
+
+  const component = { items }
+
+  if (selectedPageNumber !== 1) {
+    component.previous = { href: `/system/bill-runs?page=${selectedPageNumber - 1}` }
```
Jozzey, line 132: More of a question really. If the intention is to reuse this presenter on other pages that require this component, should we be hardcoding the URL `/system/bill-runs`? I guess currently it's fine as it is the only page that uses this presenter.  But is it worth setting up the URL as an additional parameter now, or is that something to do if/when we need to?

Jozzey (dismissed):
Just a comment to do with as you please 😁 Looks very good 👍🏼

app/presenters/paginator.presenter.js
Cruikshanks: 🎯 Brilliant spot. Yes, I should be making this a param.

Jozzey (approved):

PR #925, opened 2024-04-18
Migrate view bill runs page from legacy UI

Cruikshanks:

> Part of a series of changes to migrate the legacy view bill runs page into this project

This is the final step in migrating the legacy view bill runs page to this project. We dealt with the support services we need in previous changes. Now we need to add

- the route
- the controller handler
- the orchestration service
- the view template

We also update the menu to point to it and not the legacy `/billing/batch/list`. Worth noting (because this is the first) we use the convention of naming the handler and subsequent files `index` to reflect they deal with displaying a list of bill runs, not a single one.

Screenshot 2024-04-22 at 22 50 51

### DIFF
--- a/app/presenters/bill-runs/index-bill-runs.presenter.js
+++ b/app/presenters/bill-runs/index-bill-runs.presenter.js
@@ -52,6 +52,10 @@ function go (billRuns) {
 }
 
 function _link (billRunId, status) {
+  if (status === 'cancel') {
+    return null
+  }
+
   if (status === 'review') {
     return `/system/bill-runs/${billRunId}/review`
   }
--- a/app/presenters/bill-runs/index-bill-runs.presenter.js
+++ b/app/presenters/bill-runs/index-bill-runs.presenter.js
@@ -34,8 +34,6 @@ function go (billRuns) {
       status
     } = billRun
 
-    const type = formatBillRunType(batchType, scheme, summer)
-
     return {
       id,
       createdAt: formatLongDate(createdAt),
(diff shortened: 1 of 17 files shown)


test/presenters/bill-runs/index.bill-runs.presenter.test.js
```diff
@@ -0,0 +1,90 @@
...
+          region: 'Albion',
+          scheme: 'sroc',
+          status: 'sent',
+          total: '£300.00',
+          type: 'Supplementary'
+        }
+      ])
+    })
+
+    describe("the 'link' property", () => {
```
Jozzey, line 52: Could we add a test for the link property to test the link returned when the bill run has a status of review.

Jozzey (changes requested):
Just 1 comment

test/presenters/bill-runs/index.bill-runs.presenter.test.js
Cruikshanks: 🎯 💯

app/presenters/bill-runs/index-bill-runs.presenter.js
```diff
@@ -0,0 +1,68 @@
...
+      id,
+      netTotal,
+      numberOfBills,
+      region,
+      scheme,
+      summer,
+      status
+    } = billRun
+
+    const type = formatBillRunType(batchType, scheme, summer)
```
Beckyrose200, line 37: Just curious what made you choose to define type here and not inside the returned object like what you have done with createdAt, link, region and total?

Cruikshanks: I think it comes from a refactor. I previously needed `type` for something else but now no longer do and have overlooked updating the code. Good spot!

Jozzey (approved):

PR #927, opened 2024-04-22
Remove Licence from bill 2PT bill run during review

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4216

On the licence review page add a button for “remove from bill run”. When clicked you should be presented with a confirmation page with a confirm button, when this is clicked the licence should be removed from the current bill run.

Once the licence is removed the user should be returned to the bill run review page with a banner confirming the licence has been removed. When removed the licence will be flagged for two-part tariff supplementary billing.


app/services/bill-runs/two-part-tariff/review-bill-run.service.js
```diff
@@ -14,11 +14,12 @@ const ReviewBillRunPresenter = require('../../../presenters/bill-runs/two-part-t
  * @param {String} id The UUID for the bill run to review
  * @param {Object} payload The `request.payload` containing the filter data. This is only passed to the service when
  * there is a POST request, which only occurs when a filter is applied to the results.
- *
+  * @param {Object} sessionManager - The Hapi `request.yar` session manager passed on by the controller
+*
```
Cruikshanks, line 18: NIT

```suggestion
 * @param {Object} yar - The Hapi `request.yar` session manager passed on by the controller
 *
```

app/services/bill-runs/two-part-tariff/submit-remove-bill-run-licence.service.js
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Orchestrates removing a licence from a bill run whilst it is at the review stage. It does this by deleting all of the
+ * persisted data relating to the licence from the review tables. The licence will then be flagged for 2PT supplementary
+ * billing. If after removing a licence the bill run is empty, the bill run status will be set to set to `empty` and
+ * `true` returned so that the user is redirected back to the Bill runs page rather than Review licences.
+ *
+ * @param {string} billRunId - The UUID of the bill run that the licence is in
+ * @param {string} licenceId UUID of the licence to remove from the bill run
+ * @param {Object} sessionManager - The Hapi `request.yar` session manager passed on by the controller
```
Cruikshanks, line 18: Our convention for the documentation is to have a single line, then a space, a more detailed description (optional) and then the params. And if the function returns something that needs to be included.

```suggestion
 * Orchestrates removing a licence from a bill run whilst it is at the review stage
 *
 * It does this by deleting all of the persisted data relating to the licence from the review tables. The licence will
 * then be flagged for 2PT supplementary billing. If after removing a licence the bill run is empty, the bill run status
 * will be set to `empty` and `true` returned so that the user is redirected back to the Bill runs page rather
 * than Review bill run.
 *
 * @param {string} billRunId - The UUID of the bill run that the licence is in
 * @param {string} licenceId - UUID of the licence to remove from the bill run
 * @param {Object} yar - The Hapi `request.yar` session manager passed on by the controller
 *
 * @returns {Promise<boolean>} true if all the licences have been removed from the bill run else false
```

app/services/bill-runs/two-part-tariff/submit-remove-bill-run-licence.service.js
```diff
@@ -0,0 +1,130 @@
...
+ */
+async function go (billRunId, licenceId, yar) {
+  await _removeChargeElementReturns(billRunId, licenceId)
+  await _removeReturns(billRunId, licenceId)
+  await _removeChargeElements(billRunId, licenceId)
+  await _removeChargeReferences(billRunId, licenceId)
+  await _removeChargeVersions(billRunId, licenceId)
+  await _removeLicence(billRunId, licenceId)
```
Cruikshanks, line 26: I'd suggest moving these off into a separate function, for example, `_removeReviewData()`.

app/services/bill-runs/two-part-tariff/submit-remove-bill-run-licence.service.js
```diff
@@ -0,0 +1,130 @@
...
+
+  return allLicencesRemoved
+}
+
+async function _allLicencesRemoved (billRunId, licenceRef, yar) {
+  const { licenceCount } = await db
+    .count('id AS licenceCount')
+    .from('reviewLicences')
+    .where('billRunId', billRunId)
+    .first()
```
Cruikshanks, line 40: If we use the Objection model we can replace this with 1 line

```suggestion
  const count = await ReviewLicenceModel.query().where('billRunId', billRunId).resultSize()
```

app/services/bill-runs/two-part-tariff/submit-remove-bill-run-licence.service.js
```diff
@@ -0,0 +1,130 @@
...
+
+  if (licenceCount === 0) {
+    await db
+      .update('status', 'empty')
+      .from('billRuns')
+      .where('id', billRunId)
+
+    return true
+  }
```
Cruikshanks, line 49: If we use the Objection model we can replace this with 1 line

```suggestion
  if (count === 0) {
    await BillRunModel.query().findById(billRunId).patch({ status: 'empty' })

    return true
  }
```

app/services/bill-runs/two-part-tariff/submit-remove-bill-run-licence.service.js
```diff
@@ -0,0 +1,130 @@
...
+
+  // NOTE: The banner message is only set if licences remain in the bill run. This is because if there are no longer any
+  // licences remaining in the bill run the user is redirected to the "Bill runs" page instead of "Review licences". As
+  // the banner isn't displayed on the "Bill runs" page the message would remain in the cookie which could cause issues.
+  yar.flash('banner', `Licence ${licenceRef} removed from the bill run.`)
```
Cruikshanks, line 54: We shouldn't be doing this in this method as it is a clear violation of 'Single responsibility principle' from [SOLID](https://en.wikipedia.org/wiki/SOLID).

We can use the result this function returns in the `go()` to determine if the flash banner needs setting. That also means we do not need to pass yar or the licence ref into this function.

app/services/bill-runs/two-part-tariff/submit-remove-bill-run-licence.service.js
```diff
@@ -0,0 +1,130 @@
...
+
+async function _setSupBillingTrueAndReturnRef (licenceId) {
+  const licence = await db
+    .update('includeInSrocTptBilling', true)
+    .from('licences')
+    .where('id', licenceId)
+    .returning('licenceRef')
+
+  return licence[0].licenceRef
+}
```
Cruikshanks, line 126: If we use the Objection model we can replace this with 1 line. I would also suggest a change to the function name

```suggestion
async function _flagForSupplementaryBilling (licenceId) {
  const licence = await LicenceModel.query().findById(licenceId).patch({ includeInSrocTptBilling: true })

  return licence.licenceRef
}
```

app/views/bill-runs/remove-licence.njk
```diff
@@ -0,0 +1,60 @@
...
+            {
+              key: { text: "Billing account", classes: "meta-data__label" },
+              value: { html: '<span data-test="meta-data-account">' + billingAccount + '</span>', classes: "meta-data__value" }
+            }
+          ]
+        })
+      }}
+
+      <form method="post">
+        {{ govukButton({ text: "Remove licence" }) }}
```
Cruikshanks, line 56: The `preventDoubleClick` option needs to be enabled on the button

```suggestion
        {{ govukButton({
          text: "Remove licence",
          preventDoubleClick: true
        }) }}
```

Cruikshanks (changes requested):
I've not got into the tests because I figure it would better to hold off until some of these suggested changes are reviewed and we get an answer back from our design team.

Cruikshanks:
I've put this on-hold because I've seen that the intended 'confirm remove licence from bill run' is quite different from the one that already exists and is being used.

I'd prefer that we keep things consistent but they'll get the final say. We just need to double-check what they want to do. Like most things, they've probably forgotten or are not aware the page already exists!

app/presenters/bill-runs/two-part-tariff/remove-bill-run-licence.presenter.js
```diff
@@ -0,0 +1,46 @@
...
+ * Prepares and processes bill run data for presentation
+ *
+ * @param {module:BillRunModel} billRun - an instance of `BillRunModel`
+ * @param {Object[]} licence - an array containing the licence id, ref & account number for a single licence. For the
+ * majority of licences the array will normally contain only one element. However, licences can have multiple billing
+ * accounts so more than one element can exist for a licence.
+ *
+ * @returns {Object} - the prepared data to be passed to the remove licence template
+ */
+function go (billRun, licence) {
```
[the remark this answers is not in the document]
Jozzey: It can be an array, but the array will only contain data for a single licence. There will only be multiple elements if a licence has more than one billing account. So I would say calling it licence is correct.

app/services/bill-runs/two-part-tariff/review-bill-run.service.js
Jozzey: Yes I didn't like that either :grin:. I copied it from the original one you did in `app/services/bill-runs/two-part-tariff/review-licence.service.js` so figured it was your convention for `yar`

Jozzey: I've also sorted that out in `app/services/bill-runs/two-part-tariff/review-licence.service.js`

app/services/bill-runs/two-part-tariff/submit-remove-bill-run-licence.service.js
Jozzey: I'll make this change to use objection. Just to note though that for it to return the `licenceRef` I'll need to add `.returning('licenceRef')` to the end of it otherwise it just returns the number of records updated. This also means that it is no longer a 1 liner as it goes over the line length.

app/presenters/bill-runs/two-part-tariff/remove-bill-run-licence.presenter.js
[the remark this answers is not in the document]
Jozzey: It is not an issue now that we are not displaying the billing account number

Jozzey:
Taken off hold now we have the new page layout confirmed and built

Cruikshanks (approved):

PR #933, opened 2024-04-24
Fix from and to date to start and end date

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4296

When creating the abstraction period requirements page, **to** and **from** were used to specify the period for the return requirements. This has now been changed to **start** and **end** date to match the original error messages in the ticket above.


app/views/return-requirements/abstraction-period.njk
```diff
@@ -17,21 +17,21 @@
 {% endblock %}
 
 {# If there is a fromAbstractionPeriod error, the error class and message are set #}
```
Cruikshanks, line 19: You probably want to update the comment as well!

app/views/return-requirements/abstraction-period.njk
```diff
@@ -17,21 +17,21 @@
...
+  {% set startResultErrorClass = " govuk-input--error" %}
+  {% set startResultErrorMessage = { text: error.text.startResult } %}
 {% else %}
-  {% set fromResultErrorClass = "" %}
-  {% set fromResultErrorMessage = null %}
+  {% set startResultErrorClass = "" %}
+  {% set startResultErrorMessage = null %}
 {% endif %}
 
 {# If there is a toAbstractionPeriod error, the error class and message are set #}
```
Cruikshanks, line 28: You probably want to update the comment as well!

Cruikshanks (changes requested):
So, _so_ close! 😂

Cruikshanks (approved):

PR #936, opened 2024-04-24
Add charge reference links to licence review page

Beckyrose200:
This PR is an enhancement to the licence review page for two-part tariff bill runs. It adds a dynamic link to the charge references being displayed. The link will display 1 of two pieces of text depending on whether the charge reference has either an aggregate value or a charge adjustment factor. If it does it will display 'Change details' and if it doesn't then 'View details'.


test/presenters/bill-runs/two-part-tariff/review-licence.presenter.test.js
```diff
@@ -186,6 +187,28 @@ describe('Review Licence presenter', () => {
...
+        beforeEach(() => {
+          licence[0].reviewChargeVersions[0].reviewChargeReferences[0].chargeReference.adjustments = { charge: null, aggregate: null }
+        })
+
+        it("changes the chargeReferenceLink to 'View details'", () => {
+          const result = ReviewLicencePresenter.go(billRun, licence)
+
+          expect(result.chargeData[0].chargeReferences[0].chargeReferenceLink.linkName).to.equal('View details')
+        })
+      })
```
Cruikshanks, line 210: I'm guessing if you look at your code coverage it is going to say you are missing some on line 133 in the presenter.

It's only one more test that can be created by copying & pasting another. So, I would suggest having both `charge: 0.5` and `aggregate: 0.5` being tested.

app/presenters/bill-runs/two-part-tariff/review-licence.presenter.js
```diff
@@ -119,13 +119,24 @@ function _chargeReferenceDetails (reviewChargeVersion, chargePeriod) {
...
+function _chargeReferenceLink (reviewChargeReference) {
+  const { charge: chargeFactor, aggregate } = reviewChargeReference.chargeReference.adjustments
+
+  if (chargeFactor || aggregate) {
+    return { linkName: 'Change details' }
+  }
+
+  return { linkName: 'View details' }
+}
+
```
Cruikshanks, line 139: With no ticket so details coming by word of mouth this is my bad. But _just like_ volume, because we are going to amend these values the final design will require them being persisted to the review record alongside an amendable version as part of matching & allocating.

Using volume on the charge element as an example, there we have

- calculated: what the engine calculated
- allocated: defaults to calculated but this is what the user will change

We need the same for both aggregate and authorised volume on the charge reference. The one niggle for aggregate is when we look in the charge reference and find there is nothing we default the 'calculated' value to `1`.

So this obviously has implications on the logic. But also, when done the logic needs to be based on these values _not_ being equal to 1.

Cruikshanks (changes requested):
The details and reasons why are in the comment for the charge reference link function.

But I think we need a PR prior to this one to update the table to first capture existing values alongside amendable ones. Then this PR can be refactored to use those values.

Jozzey (approved):

Cruikshanks (approved):

PR #939, opened 2024-04-25
Saving input from return requirements in session

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4384

During the initial development of the return requirements pages, user input was not being saved in session. Later tickets and pages have this functionality in place. The session data is needed for the Check Your Answers page at the end of the journey.

This PR is focused on saving user input into the session where it is missing.


Cruikshanks (changes requested):
I hadn't realised that we'd made changes to Site Description to support replaying invalid entries in _this_ PR.

That being the case, if we are making the changes we should look to copy the pattern we used for abstraction period. Shout if that is not clear and we can quickly pair and get this change completed!

rvsiyad:
> I hadn't realised that we'd made changes to Site Description to support replaying invalid entries in _this_ PR.
> 
> That being the case, if we are making the changes we should look to copy the pattern we used for abstraction period. Shout if that is not clear and we can quickly pair and get this change completed!

I added it in a couple days ago but I have now changed it to copy the new pattern created!

Cruikshanks (approved):

PR #940, opened 2024-04-25
Fix validation messages for two-part tariff billable returns page

Beckyrose200:

During testing for the above ticket, it was found that a new validation was needed for the custom quantity that a user can enter. This quantity can not be greater than the authorised quantity on the charge element. This PR is to add this validation and to amend the original validation messages to update them as per the design team's request.


Jozzey (commented):
Just a typo, think it got overlooked in the last PR, sorry. Looks OK to me apart from that. Think Alan is also reviewing this so maybe wait and see if he spots anything.

Cruikshanks (changes requested):
I also spotted just a typo!

Jozzey (approved):

Cruikshanks (approved):

PR #946, opened 2024-04-26
Fix bill run 'busy' tags in bill runs page

Cruikshanks:

We just [fixed the status tag for cancelling bill runs](https://github.com/DEFRA/water-abstraction-system/pull/945) in the bill runs page when it was spotted the text no longer matched what you see in the legacy service.

But this made us realise the same goes for `'processing'` bill runs. They should be displayed as `'building'`. Add to that, it and the tags for anything where a bill run is 'busy' doing stuff (excluding cancelling) should all be grey.

With these changes to the tags and the need to be consistent wherever they are shown, we've decided to move the logic into the macro rather than just have it in the `index-bill-runs.presenter.js`.

A further change we have made is that any busy bill run should not be 'clickable'. It's busy because we are in the process of changing the data for the bill run. Therefore, it doesn't make sense to let users click through and see the in-progress or to-be-updated results.

So, this change also removes the link for these busy bill runs.

Screenshot 2024-04-26 at 13 18 08


rvsiyad (approved):
LGTM

PR #947, opened 2024-04-26
Add validation for amend billable returns page

Beckyrose200:

During the testing of the amended billable returns volume page, it was highlighted that there was 1 validation test missing. When a user enters a new volume for the billable returns currently the volume cannot be greater than the authorised volume on the charge element. This is correct however it should also not go above the charge reference authorised volume. So whichever volume is LOWER the authorised charge element or the authorised charge reference should be the limit for the user's input. This PR is to add in that additional validation.


app/presenters/bill-runs/two-part-tariff/amend-billable-returns.presenter.js
```diff
@@ -38,10 +37,24 @@ function go (billRun, reviewChargeElement, licenceId) {
...
+ * The user can only enter a volume on the billable returns that is less than the authorised volume. The authorised
+ * volume is either the authorised volume on the charge element or the authorised volume on the charge reference.
+ * Whichever is lower.
+ */
+function _authorisedQuantity (reviewChargeElement) {
+  const { chargeElement, reviewChargeReference } = reviewChargeElement
+  const chargeElementQuantity = chargeElement.authorisedAnnualQuantity
+  const chargeReferenceQuantity = reviewChargeReference.chargeReference.volume
+
+  return chargeElementQuantity > chargeReferenceQuantity ? chargeReferenceQuantity : chargeElementQuantity
```
Jozzey, line 55: Personally I'd favor using `Math.min` which we have done in a few other files, and would do `return Math.min(chargeElement.authorisedAnnualQuantity, reviewChargeReference.chargeReference.volume` and get rid of lines 52 & 53.

I suspect you may disagree from chats we have had previously,  in which case wonder what @Cruikshanks thinks?

test/presenters/bill-runs/two-part-tariff/amend-billable-returns.presenter.test.js
Jozzey, line 1: I'd like to see a test where the authorised quantity on the `chargeElement` is less than the `chargeReference`

Jozzey (changes requested):
Looks good, just a couple of things

Jozzey (approved):

PR #949, opened 2024-04-26
Display view licence tabs only to permitted users

jonathangoulding:
Display view licence tabs only to permitted users

https://eaflood.atlassian.net/browse/WATER-4432

> Part of a series of changes to replace the legacy view licence page

This change will only show the billing and charge information tabs on the view license page if the user has the appropriate role.


app/controllers/licences.controller.js
```diff
@@ -25,9 +25,9 @@ async function returnsRequired (request, h) {
 }
 
 async function view (request, h) {
-  const { id } = request.params
+  const { params: { id }, auth } = request
 
-  const data = await ViewLicenceService.go(id)
+  const data = await ViewLicenceService.go(id, auth)
```
jonathangoulding, line 30: I could not see the benefit adding to this test as the ViewLicenceService is mocked. 

I can not see other places where the the mocks args are tested in the rest of the code base. 

I think the presenter test is sufficient to test the auth functionality but happy to discuss.

app/presenters/licences/view-licence.presenter.js
```diff
@@ -71,7 +71,8 @@ function go (licence, licenceAbstractionConditions) {
     registeredTo,
     sourceOfSupply: abstractionDetails.sourceOfSupply,
     startDate: formatLongDate(startDate),
-    warning: _generateWarningMessage(ends)
+    warning: _generateWarningMessage(ends),
+    roles: _getAuthRoles(auth)
```
jonathangoulding, line 75: Started to get a sonar cloud error - https://sonarcloud.io/project/issues?resolved=false&severities=MAJOR&sinceLeakPeriod=true&branch=view-license-role-based-tabs-for-bills-and-charge-information&id=DEFRA_water-abstraction-system&open=AY8pIrHhp7fcpGn2GYJe

this did not happen when the notification work was done.

app/controllers/licences.controller.js
Cruikshanks: Agreed. This is why our controllers are 'dumb' and leave all the work to the services. In general, we limit our controller tests to 'when the request is valid it [returns x/redirects to y]' and 'when request is invalid it [returns z]'

Any other tests of behaviour are done directly against the service/presenter.

test/presenters/licences/view-licence.presenter.test.js
```diff
@@ -1120,6 +1123,40 @@ describe('View Licence presenter', () => {
...
+
+        expect(result.roles).to.equal(['role 1', 'role 2'])
+      })
+    })
+    describe('when the authenticated user has NO roles', () => {
+      beforeEach(() => {
+        auth = undefined
+      })
+
+      it('returns the roles in a flat array', () => {
```
Cruikshanks, line 1153: Guessing a copy & paste typo

```suggestion
      it('returns null for the roles', () => {
```

app/presenters/licences/view-licence.presenter.js
```diff
@@ -132,6 +133,12 @@ function _endDate (expiredDate) {
   return formatLongDate(expiredDate)
 }
 
+function _getAuthRoles (auth) {
+  return auth?.credentials?.roles?.map((role) => {
+    return role?.role
+  }) || null
```
Cruikshanks, line 139: Having experienced being thrown at a Node.js project when I knew very little about the language I try and keep things relatively simple.

In this case, I'd just roll it back one step and suggest this

```suggestion
  const roles = auth?.credentials?.roles?.map((role) => {
    return role?.role
  })

  return roles || null
```

Cruikshanks (changes requested):
Just a few corrections and suggestions!

app/controllers/licences.controller.js
```diff
@@ -25,9 +25,9 @@ async function returnsRequired (request, h) {
 }
 
 async function view (request, h) {
-  const { id } = request.params
+  const { params: { id }, auth } = request
```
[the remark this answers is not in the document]
jonathangoulding: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#nested_object_and_array_destructuring

For one nested object its very useful. 

If this was  `const { params: { id, fella }, auth } = request` and `fella` did not exist then fella would be undefined. 

If this was  `const { params: { id, fella: { thing }  }, auth } = request` and `thing` did not exist then if would error: 
"Cannot read properties of undefined (reading 'thing')"

Cruikshanks (approved):

PR #952, opened 2024-04-29
Add two-part tariff charge reference details review page

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4438

As part of the ongoing work for the two-part tariff review screens, this PR introduces the new details page for a charge reference. The charge references have different charge factors that are displayed on these screens and with future logic can be changed/updated. This initial PR is just the page with its content and no update buttons.


app/presenters/bill-runs/two-part-tariff/charge-reference-details.presenter.js
```diff
@@ -0,0 +1,117 @@
...
+
+  return { adjustments, hasAggregateOrChargeFactor }
+}
+
+function _financialYear (financialYearEnding) {
+  const startYear = financialYearEnding - 1
+  const endYear = financialYearEnding
+
+  return `${startYear} to ${endYear}`
+}
```
Cruikshanks, line 96: You can now reuse the `formatFinancialYear()` in `app/presenters/base.presenter.js` to do this.

app/presenters/bill-runs/two-part-tariff/charge-reference-details.presenter.js
```diff
@@ -0,0 +1,117 @@
...
+
+function _totalBillableReturns (reviewChargeElements) {
+  let totalBillableReturns = 0
+
+  reviewChargeElements.forEach((reviewChargeElement) => {
+    totalBillableReturns += reviewChargeElement.amendedAllocated
+  })
+
+  return totalBillableReturns
```
Cruikshanks, line 112: I think this is one of those rare cases where `reduce()` is actually applicable!

```suggestion
  return reviewChargeElements.reduce((total, reviewChargeElement) => {
    total += reviewChargeElement.amendedAllocated

    return total
  }, 0)
```

Cruikshanks (changes requested):
Great! Just got a couple of suggested changes to the presenter.

Cruikshanks (approved):

PR #956, opened 2024-04-30
Replay submitted data with a validation error

rvsiyad:
This PR is focused on replaying submitted start dates with a validation error for page 1 in the return requirements journey.

When a user inputs an invalid start date that returns a validation error, the page will clear the input. This needs to be fixed so that data with errors is replayed to the user so they can make changes to what they have submitted.


Cruikshanks (approved):
Just one request. Could you update the PR description/squashed commit message to specify which page we are doing this to? 🙇

PR #958, opened 2024-04-30
add financial agreements and licence agreements models

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4442

As part of the work to build return requirements from abstraction data we need to know if the licence has been marked for two part tariff. This PR adds the models for the tables that contain that information.


db/migrations/public/20240430133702_water-add-licence-agreements-view.js
```diff
@@ -0,0 +1,29 @@
...
+
+const viewName = 'licence_agreements'
+
+exports.up = function (knex) {
+  return knex
+    .schema
+    .createView(viewName, (view) => {
+      view.as(knex('licence_agreements').withSchema('water').select([
+        'licence_agreement_id AS id',
+        'financial_agreement_type_id',
```
Cruikshanks, line 11: As our view name is different from the table name we'll need to alias this field as well.

```suggestion
        'financial_agreement_type_id AS financial_agreement_id',
```

Cruikshanks (changes requested):
Because I've asked for the view to use a different name from the source table (`financial_agreements` vs `financial_agreement_types`) it has a knock on effect wit foreign keys.

I'll keep checking the rest but figured I'd flag this now because it will also impact the unit tests and the setup of test data.

db/migrations/public/20240430133702_water-add-licence-agreements-view.js
robertparkinson: Is it weird that the change didn't break any of the tests?

robertparkinson: ah it failed on the pipeline so I've done something wrong locally when tryign to redo the migrations

db/migrations/public/20240430133702_water-add-licence-agreements-view.js
```diff
@@ -0,0 +1,29 @@
...
+    .schema
+    .createView(viewName, (view) => {
+      view.as(knex('licence_agreements').withSchema('water').select([
+        'licence_agreement_id AS id',
+        'financial_agreement_type_id AS financial_agreement_id',
+        'licence_ref',
+        'start_date',
+        'end_date',
+        'date_signed',
+        'date_deleted',
```
Cruikshanks, line 16: Sorry, @robertparkinson 😳 

Missed this on first pass. We alias this to `deleted_at` to be consistent with `created_at` and `updated_at`.

app/services/return-requirements/fetch-licence-agreements.service.js
Cruikshanks, line 1: Referring to the comments.

I thought we were building this to support creating a return requirement from existing abstraction data?

We need to know whether a licence is currently linked to 'Spary/Trickle' irrigation _and_ has a 2PT agreement. I'm not sure this is directly supporting the check your answers page but I could be wrong.

Also, for future reference ideally this would be shifted into a separate change away from the changes needed for the models. But don't panic for this PR.

app/services/return-requirements/fetch-licence-agreements.service.js
```diff
@@ -0,0 +1,41 @@
...
+ * Fetches licence agreements needed for `/return-requirements/{sessionId}/check-your-answers` page
+ * @module FetchLicenceAgreementsService
+ */
+
+const LicenceAgreementModel = require('../../models/licence-agreement.model.js')
+
+/**
+ * Fetches licence agreements needed for `/return-requirements/{sessionId}/check-your-answers` page
+ *
+ * @param {string} licenceRef - The UUID for the licence to fetch
```
Cruikshanks, line 13:
```suggestion
 * @param {string} licenceRef - The reference of the licence to fetch agreements for
```

Cruikshanks (changes requested):
My bad for a missed field name!

But I think the documentation needs updating in the fetch service unless I'm mistaken.

Cruikshanks (approved):

PR #959, opened 2024-04-30
Return to Check Your Answers after making a change

rvsiyad:
This is a PR focused on adding the ability to return to the Check Your Answers page after updating a previously submitted page in the return requirements journey.

When a user reaches the Check Your Answers page at the end of the return requirements journey, they are given the ability to update previously entered options. After making a change, users should be navigated back to the Check Your Answers page instead of continuing through the already completed journey.

### DIFF
--- a/app/controllers/return-requirements.controller.js
+++ b/app/controllers/return-requirements.controller.js
@@ -204,6 +204,10 @@ async function submitAbstractionPeriod (request, h) {
     return h.view('return-requirements/abstraction-period.njk', pageData)
   }
 
+  if (pageData.checkYourAnswersVisited) {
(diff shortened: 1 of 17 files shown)


rvsiyad:
SonarCloud is currently failing for this PR as it is complaining about the lack of test code coverage for the Return Requirements Controller. This is as a result of missing test cases for POST requests for the Return Requirements journey.

After consulting on the issue with @Cruikshanks, the tests will be implemented at a later date when the Return Requirements journey is fully completed.

SonarCloud can be ignored for now :tada: :metal:

Cruikshanks (approved):
🥂 🎉 !!

But just a heads up, we'll be returning to this for 'reasons' 😏 😁

PR #960, opened 2024-04-30
Add .editorconfig to keep project files consistent

Cruikshanks:

We've been fortunate enough to grow as a team this year which means we've had several new developers join us! 🥳

Unfortunately, it has exposed our onboarding still leaves a lot to be desired. We find we keep having to highlight the same coding convention or stylistic issues in new members' initial PRs. This is irrespective of whether we have written up the convention or not.

Wouldn't it be great if we used all this computing power we interact with daily to help them get it right the first time? That we help them fall into the [pit of success](https://www.hackterms.com/pit%20of%20success)?

One idea is to add an [.editorconfig](https://editorconfig.org/) to the repo. Depending on folks IDE choice it will either be recognised out-of-the-box or they'll need an extension. 

Unlike other tools having the file doesn't do anything _directly_ to your project. Instead, it tells your chosen editor how to behave when working with the files. For example, when you indent something whether to use tabs or spaces and how many. If your editor can remove whitespace on save, the `.editorconfig` will tell it whether to do that or not.

Even for folks using an incompatible editor or who choose not to install the extension, we at least have something that defines what we expect and can be used in our CI pipeline.

This change adds an initial version generated using the [VSCode extension](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig) based on current editor settings.

It also updates our GitHub CI workflow to check that pushed changes match our conventions.

---

This change highlighted that the line endings in our migrations were inconsistent. So, the fix for those is included.


Cruikshanks:
This looks like a scary amount of changes. But when you go to look at the files, close the `db` folder and just focus on the other files.

Screenshot 2024-04-30 at 21 47 13

Adding this and the CI action highlighted that the migration files had inconsistent line endings. I've corrected them in this change to get CI to pass and to start with no errors. But that means Git thinks I've changed thousands of lines of code!

jonathangoulding (commented):
Looks like it works for me in Webstorm. 

I think we merge this to main then check with eslint branch to make sure they are consistent and not fighting each other.

Cruikshanks:
Agreed @jonathangoulding . I'm not over the moon about the CI choice/option. There seems little support for using a stock GitHub action when it comes to linting based on your .editorconfig. But that might be because everyone _is_ using their stock linter + plugin/extension like you suggest.

So, defo once this and the ESlint branch are merged we should look into whether [eslint-plugin-editorconfig](https://www.npmjs.com/package/eslint-plugin-editorconfig) is a better solution.

Jozzey (approved):

Beckyrose200 (approved):

PR #980, opened 2024-05-05
Add new fetch return requirement session service

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4463

> This is part of changes to support adding and changing multiple return requirements in the setup journey

In this change rather than having each service in the journey fetching the session directly we move that work into a new service. In that service, we can do the work that is currently duplicated in services and presenters to make properties stored in `data` available at the top level.

Once this is in place, we can then extend the service to also determine _which_ requirement the user is currently dealing with. But that comes later. First, we need to get everything using our new service!


Cruikshanks:
Closing because we decided instead to go with [Enhance the SessionModel to expose and update data](https://github.com/DEFRA/water-abstraction-system/pull/981)

PR #986, opened 2024-05-07
Add View Licence Bills page

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4316

This change will introduce a bills controller, service and presenter to handle the view licence bills page.

This will share the same view as the summary page and load the same 'common data' established in [previous work](#957).

### DIFF
--- a/app/presenters/licences/view-licence-bills.presenter.js
+++ b/app/presenters/licences/view-licence-bills.presenter.js
@@ -31,7 +31,7 @@ function _formatBillsToTableRow (bills) {
       billNumber: bill.invoiceNumber,
       dateCreated: formatLongDate(new Date(bill.createdAt)),
       account: bill.accountNumber,
-      runType: bill.batchType,
+      runType: bill.billRun.batchType,
       financialYear: bill.financialYearEnding,
       total: _formatCurrencyToGBP(bill.netAmount),
       accountId: bill.billingAccountId,
--- a/app/presenters/licences/view-licence-bills.presenter.js
+++ b/app/presenters/licences/view-licence-bills.presenter.js
@@ -1,0 +1,21 @@
+'use strict'
+
+/**
+ * Formats data for the `/licences/{id}/bills` view licence bill page
+ * @module ViewLicenceBillsPresenter
+ */
+
+/**
+ * Formats data for the `/licences/{id}/bills` view licence bill page
+ *
+ * @returns {Object} The data formatted for the view template
+ */
+function go (billsData) {
+  return {
+    activeTab: 'bills'
+  }
+}
+
+module.exports = {
+  go
+}
--- a/app/presenters/licences/view-licence-bills.presenter.js
+++ b/app/presenters/licences/view-licence-bills.presenter.js
@@ -5,15 +5,39 @@
  * @module ViewLicenceBillsPresenter
  */
 
+const { formatLongDate } = require('../base.presenter')
+
 /**
  * Formats data for the `/licences/{id}/bills` view licence bill page
  *
  * @returns {Object} The data formatted for the view template
  */
-function go (billsData) {
+function go (bills) {
   return {
-    activeTab: 'bills'
+    activeTab: 'bills',
+    bills: _formatBillsToTableRow(bills)
   }
+}
+
+function _formatCurrencyToGBP (amount) {
+  return amount.toLocaleString('en-gb', {
+    style: 'currency',
+    currency: 'GBP'
+  })
+}
+function _formatBillsToTableRow (bills) {
+  return bills.map((bill) => {
+    return {
+      billNumber: bill.invoiceNumber,
+      dateCreated: formatLongDate(new Date(bill.createdAt)),
+      account: bill.accountNumber,
+      runType: bill.batchType,
+      financialYear: bill.financialYearEnding,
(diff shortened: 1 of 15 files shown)


app/presenters/licences/view-licence-bills.presenter.js
```diff
@@ -0,0 +1,45 @@
...
+ */
+function go (bills) {
+  return {
+    activeTab: 'bills',
+    bills: _formatBillsToTableRow(bills)
+  }
+}
+
+function _formatCurrencyToGBP (amount) {
+  return amount.toLocaleString('en-gb', {
```
jonathangoulding, line 23: This is not doing to two decimal places. 

Need to see how other areas of the code are handling this

app/services/licences/fetch-licence-bills.service.js
```diff
@@ -0,0 +1,57 @@
...
+      'id',
+      'licenceRef'
+    ])
+
+  const billLicences = await BillLicenceModel.query()
+    .select('*')
+    .where('billLicences.licence_ref', licence.licenceRef)
+    .page(page - 1, DatabaseConfig.defaultPageSize)
+
+  const billIds = JSON.parse(JSON.stringify(billLicences)).results.map(b => b.billId)
```
jonathangoulding, line 40: Change this

app/services/licences/view-licence-bills.service.js
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @returns {Promise<Object>} an object representing the `pageData` needed by the licence summary template.
+ */
+async function go (licenceId, auth, page) {
+  const commonData = await ViewLicenceService.go(licenceId, auth)
+
+  const billsData = await FetchLicenceBillsService.go(licenceId, page)
+  const pageData = ViewLicenceBillsPresenter.go(billsData.bills)
+
+  const pagination = PaginatorPresenter.go(100, Number(page), `/system/licences/${licenceId}/bills`)
```
jonathangoulding, line 26: confirm this works with a bigger license

app/services/licences/fetch-licence-bills.service.js
jonathangoulding: Fixed.

app/services/licences/view-licence-bills.service.js
jonathangoulding: I have checked and this does work when the numbers are fudged. But it would be better to test with a licence with more data.

app/presenters/licences/view-licence-bills.presenter.js
```diff
@@ -0,0 +1,39 @@
...
+  }
+}
+
+function _formatBillsToTableRow (bills) {
+  return bills.map((bill) => {
+    return {
+      billNumber: bill.invoiceNumber,
+      dateCreated: formatLongDate(new Date(bill.createdAt)),
+      account: bill.accountNumber,
+      runType: bill.billRun?.batchType,
```
robertparkinson, line 28: you're saying here that billRun may be undefined but not doing anything if it is from what I can see? will this result in undefined being output on the screen?

app/services/licences/fetch-licence-bills.service.js
```diff
@@ -0,0 +1,50 @@
+'use strict'
+
+/**
+ * Fetches all return logs for a licence which is needed for the view '/licences/{id}/returns` page
```
robertparkinson, line 4: I think the `/licences/{id}/returns` needs to be `/licences/{id}/bills` ?

robertparkinson (changes requested):

app/presenters/licences/view-licence-bills.presenter.js
jonathangoulding: When testing if `billRun` did not exists there would be an error thrown (crashing the server potentially) this is to cover that exception. 

But the code below should always return a `billRun` object. So i can remove the optional chaining if desired. 
```
    .withGraphFetched('billRun')
    .modifyGraph('billRun', (builder) => {
      builder.select(['batchType'])
    })
```

The  current result would be a empty as below. 

Screenshot 2024-05-07 at 14 59 16

I thought this would be better to have then not have.

app/services/licences/fetch-licence-bills.service.js
jonathangoulding: Done

app/presenters/licences/view-licence-bills.presenter.js
jonathangoulding: In think in this case i think it's nunjucks - https://mozilla.github.io/nunjucks/templating.html#variables

`If a value is undefined or null, nothing is displayed. The same behavior occurs when referencing undefined or null objects. The following all output nothing if foo is undefined: {{ foo }}, {{ foo.bar }}, {{ foo.bar.baz }}.`

But i am not 100%.

Cruikshanks: You can always assume that a bill run exists. There would never be a scenario where a bill existed but the bill run it linked to didn't.

jonathangoulding: Optional chaining removed

robertparkinson (approved):

PR #988, opened 2024-05-07
Select additional submission options page

Demwunz:
https://eaflood.atlassian.net/browse/WATER-4306

Additional options for submission page.

This branch will provide options for "multiple upload", or "none".

Appropriate notifications will be shown to user based on selection.

### DIFF
--- a/app/presenters/base.presenter.js
+++ b/app/presenters/base.presenter.js
@@ -1,27 +1,6 @@
 'use strict'
 
 /**
- * Capitalize the first letter of each word in a string
- *
- * Will work for strings containing multiple words or only one.
- *
- * @param {string} value The string to capitalize
- *
- * @returns {string} The capitalized string
- */
-function capitalize (value) {
-  const words = value.split(' ')
-  const capitalizedWords = []
-
-  words.forEach((word) => {
-    const capitalizedWord = word.charAt(0).toUpperCase() + word.slice(1)
-    capitalizedWords.push(capitalizedWord)
-  })
-
-  return capitalizedWords.join(' ')
-}
-
-/**
  * Converts a number which represents pence into pounds by dividing it by 100
  *
  * This is such a simply calculation it could be done in place. But by having it as a named method we make it clear
@@ -52,7 +31,7 @@ function convertPenceToPounds (value) {
 function generateBillRunTitle (regionName, batchType, scheme, summer) {
   const billRunType = formatBillRunType(batchType, scheme, summer)
 
-  return `${capitalize(regionName)} ${billRunType.toLowerCase()}`
+  return `${titleCase(regionName)} ${billRunType.toLowerCase()}`
 }
 
 /**
@@ -99,7 +78,7 @@ function formatAbstractionPeriod (startDay, startMonth, endDay, endMonth) {
  */
 function formatBillRunType (batchType, scheme, summer) {
   if (batchType !== 'two_part_tariff') {
-    return capitalize(batchType)
+    return titleCase(batchType)
   }
 
   if (scheme === 'sroc') {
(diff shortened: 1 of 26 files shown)


test/services/return-requirements/submit-additional-submission-options.service.test.js
Cruikshanks, line 1: I don't think we need to test both a new option and an updated option. If we did something different in the service for each scenario I think it would be worth it. But as we don't we can skip it.

This can just follow the pattern laid out in ones like `test/services/return-requirements/submit-agreements-exceptions.service.test.js`, with the additional check for the notification message.

Cruikshanks (changes requested):
I don't often say this, but I think we might have _too_ many tests! 😂 

If you _really_ think we should be testing both setting and updating values in our pages do please argue the case. But be warned, if I end up agreeing then the principle applies to _all_ the other pages! 😜

test/services/return-requirements/submit-additional-submission-options.service.test.js
Demwunz: I was in autopilot from doing the notes :)

removed the "updated" tests. should be simpler now.

Cruikshanks (changes requested):
Really sorry @Demwunz I should have spotted the single option select thing on my first pass.

app/services/return-requirements/submit-additional-submission-options.service.js
```diff
@@ -0,0 +1,100 @@
...
+ * @param {Object} payload - The submitted form data
+ * @param {Object} yar - The Hapi `request.yar` session manager passed on by the controller
+ *
+ * @returns {Promise<Object>} If no errors it returns an empty object else the page data for the note page including the
+ * validation error details
+ */
+async function go (sessionId, payload, yar) {
+  const session = await SessionModel.query().findById(sessionId)
+  // Ensure 'additional-submission-options' is an array and assign it to options
+  const options = payload['additional-submission-options'] = _ensureIsArray(payload['additional-submission-options'])
```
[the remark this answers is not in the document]
Demwunz: I've reverted the function to how I set it up before. I only changed it to this, so that SonarCloud would allow it through the "20 lines" rule.

I looked at `_handleOneOptionSelected()`, and it manipulates `payload` outside of it's scope, but this also means it doesn't have proper closure.

This is why I am using `_ensureIsArray()` which does the same thing, but returns properly scoped. I also wrote the function so it could potentially be used as a generic helper function.

test/services/return-requirements/submit-additional-submission-options.service.test.js
```diff
@@ -0,0 +1,107 @@
...
+    yarStub = { flash: Sinon.stub() }
+  })
+
+  afterEach(() => {
+    Sinon.restore()
+  })
+
+  describe('when called', () => {
+    describe('with a valid payload', () => {
+      describe('that is a new submission option', () => {
```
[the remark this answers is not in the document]
Demwunz: dropped

app/controllers/return-requirements.controller.js
```diff
@@ -10,6 +10,7 @@ const AddService = require('../services/return-requirements/add.service.js')
 const AgreementsExceptionsService = require('../services/return-requirements/agreements-exceptions.service.js')
 const CancelService = require('../services/return-requirements/cancel.service.js')
 const CheckService = require('../services/return-requirements/check.service.js')
+const AdditionalSubmissionOptionsService = require('../services/return-requirements/additional-submission-options.service.js')
```
Cruikshanks, line 13: NIT - think this just needs to go up a few lines to be under `AbstractionPeriodService`.

app/services/return-requirements/submit-additional-submission-options.service.js
```diff
@@ -0,0 +1,109 @@
...
+    pageTitle: 'Select any additional submission options for the return requirements',
+    ...submittedSessionData
+  }
+}
+
+/**
+ * When a single agreement and exception is checked by the user, it returns as a string. When multiple agreements and
+ * exceptions are checked, the 'additionalSubmissionOptions' is returned as an array.
+ * This function works to make those single
+ * selected string 'additionalSubmissionOptions' into an array for uniformity.
```
Cruikshanks, line 60: Just needs an update after the copy and paste.

```suggestion
 * When a single additional submission option is checked by the user, it returns as a string. When multiple options are
 * checked, the 'additionalSubmissionOptions' is returned as an array. This function works to make those single selected
 * string 'additionalSubmissionOptions' into an array for uniformity.
```

app/validators/return-requirements/additional-submission-options.validator.js
```diff
@@ -0,0 +1,47 @@
...
+ * @returns {Object} The result from calling Joi's schema.validate(). If any errors are found the `error:` property will
+ * also exist detailing what the issue is.
+ */
+function go (payload) {
+  /**
+  * NOTE: When a single point is checked by a user, it returns as a string.
+  * When multiple additionalSubmissionOptions are checked, the 'payload' is returned as an array.
+  * To make Joi validation straightforward, if the "payload.additionalSubmissionOptions" is a string,
+  * it is turned into an array and validated as such.
+  */
```
Cruikshanks, line 28: I think this comment can be deleted now.

```suggestion
```

Cruikshanks (changes requested):
Just some really minor NITs left to resolve then all done!

Cruikshanks (approved):

PR #991, opened 2024-05-08
Switch to new eslint config format

Cruikshanks:

In [Add linting using eslint with standard as the base](https://github.com/DEFRA/water-abstraction-system/pull/948) we switched from using [standard](https://www.npmjs.com/package/standard) directly for linting to using [ESlint](https://eslint.org/).

This is all part of the work we are doing to start codifying our conventions rather than writing them up for no one to read! 😁

We then went to make a config change but saw that we were using a deprecated core rule and instead needed to [use @stylistic/eslint-plugin-js for max-len rule](https://github.com/DEFRA/water-abstraction-system/pull/989). So, we fixed that! 💪

But then we clocked ESLint was telling us our `.eslintrc.js` [config file format is deprecated](https://eslint.org/docs/latest/use/configure/configuration-files-deprecated)! 😩 We instead should be using the `eslint.config.js` [flat file config](https://eslint.org/docs/latest/use/configure/configuration-files).

So, before we start making changes to our config this change gets us using the latest supported format.


Cruikshanks:
This has been superseded by https://github.com/DEFRA/water-abstraction-system/pull/1476

PR #993, opened 2024-05-08
Add View Licence contact details page

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4433

This change will introduce a contact details controller, service and presenter to handle the view licence contact details page.

This will share the same view as the summary page and load the same 'common data' established in [previous work](#957).


test/controllers/licences.controller.test.js
```diff
@@ -199,6 +201,49 @@ describe('Licences controller', () => {
...
+
+        expect(response.statusCode).to.equal(200)
+        expect(response.payload).to.contain('Contact Details')
+        // Table row titles
+        expect(response.payload).to.contain('Name')
+        expect(response.payload).to.contain('Communication type')
+        expect(response.payload).to.contain('Send to')
+      })
+    })
+    describe('when a request is valid and has no contact details', () => {
```
Cruikshanks, line 232: NIT - whitespace

```suggestion

    describe('when a request is valid and has no contact details', () => {
```

Cruikshanks (changes requested):
Putting this in just to record I'm still looking at this PR rather than this is all I found!

I'm digging into the Fetch as behind the scenes I can see it's running some big queries but I need a sec to get a handle on what it is currently doing before I feel I can pass any judgement.

Cruikshanks (approved):

PR #994, opened 2024-05-08
Change JOI char count validation to match GOVUK component

Demwunz:
GOVUK character count goes to 500, but JOI seems to stop at 499.

Increasing to 501 ensures they are both in sync.

This is in validator for the `add-note` page.


Cruikshanks (commented):
I don't think this change is needed. I've used max length in the Joi validation before and I thought it was fine. So, went and had a play.

I reset the validator back to using `.max(500` and then added these tests.

```javascript
describe('Note validator', () => {
  describe.only('alan has a play', () => {
    describe('500 character note', () => {
      it('passes validation', () => {
        const note = 'jHaVSQXT3Se9gxzN4AjTnY17QlnhCRaCGi5ALFugKdQGt7dyVE6TxYMRzhYzv6OOCCdwf4tLwftNwwiv67mMw7LSx7dcltPQgLzyftVjm20Zskvir9OrsWRHajdITqSmJr5PIggs3rzP6zkunNdNMxDzH1s35CREMaTqW9O0mBU6IEcBQusZA1BXsr6VTidCsNnq4XvYWO9GU3r3sh1Q2PUGIwBLIF9hHpUa0IOGc83byJ6wBAbWXnTYPKMYtSx4l9hBv6fwO4rrocaHYdwFSOMFdmIWvJWi3h0Sd3An9GzH3mSoaYBLxGVqiACdD3LrmHfLZVBbXa8nvJY9dcKkndoFW8YixULWtxhzCOjJpEL1WjqkmXn65ynskGuFWaqTrBYnrf1Uft0v0xQd49aGch4i5EiO7zjOg7Hre2w94C8U7kZ0Duw970Fq3Ii6DvE0JFbAnkMsxFFKSI9XQQaoDFDtoTS1LeQchVnAW4ydCMzFBvGJytk2'
        const result = NoteValidator.go({ note })

      })
    })

    describe('501 character note', () => {
      it('fails validation', () => {
        const note = 'jHaVSQXT3Se9gxzN4AjTnY17QlnhCRaCGi5ALFugKdQGt7dyVE6TxYMRzhYzv6OOCCdwf4tLwftNwwiv67mMw7LSx7dcltPQgLzyftVjm20Zskvir9OrsWRHajdITqSmJr5PIggs3rzP6zkunNdNMxDzH1s35CREMaTqW9O0mBU6IEcBQusZA1BXsr6VTidCsNnq4XvYWO9GU3r3sh1Q2PUGIwBLIF9hHpUa0IOGc83byJ6wBAbWXnTYPKMYtSx4l9hBv6fwO4rrocaHYdwFSOMFdmIWvJWi3h0Sd3An9GzH3mSoaYBLxGVqiACdD3LrmHfLZVBbXa8nvJY9dcKkndoFW8YixULWtxhzCOjJpEL1WjqkmXn65ynskGuFWaqTrBYnrf1Uft0v0xQd49aGch4i5EiO7zjOg7Hre2w94C8U7kZ0Duw970Fq3Ii6DvE0JFbAnkMsxFFKSI9XQQaoDFDtoTS1LeQchVnAW4ydCMzFBvGJytk2z'
        const result = NoteValidator.go({ note })

        expect(result.error.details[0].message).to.equal('Enter no more than 500 characters')
      })
    })
  })
})
```

This is the output I got

```
Note validator
  alan has a play
    500 character note
      ✔ 909) passes validation (3 ms)
    501 character note
      ✔ 910) fails validation (1 ms)
```

In the UI on the Note page I see this using the 500 character string.

Screenshot 2024-05-10 at 11 49 35

And it lets me save it.

Screenshot 2024-05-10 at 11 50 14

So, I'm not sure we need this change. Happy to pair to dig into what the problem is 🤷

Demwunz:
As discussed with Ven this morning, the current validation and behaviour is correct. The phantom extra characters would seem to be coming from Windows/MS Word.

This PR is no longer required.

PR #996, opened 2024-05-09
You are about to remove these return requirements

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4304

During the return requirements journey, users are given the option to add more requirements to the returns. As a result, users need to have an option to remove a single requirement at a time.

This PR is focused on setting up the page and functionality for users to remove a single requirement.


app/presenters/return-requirements/remove.presenter.js
```diff
@@ -0,0 +1,71 @@
...
+    return null
+  }
+
+  let returnsCycle = requirement.returnsCycle.charAt(0).toUpperCase() + requirement.returnsCycle.slice(1)
+  returnsCycle = returnsCycle.replaceAll('-', ' ')
+
+  const { frequencyReported } = requirement
+  const { siteDescription } = requirement
+
+  return `${returnsCycle} ${frequencyReported} requirements for returns, ${siteDescription}.`
```
Cruikshanks, line 49: I think we can simplify this a little and copy what we're now doing in `CancelPresenter`.

```suggestion
  const { frequencyReported, returnsCycle, siteDescription } = requirement
  const cycle = returnsCycle === 'summer' ? 'Summer' : 'Winter and all year'

  return `${cycle} ${frequencyReported} requirements for returns, ${siteDescription}.`
```

I've dropped the check on `returnsCycle` being null because we should only be getting to this page with a 'complete' return requirement.

app/presenters/return-requirements/remove.presenter.js
```diff
@@ -0,0 +1,71 @@
...
+}
+
+function _startDate (session) {
+  const selectedOption = session.startDateOptions
+  let date
```
Cruikshanks, line 54: NIT - adding some whitespace between variable types

```suggestion

  let date
```

app/routes/return-requirement.routes.js
```diff
@@ -341,6 +341,32 @@ const routes = [
...
+    method: 'POST',
+    path: '/return-requirements/{sessionId}/remove/{requirementIndex}',
+    handler: ReturnRequirementsController.submitRemove,
+    options: {
+      auth: {
+        access: {
+          scope: ['billing']
+        }
+      },
+      description: 'You are about to remove these requirements for returns'
```
Cruikshanks, line 367: NIT

```suggestion
      description: 'Submit you are about to remove these requirements for returns'
```

app/services/return-requirements/submit-remove.service.js
```diff
@@ -0,0 +1,34 @@
...
+ * is deleted from the session in the database.
+ *
+ * @param {string} sessionId - The UUID for the return requirement setup session record
+ *
+ * @returns {Promise} the promise returned is not intended to resolve to any particular value
+ */
+async function go (sessionId, requirementIndex) {
+  const session = await SessionModel.query().findById(sessionId)
+
+  return await _removeRequirementFromSession(session, requirementIndex)
```
Cruikshanks, line 23: As we are awaiting this in the controller we don't need to call `await` here.

```suggestion
  return _removeRequirementFromSession(session, requirementIndex)
```

test/controllers/return-requirements.controller.test.js
```diff
@@ -274,6 +275,22 @@ describe('Return requirements controller', () => {
     })
   })
 
+  describe('GET /return-requirements/{sessionId}/remove', () => {
+    beforeEach(async () => {
+      Sinon.stub(RemoveService, 'go').resolves({
+        id: '8702b98f-ae51-475d-8fcc-e049af8b8d38', pageTitle: 'You are about to remove these requirements for returns'
+      })
+    })
+    describe('when the request succeeds', () => {
```
Cruikshanks, line 284: NIT - Adding some whitespace

```suggestion
    })

    describe('when the request succeeds', () => {
```

app/presenters/return-requirements/remove.presenter.js
```diff
@@ -0,0 +1,71 @@
...
+ */
+function go (session, requirementIndex) {
+  const { id: sessionId, licence, requirements } = session
+  const requirement = requirements[requirementIndex]
+
+  return {
+    backLink: _backLink(session),
+    licenceId: licence.id,
+    licenceRef: licence.licenceRef,
+    returnRequirement: _formattedReturnRequirement(requirement) ? _formattedReturnRequirement(requirement) : null,
```
Cruikshanks, line 26: I've suggested a change which means you should never get `null` back from the function. But FYI, the function was determining whether null or a value was being returned. There wasn't any need for the ternary here! 😁 

```suggestion
    returnRequirement: _formattedReturnRequirement(requirement),
```

Cruikshanks (changes requested):
Just some minor NITS and suggestions.

Cruikshanks (approved):

PR #999, opened 2024-05-10
Use `yar` to store 2PT Review Licences filter data

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4462

As part of adding pagination to the results on the two-part tariff "Review Licences" page. The way the data from the filter is passed back to the view needs to change.

How it is currently set up will mean that once the user selects a different page using the pagination, which triggers a new GET request, the filter data will be lost as it isn't persisted anywhere.

In this PR the filter data is going to be persisted using `yar` to resolve the issue.


app/controllers/bill-runs.controller.js
```diff
@@ -185,6 +186,13 @@ async function submitRemoveLicence (request, h) {
   return h.redirect(`/system/bill-runs/${billRunId}/review`)
 }
 
+async function submitReview (request, h) {
+  const { id } = request.params
+  await SubmitReviewBillRunService.go(request.payload, request.yar)
```
Cruikshanks, line 191: NIT - add some whitespace

```suggestion

  await SubmitReviewBillRunService.go(request.payload, request.yar)
```

app/services/bill-runs/two-part-tariff/review-bill-run.service.js
```diff
@@ -12,17 +12,17 @@ const ReviewBillRunPresenter = require('../../../presenters/bill-runs/two-part-t
...
  *
  * @returns {Promise<Object>} An object representing the `pageData` needed by the review bill run template. It contains
  * details of the bill run and the licences linked to it as well as any data that has been used to filter the results.
  */
-async function go (id, payload, yar) {
-  const filterIssues = payload?.filterIssues
-  const filterLicenceHolder = payload?.filterLicenceHolder
-  const filterLicenceStatus = payload?.filterLicenceStatus
+async function go (id, yar) {
+  const filters = yar.get('filters')
```
Cruikshanks, line 21: Just to avoid any collisions in the future I think we should give this a less generic name.

```suggestion
  const filters = yar.get('reviewFilters')
```

app/services/bill-runs/two-part-tariff/submit-review-bill-run.service.js
```diff
@@ -0,0 +1,33 @@
...
+async function go (payload, yar) {
+  const clearFilters = payload?.clearFilters
+  const filterIssues = payload?.filterIssues
+  const filterLicenceHolder = payload?.filterLicenceHolder
+  const filterLicenceStatus = payload?.filterLicenceStatus
+
+  if (clearFilters) {
+    yar.clear('filters')
+  } else {
+    yar.set('filters', {
```
Cruikshanks, line 23: As per my previous suggestion to avoid any possible issues should we add filters elsewhere in the service.

```suggestion
    yar.clear('reviewFilters')
  } else {
    yar.set('reviewFilters', {
```

Cruikshanks (changes requested):
Brilliant! I've just got a couple of minor tweaks. But I think this is going to make for a better experience all round!

Cruikshanks (approved):

PR #1000, opened 2024-05-10
Suggested View contact details changes

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4433

This relates to [Add View Licence contact details page](https://github.com/DEFRA/water-abstraction-system/pull/993). What started as experimenting with an alternate fetch implementation has grown into a detailed comparison of the original changes and our current patterns and conventions.

In this we have tried to highlight how we think the changes would be implemented if they followed our current norms.

These are suggested and can be challenged. But we felt in this case "show don't tell" would be more appropriate to support the conversation!


app/presenters/licences/licence-contacts.presenter.js
```diff
@@ -0,0 +1,61 @@
...
+
+  if (customerContact) {
+    return customerContact.companyId
+  }
+
+  return null
+}
+
+function _licenceContactName (contact) {
+  if (contact.contactId) {
```
jonathangoulding, line 33: This logic does not match the previous changes. Both First AND last name should be present. 

This is the old UI code: 

```
     <td class="govuk-table__cell" scope="row">
                        {% if contact.firstName and contact.lastName %}
                            {{ contact.firstName }} {{contact.lastName}}
                        {% else %}
                            {{ contact.name }}
                        {% endif %}
                      </td>
```

jonathangoulding: Contact name being the company name.

Cruikshanks: So, I believe in the legacy UI code 'contact' is not a direct mapping to `crm_v2.contacts`. Its mappers etc will decorate the entity with stuff you don't get from the table. `name` doesn't exist as a column in the table. In fact, we have an equivalent on the model (because it is more complicated than it first appears! https://github.com/DEFRA/water-abstraction-system/blob/main/app/models/contact.model.js#L78 )

When we've dug into the legacy code regards showing a 'contact' the general principle seems to be if `crm_v2.contact` is present it takes precedence. But if you check the database `last_name` is never null but `first_name` sometimes is. Hence not assuming it will be there and having a test for if it's not.

test/presenters/licences/licence-contacts.presenter.test.js
```diff
@@ -0,0 +1,119 @@
...
+// Thing under test
+const LicenceContactsPresenter = require('../../../app/presenters/licences/licence-contacts.presenter.js')
+
+describe('Licence Contacts presenter', () => {
+  let licenceContacts
+
+  beforeEach(() => {
+    licenceContacts = [
+      {
+        communicationType: 'Licence Holder',
```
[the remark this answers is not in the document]
Cruikshanks: I wouldn't know how to enforce/guide this using ESLint.

IMHO seeing the data where it is being used is better than always putting it at the bottom. But there are times when the amount of data we need to generate is _so_ large it starts masking/blocking the visibility of the tests.

Hence, sometimes we have a function at the bottom and sometimes we generate it at the point of use.

However, I'm also conscious this is completely at odds with my other of 'consistency'! 😩  So, happy for this suggestion to be ignored and maybe we need to have a discussion on the dev call as to what folks would prefer to do.

Cruikshanks:
Because @Cruikshanks is useless with git he has messed up completely the merge conflict in this branch!

So, he is bailing out and starting afresh over at https://github.com/DEFRA/water-abstraction-system/pull/1011.

Sorry folks, ignore this one!

PR #1006, opened 2024-05-12
Ensure use of block body for arrow functions

Cruikshanks:

Just like [parens in arrow functions](https://github.com/DEFRA/water-abstraction-system/pull/1004) how you write an [arrow function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) can differ.

```javascript
const materials = [
  'Hydrogen',
  'Helium',
  'Lithium',
  'Beryllium'
]

// Block body
materials.forEach((material) => {
  console.log(`${material} - ${material.length}`)
})

// Concise body
materials.forEach((material) => console.log(`${material} - ${material.length}`))
```

The _block-body_ version can be applied in all cases. But like with parens where devs will often drop the parens when there is only 1 param, they'll use _concise_ for simple arrow functions and _block-body_ for the rest.

To us though, the argument is the same. For those new or inexperienced with JavaScript, this inconsistency can be confusing and a blocker to progressing with the language. The more we can do to keep the code base simple and consistent, the easier new folks will find working with it.

So, this change enables [the rule in ESLint](https://eslint.org/docs/latest/rules/arrow-body-style) that requires all arrow functions to use the block-body.

---

The [ESLint stylistic plugin](https://eslint.style/) we use has a related rule; [https://eslint.style/rules/js/implicit-arrow-linebreak#when-not-to-use-it](https://eslint.style/rules/js/implicit-arrow-linebreak#when-not-to-use-it) which suggests it should not be used when `arrow-body-style` is set to 'always'. This is why it is disabled as part of this change.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,6 +13,7 @@ module.exports = {
     'import/extensions': ['error', 'always'],
     strict: ['error', 'global'],
     '@stylistic/js/arrow-parens': ['error', 'always'],
+    '@stylistic/js/implicit-arrow-linebreak': ['off'],
     '@stylistic/js/max-len': ['error', {
       code: 120,
       ignoreStrings: true,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,6 +9,7 @@ module.exports = {
     '@stylistic/js'
   ],
   rules: {
+    'arrow-body-style': ['error', 'always'],
     'import/extensions': ['error', 'always'],
     strict: ['error', 'global'],
     '@stylistic/js/arrow-parens': ['error', 'always'],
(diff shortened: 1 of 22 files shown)


Cruikshanks:
This will mean the [exception in our conventions](https://github.com/DEFRA/water-abstraction-team/blob/main/coding_conventions.md#%EF%B8%8F%E2%83%A3-the-exception) will have to go. But if consistency was the name of the game we probably shouldn't have had the exception in the first place 😬

jonathangoulding (approved):

robertparkinson (approved):

Jozzey (commented):
Looks good to me. You seem to have upset SonarCloud a bit though.

PR #1012, opened 2024-05-13
Add View Licence customer details

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4434

This change will enhance the contact details controller, service and presenter to handle customer contacts, introducing a customer contacts presenter and fetch service.

This will share the same view as the summary page and load the same 'common data' established in previous work.


app/presenters/licences/customer-contacts.presenter.js
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Formats data for the `/licences/{id}/contact-details` view customer contact details page
+ * @module CustomerContactDetailsPresenter
+ */
+
+const ContactModel = require('../../models/contact.model.js')
+
+/**
+ * Formats data for the `/licences/{id}/contact-details` view customer contact details page
+ *
```
Cruikshanks, line 12: NIT - Just the missing param

```suggestion
 * Formats data for the `/licences/{id}/contact-details` view customer contact details page
 *
 * @param {Object[]} customerContacts - The results from `FetchCustomerContactsService` to be formatted for the view
 * template
 *
 * @returns {Object} The data formatted for the view template
```

app/services/licences/fetch-customer-contacts.service.js
```diff
@@ -0,0 +1,50 @@
+'use strict'
+
+/**
+ * Fetches all return logs for a licence which is needed for the view '/licences/{id}/contact-details` page
```
Cruikshanks, line 4: NIT - copy and paste correction

```suggestion
 * Fetches all customer contacts for a licence which is needed for the view '/licences/{id}/contact-details` page
```

app/services/licences/fetch-customer-contacts.service.js
```diff
@@ -0,0 +1,50 @@
...
+ * @module FetchCustomerContactDetailsService
+ */
+
+const { db } = require('../../../db/db.js')
+
+/**
+ * Fetches all contact details for a licence which is needed for the view '/licences/{id}/contact-details` page
```
Cruikshanks, line 11: NIT - just being verbose about what we are fetching

```suggestion
 * Fetches all customer contacts for a licence which is needed for the view '/licences/{id}/contact-details` page
```

app/views/licences/tabs/contact-details.njk
```diff
@@ -4,14 +4,18 @@
   <p><a href='/customer/{{ customerId }}/#contacts'>Go to customer contacts</a></p>
 {% endif %}
 
+{% if not customerContacts.length > 0  and not licenceContacts.length > 0 %}
+  <p>No contacts found.</p>
+{% endif %}
```
Cruikshanks, line 9: This one is really a matter of opinion (so feel free to 🥊 me!) But I think it reads clearer as

```suggestion
{% if customerContacts.length == 0 and licenceContacts.length == 0 %}
  <p>No contacts found.</p>
{% endif %}
```

🤷

test/services/licences/view-licence-contact-details.service.test.js
```diff
@@ -37,6 +39,16 @@ describe('View Licence Contact Details service', () => {
       postcode: 'CF71 7DG',
       country: 'United Kingdom'
     }])
+    Sinon.stub(FetchCustomerContactDetailsService, 'go').returns([{
```
Cruikshanks, line 42: NIT - Love a bit of white space!

```suggestion

    Sinon.stub(FetchCustomerContactDetailsService, 'go').returns([{
```

test/services/licences/view-licence-contact-details.service.test.js
```diff
@@ -51,6 +63,12 @@ describe('View Licence Contact Details service', () => {
 
       expect(result).to.equal({
         activeTab: 'contact-details',
+        customerContacts: [{
+          communicationType: 'Additional Contact',
+          email: 'dfd@email.com',
+          name: 'Donald Duck'
+        }
+        ],
```
Cruikshanks, line 71: NIT - Just being consistent with the start of the declaration

```suggestion
        }],
```

Cruikshanks (changes requested):

app/views/licences/tabs/contact-details.njk
jonathangoulding: If it passes the test no probs.

Cruikshanks: I can see that this has caused the test to fail. However, I believe it's the test that needs fixing not the view.

Assuming I've followed and understood what is happening correctly; `ViewLicenceContactDetails` will _always_ return

```javascript
  // ...
  customerContacts: [],
  licenceContacts: []
}
```

It is just whether they are populated or not. However, when we stub `ViewLicenceContactDetailsService` in the "when a request is valid and has no contact details" test we are only returning `{ activeTab: 'contact-details' }` which doesn't reflect what would actually happen.

If we update the stub to be

```javascript
Sinon.stub(ViewLicenceContactDetailsService, 'go').resolves({
  activeTab: 'contact-details',
  customerContacts: [],
  licenceContacts: []
})
```

then the test will pass again.

jonathangoulding: soddy test that, poor craftmanship. 

have updated and fixed.

jonathangoulding: unresolved. 

@Cruikshanks resolve if you are happy.

Cruikshanks (approved):

PR #1013, opened 2024-05-13
Add CRM V2 Company contacts view

jonathangoulding:
Add the crm_v2.company_contacts view to the database.

Add a model for the company contacts with a relationship to the companies and conatcts table.

This is needed to for the current view licence contact details page. The customer and licence contact are used on this page.


robertparkinson (approved):

test/support/helpers/company-contact.helper.js
```diff
@@ -0,0 +1,56 @@
+'use strict'
+
+/**
+ * @module CompanyContactHelper
+ */
+
+const CompanyContactModel = require('../../../app/models/company-contact.model.js')
+const { generateUUID } = require('../../../app/lib/general.lib.js')
+/**
```
Cruikshanks, line 9: NIT - There had to be one to prove I looked at the thing!

```suggestion
const { generateUUID } = require('../../../app/lib/general.lib.js')

/**
```

test/support/helpers/company-contact.helper.js
```diff
@@ -0,0 +1,56 @@
...
+ * Returns the defaults used
+ *
+ * It will override or append to them any data provided. Mainly used by the `add()` method, we make it available
+ * for use in tests to avoid having to duplicate values.
+ *
+ * @param {Object} [data] Any data you want to use instead of the defaults used here or in the database
+ */
+function defaults (data = {}) {
+  const defaults = {
+    id: generateUUID(),
```
Cruikshanks, line 41: You don't need to default the `id:` if the legacy table automatically generates them. There are the odd few that don't hence we might have specified this in some helpers. But they are the exception rather than the norm.

```suggestion
```

N.B. This will also mean it can be dropped from the JSDoc description.

Cruikshanks (changes requested):
Had to prove I was looking!

test/support/helpers/company-contact.helper.js
```diff
@@ -0,0 +1,56 @@
...
+
+const CompanyContactModel = require('../../../app/models/company-contact.model.js')
+const { generateUUID } = require('../../../app/lib/general.lib.js')
+
+/**
+ * Add a new company contact
+ *
+ * If no `data` is provided, default values will be used. These are
+ *
+ * - `id` - [random UUID]
```
jonathangoulding, line 15:
```suggestion

```

Cruikshanks (approved):

PR #1014, opened 2024-05-13
Adding multiple return requirements to the session

rvsiyad:

When setting up 'requirements for returns', a user can set up multiple requirements (it's why we call it '_requirements_ for returns' not '_requirement_ for returns'!

Unfortunately, the wording for this feature isn't helpful to us dev's. So, when a user opts to setup a new 'requirements for returns' we actually create a _return version_. Against that we'll record the start date, reason, a note (optional) and any additional agreements.

Under that must be _at least_ 1 return requirement. Against that we record details like the purpose, point, site description etc.

But users need flexibility. For example, suppose a licence is granted to abstract water for 2 different purposes. We may require separate returns for each purpose. In this case, we only need 1 return version, for example, 'new licence'. But we need 2 separate return requirements in order to detail the 2 different return submissions we'll expect.

In [Add support for multiple rtn. reqs. plus tidy up](https://github.com/DEFRA/water-abstraction-system/pull/977) we updated the session handling and journey to support working with multiple return requirements.

This change is allows users to add and additional requirements to the session and set them up once they have reached the `/check` page in the journey. When a user clicks the add button, they'll be redirected to the `/purpose/{new requirements index}` page after we've inserted a new blank requirement in the session.

### DIFF
--- a/app/presenters/return-requirements/check.presenter.js
+++ b/app/presenters/return-requirements/check.presenter.js
@@ -40,6 +40,9 @@ function _requirements (session) {
 
   for (const requirement of requirements) {
     const { siteDescription, agreementsExceptions } = requirement
+
+    // NOTE: We determine a requirement is complete because agreement exceptions is populated and it is the last step in
+    // the journey
     if (agreementsExceptions) {
       completedRequirements.push({ siteDescription })
     }
--- a/app/presenters/return-requirements/check.presenter.js
+++ b/app/presenters/return-requirements/check.presenter.js
@@ -18,6 +18,7 @@ function go (session) {
     pageTitle: `Check the return requirements for ${licence.licenceHolder}`,
     reason: returnRequirementReasons[reason],
     reasonLink: _reasonLink(sessionId, journey),
+    requirements: _requirements(session),
     sessionId,
     startDate: _startDate(session),
     userEmail: note ? note.userEmail : 'No notes added'
@@ -30,6 +31,21 @@ function _reasonLink (sessionId, journey) {
   }
 
   return `/system/return-requirements/${sessionId}/no-returns-required`
+}
+
+function _requirements (session) {
+  const { requirements } = session
+
+  const completedRequirements = []
+
+  for (const requirement of requirements) {
+    const { siteDescription, agreementsExceptions } = requirement
+    if (agreementsExceptions) {
+      completedRequirements.push({ siteDescription })
+    }
+  }
+
+  return completedRequirements
 }
 
 function _startDate (session) {
--- a/app/presenters/return-requirements/purpose.presenter.js
+++ b/app/presenters/return-requirements/purpose.presenter.js
@@ -31,6 +31,15 @@ function go (session, requirementIndex, purposesData) {
 function _backLink (session) {
   const { checkPageVisited, id, requirements } = session
 
+  // NOTE: Purpose is the first page in the manual setup journey. So, when a user first comes through, we want to allow
+  // them to go back to `/setup`. Once they've got to the `/check` page they may return because they clicked the
+  // 'Change' link for the purpose. When this happens, `checkPageVisited` will be true and 'Back' needs to take them
+  // back there.
+  //
+  // But if they click 'Add requirement' on the `/check` page they'll also be directed here. When that happens
+  // `checkPageVisited` will have been reset to false to allow the user to progress through the setup journey. In this
+  // scenario 'Back' also needs to take them back to `/check`. Hence, the logic is different in this presenter when
+  // compared with the other setup pages.
   if (checkPageVisited || requirements.length > 1) {
     return `/system/return-requirements/${id}/check`
   }
--- a/app/presenters/return-requirements/purpose.presenter.js
+++ b/app/presenters/return-requirements/purpose.presenter.js
@@ -29,9 +29,9 @@ function go (session, requirementIndex, purposesData) {
 }
 
 function _backLink (session) {
-  const { checkPageVisited, id } = session
+  const { checkPageVisited, id, requirements } = session
 
-  if (checkPageVisited) {
+  if (checkPageVisited || requirements.length > 1) {
     return `/system/return-requirements/${id}/check`
   }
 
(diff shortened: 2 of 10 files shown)


app/presenters/return-requirements/check.presenter.js
```diff
@@ -32,6 +33,21 @@ function _reasonLink (sessionId, journey) {
...
 }
 
+function _requirements (session) {
+  const { requirements } = session
+
+  const completedRequirements = []
+
+  for (const requirement of requirements) {
+    const { siteDescription, agreementsExceptions } = requirement
+    if (agreementsExceptions) {
```
Cruikshanks, line 43: I'd suggest adding a note for context here.

```suggestion
    // NOTE: We determine a requirement is complete because agreement exceptions is populated and it is the last step in
    // the journey
    if (agreementsExceptions) {
```

Cruikshanks, line 43: Adding some white space

```suggestion
    const { siteDescription, agreementsExceptions } = requirement

    if (agreementsExceptions) {
```

app/presenters/return-requirements/purpose.presenter.js
```diff
@@ -29,9 +29,9 @@ function go (session, requirementIndex, purposesData) {
 }
 
 function _backLink (session) {
-  const { checkPageVisited, id } = session
+  const { checkPageVisited, id, requirements } = session
 
-  if (checkPageVisited) {
+  if (checkPageVisited || requirements.length > 1) {
```
Cruikshanks, line 34: I'd suggest adding a note for context here

```suggestion
  // NOTE: Purpose is the first page in the manual setup journey. So, when a user first comes through, we want to allow
  // them to go back to `/setup`. Once they've got to the `/check` page they may return because they clicked the
  // 'Change' link for the purpose. When this happens, `checkPageVisited` will be true and 'Back' needs to take them
  // back there.
  //
  // But if they click 'Add requirement' on the `/check` page they'll also be directed here. When that happens
  // `checkPageVisited` will have been reset to false to allow the user to progress through the setup journey. In this
  // scenario 'Back' also needs to take them back to `/check`. Hence, the logic is different in this presenter when
  // compared with the other setup pages.
  if (checkPageVisited || requirements.length > 1) {
```

app/routes/return-requirement.routes.js
```diff
@@ -29,6 +29,20 @@ const routes = [
...
+    method: 'POST',
+    path: '/return-requirements/{sessionId}/add',
+    handler: ReturnRequirementsController.add,
+    options: {
+      auth: {
+        access: {
+          scope: ['billing']
+        }
+      },
+      description: 'Submit the abstraction period for the return requirement'
```
Cruikshanks, line 42: Think you've got a copy & paste error here!

app/services/return-requirements/add.service.js
```diff
@@ -0,0 +1,36 @@
+'use strict'
+
+/**
+ * Orchestrates adding an empty object to requirements array to the setup session record.
```
Cruikshanks, line 4:
```suggestion
 * Orchestrates adding an empty object to the requirements array in the session
```

app/services/return-requirements/add.service.js
```diff
@@ -0,0 +1,36 @@
...
+const SessionModel = require('../../models/session.model.js')
+
+/**
+ * Orchestrates adding an empty object to requirements array to the setup session record.
+ *
+ * Supports adding another object to the requirements array in the session when a user hits the add requirements button.
+ * It also returns the index of the object added to add the new returns requirements data.
+ *
+ * @param {string} sessionId - The UUID of the current session
+ * @returns {number} - The index of the object in the requirements array
```
Cruikshanks, line 17:
```suggestion
 * Orchestrates adding an empty object to the requirements array in the session
 *
 * Supports adding another object to the requirements array in the session when a user hits the 'Add another requirement' button.
 *
 * @param {string} sessionId - The UUID of the current session
 *
 * @returns {number} - The index of new requirement. Needed by the setup pages so they know which requirement to display
 * and update
```

test/services/return-requirements/add.service.test.js
```diff
@@ -0,0 +1,52 @@
...
+  describe('when called', () => {
+    it('adds another empty object to the requirement array in the current setup session record', async () => {
+      await AddService.go(session.id)
+
+      const refreshedSession = await session.$query()
+
+      expect(refreshedSession.data.requirements.length).to.equal(2)
+      expect(refreshedSession.data.requirements).to.equal([{}, {}])
+    })
+  })
```
Cruikshanks, line 51: We're not testing that the service also returns the much needed requirement index. Would you mind adding something for this as well?

Cruikshanks (changes requested):

app/services/return-requirements/add.service.js
```diff
@@ -1,21 +1,22 @@
...
  *
- * Supports adding another object to the requirements array in the session when a user hits the add requirements button.
- * It also returns the index of the object added to add the new returns requirements data.
+ * Supports adding another object to the requirements array in the session when a user hits the 'Add another
+ * requirement' button.
  *
  * @param {string} sessionId - The UUID of the current session
- * @returns {number} - The index of the object in the requirements array
- */
+ * @returns {number} - The index of the new requirement. Needed by the setup pages so they know which requirement to
```
Cruikshanks, line 17: NIT - white space

```suggestion
 *
 * @returns {number} - The index of the new requirement. Needed by the setup pages so they know which requirement to
```

test/services/return-requirements/add.service.test.js
```diff
@@ -48,5 +48,11 @@ describe('Return Requirements - Add service', () => {
       expect(refreshedSession.data.requirements.length).to.equal(2)
       expect(refreshedSession.data.requirements).to.equal([{}, {}])
     })
+
+    it('returns the index of the the new requirement', async () => {
```
Cruikshanks, line 52: NIT - typo

```suggestion
    it('returns the index of the new requirement', async () => {
```

Cruikshanks (changes requested):

Cruikshanks (approved):

PR #1018, opened 2024-05-15
Add Scheduled Notifications model

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4315

> Part of work to replace the legacy view licence page

The 'Communications' tab on the view licence page needs to get the 'sent' scheduled notifications linked to a licence in order to render a list of communications.

The data in the legacy communications tab depends on `water.events` table data. Events are linked to scheduled notifications.

So, this change adds the new model and sets a link between it and the `EventModel`.


db/migrations/legacy/20240515104554_water-add-scheduled-notifications.js
Cruikshanks, line 1: We've been bitten by not adding a column in the legacy migration that we come to depend on later.

So, to avoid polluting our already noisy legacy migrations folder with `alter_` scripts we always try and create the legacy table 'as-is'.

We don't bother worrying about custom types or foreign key constraints (our patience only goes so far!) but we will add all columns and any other unique constraints they may have setup in addition to the ID.

test/models/scheduled-notification.model.test.js
```diff
@@ -0,0 +1,68 @@
...
+const Code = require('@hapi/code')
+
+const { describe, it, beforeEach } = exports.lab = Lab.script()
+const { expect } = Code
+
+// Test helpers
+const DatabaseSupport = require('../support/database.js')
+const EventHelper = require('../support/helpers/event.helper.js')
+const ScheduledNotificationHelper = require('../support/helpers/scheduled-notification.helper.js')
+const EventModel = require('../../app/models/event.model.js')
```
Cruikshanks, line 14: I'll get the ESLint plugin added soon I promise!

```suggestion
const EventHelper = require('../support/helpers/event.helper.js')
const EventModel = require('../../app/models/event.model.js')
const ScheduledNotificationHelper = require('../support/helpers/scheduled-notification.helper.js')
```

Cruikshanks (changes requested):
Just the notes regarding the legacy migration which obviously you could have done with before, not after starting this PR 😬

db/migrations/legacy/20240515104554_water-add-scheduled-notifications.js
jonathangoulding: Okay will get to it

Cruikshanks: My tip is most DB tools will have an option to generate the SQL to re-create the table (in DBeaver it is right click -> Generate SQL -> DDL). It should give you something like this that you can then use for reference.

```sql
-- water.scheduled_notification definition

-- Drop table

-- DROP TABLE water.scheduled_notification;

CREATE TABLE water.scheduled_notification (
	recipient varchar NULL,
	message_type varchar NULL,
	message_ref varchar NULL,
	personalisation jsonb NULL,
	send_after timestamp(0) NULL,
	status varchar NULL,
	log varchar NULL,
	licences jsonb NULL,
	individual_entity_id varchar NULL,
	company_entity_id varchar NULL,
	medium varchar NULL,
	notify_id varchar NULL,
	notify_status varchar NULL,
	plaintext varchar NULL,
	event_id uuid NULL,
	metadata jsonb NULL,
	status_checks int8 NULL,
	next_status_check timestamp(0) NULL,
	id uuid DEFAULT gen_random_uuid() NOT NULL,
	date_created timestamp NULL,
	notification_type numeric NULL,
	job_id varchar NULL,
	CONSTRAINT scheduled_notification_job_id_key UNIQUE (job_id),
	CONSTRAINT scheduled_notification_pkey PRIMARY KEY (id)
);
CREATE INDEX idx_scheduled_notification_statuses ON water.scheduled_notification USING btree (status, notify_status);
CREATE INDEX scheduled_notification_idx_send_after ON water.scheduled_notification USING btree (send_after);


-- water.scheduled_notification foreign keys

ALTER TABLE water.scheduled_notification ADD CONSTRAINT fk_scheduled_notification_event FOREIGN KEY (event_id) REFERENCES water.events(event_id);
```

jonathangoulding: Migration Updated 

@Cruikshanks i will leave to you to resolve if happy.

Cruikshanks (approved):

PR #1019, opened 2024-05-15
Check your requirements page (with requirements summary cards)

Demwunz:

This branch will add summary cards to the `check` page in the returns journey to display the customer's requirements.

Each box will reflect the requirements details (item, quantity etc) and link to the relevant requirement page.

This will provide a quick overview for the user on the `check` page.


app/views/return-requirements/check.njk
```diff
@@ -113,13 +113,160 @@
...
+                  key: {
+                    text: "Purpose"
+                  },
+                  value: {
+                    html: "requirement.purposes"
+                  },
+                  actions: {
+                    items: [
+                      {
+                        href: "purpose/" + rowIndex,
```
Cruikshanks, line 134: We can't use the row index as the requirement index here. When @rvsiyad looked at adding an additional requirement, he found it was perfectly possible for a user to start the journey, but then back all the way out without completing it.

That is why we have the concept of creating an array of `completedRequirements` in the presenter.

So, imagine you have create the first, started adding a second but then backed out (maybe you were unsure of something). Having confirmed your missing detail you then start adding a third which you complete.

In this page you'll just see 2 requirements listed. Left as is, when I click the 'Change' links for the second _display_ requirement, `/points/1` for example, I'll be shown the details for the second incomplete one when I follow it.

Or (!!) we could. But we'd need to add something to `CheckService` that clears out incomplete requirements from `session.requirements`, then save the update to the session. You could then drop leave this view as is and drop the checks for completed requirements in the presenter.

🤷 I'm fine either way!

Demwunz: @Cruikshanks amended, works as expected

app/views/return-requirements/check.njk
```diff
@@ -108,18 +108,164 @@
           preventDoubleClick: true
         }) }}
       </form>
-
       {% for requirement in requirements %}
         {# Set an easier to use index #}
-        {% set rowIndex = loop.index0 %}
+        {% set rowIndex = requirement.index %}
```
Cruikshanks, line 113: This will defo solve our problem! The only thing I'd question is, why not just add `requirement.index` to the end of the links instead of `rowIndex`? You could then do away with `rowIndex` altogether.

Cruikshanks (dismissed):
Got a slight tweak I think we can make. But also, we should probably have some test coverage for how requirements now get presented based on what we've added to the presenter.

app/views/return-requirements/check.njk
Demwunz: I only did it for legibility, happy with either way

robertparkinson (approved):

robertparkinson (approved):

jonathangoulding:
I have pushed the changes for - https://github.com/DEFRA/water-abstraction-system/pull/1019#pullrequestreview-2074415330

I have added a map function for the requirement to extract all fields needed for the view. 

I will work on the other requirment fields in the session object / view in another PR. It is assumed we need to do some fetch requests to get the information for the view (requirement.purposes) so it makes sense to do that in a separate PR.

robertparkinson (approved):

PR #1025, opened 2024-05-16
Refactor purposes page to use purpose ids as values

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4294

As part of our work to get the requirements for returns set up we need to know which of the purposes the user has selected in the manual journey so that we can then get the information later at the check your answers page.

This PR changes the logic to store the ID rather than the string which describes the purpose. that way we can look it up again.


app/services/return-requirements/submit-purpose.service.js
```diff
@@ -69,8 +69,13 @@ async function _save (session, requirementIndex, payload) {
   return session.$update()
 }
 
-function _validate (payload) {
-  const validation = PurposeValidation.go(payload)
+async function _validate (payload, purposesData) {
+  const purposeIds = []
+  purposesData.forEach((purpose) => {
+    purposeIds.push(purpose.id)
+  })
```
Cruikshanks, line 76: Sorry! Small NIT.

I think this scenario is what `map()` is idiomatically suited for.

```suggestion
  const purposeIds = purposesData.map((purpose) => {
    return purpose.id
  })
```

test/services/return-requirements/fetch-licence-purposes.service.test.js
```diff
@@ -14,9 +14,9 @@ const LicenceVersionPurposeHelper = require('../../support/helpers/licence-versi
 const PurposeHelper = require('../../support/helpers/purpose.helper.js')
 
 // Thing under test
-const FetchPurposesService = require('../../../app/services/return-requirements/fetch-purposes.service.js')
+const FetchPurposesService = require('../../../app/services/return-requirements/fetch-licence-purposes.service.js')
 
-describe('Return Requirements - Fetch Purposes service', () => {
+describe('Return Requirements - Fetch Purposes for a given licence service', () => {
```
Cruikshanks, line 19: Can't disagree the title is more meaningful. But it breaks our convention of using the module name.

```suggestion
describe('Return Requirements - Fetch Licence Purposes service', () => {
```

test/services/return-requirements/fetch-licence-purposes.service.test.js
```diff
@@ -14,9 +14,9 @@ const LicenceVersionPurposeHelper = require('../../support/helpers/licence-versi
 const PurposeHelper = require('../../support/helpers/purpose.helper.js')
 
 // Thing under test
-const FetchPurposesService = require('../../../app/services/return-requirements/fetch-purposes.service.js')
+const FetchPurposesService = require('../../../app/services/return-requirements/fetch-licence-purposes.service.js')
```
Cruikshanks, line 17: In VSCode, highlight the word, hit `fn+F2` then type `FetchLicencePurposes` and watch the magic happen!

```suggestion
const FetchLicencePurposesService = require('../../../app/services/return-requirements/fetch-licence-purposes.service.js')
```

test/services/return-requirements/purpose.service.test.js
Cruikshanks, line 1: Before wading in, is there a reason _not_ to continue stubbing the fetch service? We've switched to injecting real records instead.

We'll have tested `FetchLicencePurposesService` elsewhere, so I don't think we are at risk by not calling it for real within this test. 🤷

test/services/return-requirements/submit-purpose.service.test.js
Cruikshanks, line 1: Same question about stubbing VS real data applies here.

Cruikshanks (changes requested):
I had 1 NIT about perhaps using `map()` instead of the `forEach()`.

My key question though is about replacing stubbing with injecting real records in the tests. I'm not sure it is necessary but happy to be proved wrong!

test/services/return-requirements/purpose.service.test.js
robertparkinson: I guss I can revert it now. I changed it because I was doing the additional call to get all the purposeIds and I wanted to make sure it was calling it correctly

Cruikshanks (approved):

PR #1032, opened 2024-05-17
Fix typo with remove button in remove requirements

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4304

Noticed a typo with the button on the remove page of the return requirements
journey. Button should be labelled "Remove" instead of "Confirm Cancel".

This PR is focused on fixing that typo.

### DIFF
--- a/app/views/return-requirements/remove.njk
+++ b/app/views/return-requirements/remove.njk
@@ -24,7 +24,7 @@
       <p class="govuk-body govuk-!-font-weight-bold govuk-!-font-size-27"> {{ returnRequirement }}</p>
 
       <p class="govuk-!-margin-bottom-7"></p>
-      {{ govukButton({ text: "Confirm cancel" }) }}
+      {{ govukButton({ text: "Remove" }) }}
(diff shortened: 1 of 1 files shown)


app/views/return-requirements/remove.njk
```diff
@@ -24,7 +24,7 @@
       <p class="govuk-body govuk-!-font-weight-bold govuk-!-font-size-27"> {{ returnRequirement }}</p>
 
       <p class="govuk-!-margin-bottom-7"></p>
-      {{ govukButton({ text: "Confirm cancel" }) }}
+      {{ govukButton({ text: "Remove" }) }}
```
Cruikshanks, line 27: Sorry, if you are making this change could you also apply the prevent double click attribute at the same time?

```suggestion
      {{ govukButton({ text: "Remove", preventDoubleClick: true }) }}
```

Cruikshanks (changes requested):

Cruikshanks (approved):

PR #1034, opened 2024-05-20
View Licence set up page

jonathangoulding:
This page replaces the existing charge information page.

The page will show charge information for a licence and allow users to create a new charge version to manage licence billing.

Only restricted users can use this page and only some of those users can create a new charge and make the licence non-chargeable.

PR #1047, opened 2024-05-23
View licence set up agreements

jonathangoulding:
This will add an agreement table and buttons to the view licence set up page.

The agreements action are based on the user role and state of an agreement.

An authorised user can delete/end/recalculate an agreement as well as set up a new agreement

This change is part of the charge information page replacement seen in this previous work - https://github.com/DEFRA/water-abstraction-system/pull/1034

### DIFF
--- a/app/views/licences/tabs/set-up.njk
+++ b/app/views/licences/tabs/set-up.njk
@@ -2,9 +2,17 @@
 {% from "govuk/components/table/macro.njk" import govukTable %}
 {% from "macros/charge-version-status-tag.njk" import statusTag %}
 
-{% set rows = [] %}
+{% macro createLink(chargeVersion) %}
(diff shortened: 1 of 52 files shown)


app/views/licences/tabs/set-up.njk
```diff
@@ -80,3 +80,68 @@
...
+  },
+  {
+    text: "Date signed"
+  },
+  {
+    text: "Action"
+  }
+] %}
+
+{% for agreement in  agreements %}
```
robertparkinson, line 106: can you move this logic into the presenter please? and just pass the array to the nunucks file?

robertparkinson (changes requested):

robertparkinson (approved):

PR #1054, opened 2024-05-28
Display return versions in view licence setup tab

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4407

> Part of the work to replace NALD for handling return requirements

This adds the **Requirements for returns** table (return versions) to the 'Licence set up' tab on our new view licence page.

It also adds the buttons to create new ones, though these are placed behind a feature flag as we are not yet ready for the switchover.

This builds on the work done in [View Licence set up page](https://github.com/DEFRA/water-abstraction-system/pull/1034).


app/presenters/licences/set-up.presenter.js
```diff
@@ -28,17 +29,25 @@ const agreementDescriptions = {
...
  * @returns {Object} The data formatted for the view template
  */
-function go (chargeVersions, workflows, agreements, auth, commonData) {
+function go (chargeVersions, workflows, agreements, returnVersions, auth, commonData) {
   return {
+    links: {
+      chargeInformation: _chargeInformationLinks(auth, commonData),
+      agreements: _agreementLinks(auth, commonData),
+      returnsRequirements: {
+        returnsRequired: `/system/licences/${commonData.licenceId}/returns-required`,
```
jonathangoulding, line 44: This links in the old system go to here but end up redirected to ‘Select the start date for the requirements for returns’. 

But the ticket just says to get to this page.

app/presenters/licences/set-up.presenter.js
```diff
@@ -161,6 +170,21 @@ function _financialAgreementCode (agreement) {
   return agreement.financialAgreements[0].financialAgreementCode
 }
 
+function _returnsRequirements (returnsRequirements = [{}]) {
+  return returnsRequirements.map((returnsRequirement) => {
+    return {
+      action: [{
+        text: 'View',
+        link: ''
```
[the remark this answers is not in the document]
jonathangoulding: Ticket states go to - ‘View Requirements for returns’

app/views/licences/tabs/set-up.njk
```diff
@@ -10,135 +10,215 @@
...
-  {
-    text: "Action"
-  }
-] %}
-
-{% for chargeVersion in  chargeInformation %}
-  {% set chargeVersionRows = (chargeVersionRows.push([
+<h2 class="govuk-heading-l">Licence set up</h2>
+
+{% if enableRequirementsForReturns %}
```
Cruikshanks, line 15: It's only the buttons we need to put behind the feature flag now. The work on cleaning up the requirements during import is now done enough that it is fine to display them to the users. In fact, the sooner we can the sooner they can confirm if they are as expected.

app/views/licences/tabs/set-up.njk
```diff
@@ -10,135 +10,215 @@
...
+      },
+      {
+        text: "Status"
+      },
+      {
+        text: "Action"
+      }
+    ] %}
+
+    {% for returnsRequirement in  returnVersions %}
```
Cruikshanks, line 43:
```suggestion
    {% for returnsRequirement in returnVersions %}
```

app/views/licences/tabs/set-up.njk
```diff
@@ -10,135 +10,215 @@
...
+      {
+        text: "Reason"
+      },
+      {
+        text: "Status"
+      },
+      {
+        text: "Action"
+      }
+    ] %}
```
Cruikshanks, line 41: Do we need to set these separately? Or could we just include them when we define the table below?

(And yes, I've just clocked the other tables were built this way and I approved the previous PRs 🤦 . But I don't think they have to be. IMHO I think it's simpler to just define them when declaring the table.)

Cruikshanks (changes requested):
Love the changes to make working with the links easier. And some of the stuff I've highlighted isn't in the ticket and wouldn't be known by you.

But anything else feel free to ping me if you need to discuss them.

app/views/licences/tabs/set-up.njk
jonathangoulding: Okay will change. 

Will update the tests

app/presenters/licences/set-up.presenter.js
```diff
@@ -161,6 +170,21 @@ function _financialAgreementCode (agreement) {
...
 
+function _returnsVersions (returnVersions = [{}]) {
+  return returnVersions.map((returnVersion) => {
+    return {
+      action: [{
+        text: 'View',
+        link: ''
+      }],
+      endDate: returnVersion.endDate ? formatLongDate(returnVersion.endDate) : '',
+      reason: returnRequirementReasons[returnVersion.reason],
```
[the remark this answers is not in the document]
jonathangoulding: Yes i agree.

app/presenters/licences/set-up.presenter.js
```diff
@@ -161,6 +170,21 @@ function _financialAgreementCode (agreement) {
...
+      startDate: returnVersion.startDate ? formatLongDate(returnVersion.startDate) : '',
+      status: returnVersion.status
```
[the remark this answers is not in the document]
jonathangoulding: Done locally. Tests updated.

app/presenters/licences/set-up.presenter.js
```diff
@@ -161,6 +170,21 @@ function _financialAgreementCode (agreement) {
   return agreement.financialAgreements[0].financialAgreementCode
 }
 
+function _returnsVersions (returnVersions = [{}]) {
```
Cruikshanks, line 173: NIT - we call the property we're setting `returnVersion` and we pass in `returnVersions`. Sorry, should have spotted on first pass.

```suggestion
function _returnVersions (returnVersions = [{}]) {
```

app/views/licences/tabs/set-up.njk
```diff
@@ -10,135 +10,207 @@
...
+          text: "Status"
+        },
+        {
+          text: "Action"
+        }
+      ],
+        rows: returnVersionsTableRows
+      }) }}
+  {% endif %}
+{% if enableRequirementsForReturns %}
```
Cruikshanks, line 65: Rather than have this check here, why not reference the feature flag in `SetUpPresenter`? You can then determine whether the link gets added to the page data there.

This simplifies the view even further and means you don't have to test for this using a controller test.

Cruikshanks (changes requested):
One NIT but also a suggested change to where we refer to the feature flag and use it which I think is worth making.

app/presenters/licences/set-up.presenter.js
```diff
@@ -28,29 +28,38 @@ const agreementDescriptions = {
...
+ * @param {module:LicenceAgreementModel[]} agreements - All agreements records for the licence
  * @param {module:ReturnVersionModel[]} returnVersions - All returns version records for the licence
  * @param {Object} auth - The auth object taken from `request.auth` containing user details
  * @param {Object} commonData - Licence data already formatted for the view's shared elements
+ * @param {boolean} enableRequirementsForReturns - feature toggle for the return versions links
  *
  * @returns {Object} The data formatted for the view template
  */
-function go (chargeVersions, workflows, agreements, returnVersions, auth, commonData) {
+function go (chargeVersions, workflows, agreements, returnVersions, auth, commonData, enableRequirementsForReturns) {
```
Cruikshanks, line 39: We're currently grabbing the feature toggle in `view-licence-set-up.service.js` only to pass it through to the presenter as a param.

We can just grab the feature toggle here in the presenter and not have to change the args.

app/services/licences/fetch-return-versions.service.js
```diff
@@ -0,0 +1,38 @@
...
+    .select([
+      'id',
+      'startDate',
+      'endDate',
+      'status',
+      'reason'
+    ])
+    .orderBy([
+      { column: 'startDate', order: 'desc' }
+    ])
```
Cruikshanks, line 33: NIT - If we can we try and chain the Objection.js calls as if we were writing the SQL itself

```suggestion
  return ReturnVersionModel.query()
    .select([
      'id',
      'startDate',
      'endDate',
      'status',
      'reason'
    ])
    .where('licenceId', licenceId)
    .orderBy([
      { column: 'startDate', order: 'desc' }
    ])
```

app/services/licences/view-licence-set-up.service.js
```diff
@@ -5,8 +5,10 @@
  * @module ViewLicenceSetUpService
  */
 
+const FeatureFlagsConfig = require('../../../config/feature-flags.config.js')
```
Cruikshanks, line 8: See my comment in the presenter; we can grab this there and save passing it through as a param.

app/views/licences/tabs/set-up.njk
```diff
@@ -10,135 +10,206 @@
...
-
 <h2 class="govuk-heading-l">Licence set up</h2>
 
+  <h3 class="govuk-heading-m"> Requirements for returns </h3>
+
+  {% if returnVersions.length == 0 %}
+    <p class="govuk-body"> No requirements for returns for this licence.</p>
+  {% endif %}
+
+  {% if returnVersions.length > 0 %}
```
Cruikshanks, line 21: You know what, just to be different I'm cool with an `else` here!

 ```suggestion
  {% else %}
```

And yes I should have spotted this when you added the other tables 🤦 😳

File: test/controllers/licences.controller.test.js [hunk 14]
```diff
@@ -355,12 +362,36 @@ describe('Licences controller', () => {
...
+        expect(response.payload).to.contain('Requirements for returns')
+        expect(response.payload).to.contain('No requirements for returns for this licence.')
         expect(response.payload).to.contain('Charge information')
         expect(response.payload).to.contain('No charge information for this licence.')
         expect(response.payload).to.contain('Agreements')
         expect(response.payload).to.contain('No agreements for this licence.')
       })
     })
+
+    describe('when a request is valid but the requirements for returns toggle is false', () => {
```
Cruikshanks, line 374: We can drop this test. It is covered by your presenter test

Cruikshanks (changes requested):

app/views/licences/tabs/set-up.njk
jonathangoulding: crickey i missed that on all of them.

app/presenters/licences/set-up.presenter.js
jonathangoulding: D'oh!

test/presenters/licences/set-up.presenter.test.js
```diff
@@ -3,10 +3,14 @@
 // Test framework dependencies
 const Lab = require('@hapi/lab')
 const Code = require('@hapi/code')
+const Sinon = require('sinon')
```
Cruikshanks, line 6: If we've added Sinon to the test we also need to add an `afterEach()` to restore things. We've found tests can be flakey when we forget to do this.

```javascript
  afterEach(() => {
    Sinon.restore()
  })
```

In this case it would be placed after the first `beforeEach()` around line 87.

Cruikshanks (changes requested):

Cruikshanks (approved):

PR #1055, opened 2024-05-29
Fix licence set up charge information buttons

jonathangoulding:
When a licence is more than 6 years the charge information buttons should not show.

The endDate key on the licence was not a date but an object. This changes fixes the  _endsSixYearsAgo function by checking the date key on the endDate object.


Jozzey (approved):
👍🏼

PR #1063, opened 2024-05-31
Update two-part tariff review views to add Cypress data test attributes

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4450

We've finished building all the pages for the two-part tariff review and are now working on the acceptance tests. While writing these tests, we found a few places where the data test attribute on the view needs to be updated to make it clearer and easier to find in the acceptance tests file.


app/views/bill-runs/review-licence.njk
```diff
@@ -83,23 +83,23 @@
   <form method="post">
     <div class="govuk-button-group">
     {{ govukButton({
-      text: statusButtonText,
+      html: '<div data-test="status-button">' + statusButtonText + '</div>',
```
Jozzey, line 86: The button component allows you to add an `attributes` option which is where the `data-test` thingy should go.

app/views/bill-runs/review-licence.njk
```diff
@@ -83,23 +83,23 @@
...
       classes: statusButtonClass,
       name: "licence-status",
       value: statusButtonValue,
       preventDoubleClick: true
     }) }}
 
     {{ govukButton({
-      text: progressButtonText,
+      html: '<div data-test="progress-button">' + progressButtonText + '</div>',
```
Jozzey, line 94: The button component allows you to add an `attributes` option which is where the `data-test` thingy should go.

app/views/bill-runs/review-licence.njk
```diff
@@ -83,23 +83,23 @@
...
       classes: "govuk-button--secondary",
       name: "mark-progress",
       value: progressButtonValue,
       preventDoubleClick: true
     }) }}
 
     {{ govukButton({
-      text: "Remove from bill run",
+      html: '<div data-test="progress-button">Remove from bill run</div>',
```
Jozzey, line 102: The button component allows you to add an `attributes` option which is where the `data-test` thingy should go.

Jozzey (changes requested):

app/views/bill-runs/review-licence.njk
Cruikshanks: You don't need `data-test=` attributes for buttons! It is fine to use `text:` for those.

Checking out the [docs](https://docs.cypress.io/guides/references/best-practices#How-It-Works) this would mean we sit in ✅ **Depends** . But the button text so rarely changes it would be better to keep the views cleaner by just using the `text:` option.

Where `data-test=` really helps us is in summary lists and tables. Buttons, links and headings are normally very easy to select with cypress and also, very rarely change.

Jozzey (approved):

PR #1068, opened 2024-06-04
Feature returns required points

jonathangoulding:
Due to the estimated complexity of adding fetch services for purposes, points and other display items. We have broken down the complex display items into their own branches of work.

This branch will add points to the return requirement summary card. 

Previous work done here - https://github.com/DEFRA/water-abstraction-system/pull/1019

### DIFF
--- a/app/presenters/return-requirements/check/returns-requirements.presenter.js
+++ b/app/presenters/return-requirements/check/returns-requirements.presenter.js
@@ -77,9 +77,6 @@ function _mapRequirement (requirement, index, purposes, points) {
 }
 
 function _mapPoints (requirementPoints, points) {
-  if (!requirementPoints || !points) {
-    return []
-  }
 
   return requirementPoints.map((point) => {
     const matchedPoint = points.find((pid) => { return pid.ID === point })
--- a/app/presenters/return-requirements/check/returns-requirements.presenter.js
+++ b/app/presenters/return-requirements/check/returns-requirements.presenter.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const { formatAbstractionDate } = require('../../base.presenter.js')
+const { generateAbstractionPointDetail } = require('../../../lib/general.lib.js')
 
 /**
  * Formats return requirements data for the `/return-requirements/{sessionId}/check` page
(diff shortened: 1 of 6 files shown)


test/services/return-requirements/check.service.test.js
```diff
@@ -20,6 +21,11 @@ describe('Return Requirements - Check service', () => {
   let yarStub
 
   beforeEach(async () => {
+    Sinon.stub(ReturnRequirementsService, 'go').resolves({
```
jonathangoulding, line 24: This feels necessary as i would have to create a licence and permitLicence for this test. 

I think it would be doing too much so have mocked it. 

The ReturnRequirementsService is tested using the database helpers so the output is covered.

app/presenters/return-requirements/check/returns-requirements.presenter.js
```diff
@@ -56,17 +64,30 @@ function _mapPurposes (requirementPurposes, purposes) {
...
+    points: _mapPoints(requirement.points, points),
     purposes: _mapPurposes(requirement.purposes, purposes),
     siteDescription: requirement.siteDescription
   }
 }
 
+function _mapPoints (requirementPoints, points) {
+  if (!requirementPoints || !points) {
+    return []
+  }
```
Cruikshanks, line 82: I don't think we need to check for this.

If the licence has no points the user wouldn't have made it past the `/points` page, which also means the return requirement is incomplete.

Assuming the licence has points, again the user wouldn't have got past this page without having selected at least one.

So, we only call `_mapRequirement()` for completed return requirements, which means it is okay for the code to assume both `requirementPoints` and `points` are populated.

```suggestion
```

app/services/return-requirements/check/returns-requirements.service.js
```diff
@@ -16,12 +17,13 @@ const ReturnRequirementsPresenter = require('../../../presenters/return-requirem
  * @returns {Promise<Object>} page data needed by the view template
  */
 async function go (session) {
-  const { requirements, journey } = session.data
+  const { licence, data: { requirements, journey } } = session
```
Cruikshanks, line 20: Sorry, I didn't spot on the previous PR that we were calling `session.data`.

Both `requirements` and `licence` should exist at the top level of session. So, you should be able to just do

```suggestion
  const { licence, journey, requirements } = session
```

test/services/return-requirements/check.service.test.js
Cruikshanks: 💯 agree!

test/services/return-requirements/check/returns-requirements.service.test.js
```diff
@@ -37,7 +47,7 @@ describe('Return Requirements service', () => {
       frequencyCollected: 'daily',
       frequencyReported: 'daily',
       points: [
-        '286'
+        '9000031' // Hard coded from teh permit licence data
```
Cruikshanks, line 50: NIT - typo

```suggestion
        '9000031' // Hard coded from the permit licence data
```

Cruikshanks (changes requested):

Cruikshanks (approved):

PR #1070, opened 2024-06-04
Adding tags to view for acceptance tests (returns)

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4485

When writing acceptance tests, due to the way [crypress](https://docs.cypress.io/guides/overview/why-cypress) works, some components of a page can be hard to locate.

For example, imagine a summary list which has nested values. This might be the address on the page provided by cypress:

```
':nth-child(2) > .govuk-summary-list.govuk-\!-margin-bottom-2 > :nth-child(1) > .govuk-summary-list__value'
```

To make it easier to locate the elements on the page we want to test, we add tags like so:
```
'<span data-test="meta-data-reason">' + reason + '</span>'
```

This PR is focused on adding tags to the views in the returns required journey as well as adding more points to the permit licence helper to provide 2 points for the return requirement points page.


app/views/return-requirements/check.njk
Cruikshanks, line 1: It can be a little 'grey' when it comes to adding the data-test attribute.

The intent is to avoid making it difficult to select stuff in Cypress. Their best practice docs will advocate putting it on nearly everything. However, this can make some of our views more complex than they need to be.

For example, I do not believe we need to add it to buttons. My experience using Cypress is that these are very easy to find. This specific issue came up in a PR by @Beckyrose200 which [I gave a more more detail in](https://github.com/DEFRA/water-abstraction-system/pull/1063#discussion_r1627499292).

So on this page, you are going to need it for the remove requirement button because you might have more than one. But the `Approve` button should be fine without it. The same might go for links where there is only the one, for example, change reason or change start date (but shout if they are actually a pain!)

Where I expected to see the tags added was to the return requirement summary cards so you can select the elements and assert the values are what you expect. I can see we have done the 'change' links. But now you'd still have a problem if you have a scenario where there is more than 1 requirement. Cypress will return 2 elements if you search using `'meta-data-change-purpose'` for example.

Ideally, you need to try and make each element unique. We've done this in [other places](https://github.com/DEFRA/water-abstraction-system/blob/main/app/views/bill-runs/index.njk#L57) by using the index from the loop combined with the data-test attribute.

I'd see if you can make use of the same thing to make the elements of the requirement summary cards uniquely identifiable. Also, double check if you really need them for the one off elements like buttons and links.

Cruikshanks (changes requested):
Give me a ping if my feedback doesn't make sense or you need some help.

Cruikshanks (approved):

PR #1075, opened 2024-06-05
Add the returns cycle text to the returns requirements check page

jonathangoulding:
Due to the estimated complexity of adding fetch services for purposes, points and other display items. We have broken down the complex display items into their own branches of work.

This branch will add returns cycles to the return requirement summary card. As of this commit there is only 'summer' and 'winter and all year' returns cycles

Previous work done here - #1019

### DIFF
--- a/app/presenters/return-requirements/check/returns-requirements.presenter.js
+++ b/app/presenters/return-requirements/check/returns-requirements.presenter.js
@@ -54,18 +54,6 @@ function _requirements (requirements, purposes, points) {
   return completedRequirements
 }
 
-function _returnsCycle (returnsCycle) {
-  if (returnsCycle === 'summer') {
-    return 'Summer'
-  }
-
-  if (returnsCycle === 'winter-and-all-year') {
-    return 'Winter and all year'
-  }
-
-  return ''
-}
-
 function _mapPurposes (requirementPurposes, purposes) {
   return requirementPurposes.map((requirementPurpose) => {
     const matchedPurpose = purposes.find((purpose) => {
--- a/app/presenters/return-requirements/check/returns-requirements.presenter.js
+++ b/app/presenters/return-requirements/check/returns-requirements.presenter.js
@@ -84,7 +84,7 @@ function _mapRequirement (requirement, index, purposes, points) {
     index,
     points: _mapPoints(requirement.points, points),
     purposes: _mapPurposes(requirement.purposes, purposes),
-    returnsCycle: _returnsCycle(requirement.returnsCycle),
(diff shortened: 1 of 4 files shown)


jonathangoulding:
Not sure why sonar is saying the if condition is not tested for the winter all year. But it is tested  in the presenter tests

app/presenters/return-requirements/check/returns-requirements.presenter.js
```diff
@@ -72,12 +84,12 @@ function _mapRequirement (requirement, index, purposes, points) {
     index,
     points: _mapPoints(requirement.points, points),
     purposes: _mapPurposes(requirement.purposes, purposes),
+    returnsCycle: _returnsCycle(requirement.returnsCycle),
```
Cruikshanks, line 87: The function currently caters for the cycle not being recognised. This is a case of being 99.99999999999999999999% sure there will only ever being 2, so if it's not one it must be the other.

On that basis I think we can reduce this down to a one-liner (which is also how we've dealt with this elsewhere! 😬 )

```suggestion
    returnsCycle: requirement.returnsCycle === 'summer' ? 'Summer' : 'Winter and all year',
```

test/presenters/return-requirements/check/returns-requirements.presenter.test.js
```diff
@@ -117,6 +119,26 @@ describe('Return Requirements presenter', () => {
...
+            const result = ReturnRequirementsPresenter.go(requirements, purposes, points, journey)
+
+            expect(result.requirements[0].returnsCycle).to.equal('Summer')
+          })
+        })
+        describe('Winter and all year', () => {
+          beforeEach(() => {
+            requirements = [{ ...requirement, returnsCycle: 'winter-and-all-year' }]
+          })
+          it('should return the text for a Winter and all year return cycle', () => {
```
Cruikshanks, line 135: NIT - whitespace

```suggestion
        })

        describe('Winter and all year', () => {
          beforeEach(() => {
            requirements = [{ ...requirement, returnsCycle: 'winter-and-all-year' }]
          })

          it('should return the text for a Winter and all year return cycle', () => {
```

Cruikshanks (changes requested):

app/presenters/return-requirements/check/returns-requirements.presenter.js
jonathangoulding: Cool. I was very tempted to just do this  !

Cruikshanks (approved):

PR #1079, opened 2024-06-06
Add flash notification when changes occur on returns Requirements

jonathangoulding:
When something changes we want to show a flash notification on the check page.This will be:

title = Updated
text = Changes made

When a new requirement has been added the notification should be:

title = Added
text = New requirement added

These will disappear oif the page is refresh or navigated away from.


app/controllers/return-requirements.controller.js
```diff
@@ -43,6 +43,13 @@ const SubmitSetupService = require('../services/return-requirements/submit-setup
 const SubmitSiteDescriptionService = require('../services/return-requirements/submit-site-description.service.js')
 const SubmitStartDateService = require('../services/return-requirements/submit-start-date.service.js')
 
+function flasher (yar, title = 'Updated', text = 'Changes made') {
```
jonathangoulding: should this be a local function or moved out.

app/controllers/return-requirements.controller.js
```diff
@@ -274,6 +283,10 @@ async function submitAgreementsExceptions (request, h) {
     return h.view('return-requirements/agreements-exceptions.njk', pageData)
   }
 
+  // This needs logic
+  flasher(request.yar)
+  flasher(request.yar, 'Added', 'New requirement added')
```
jonathangoulding: Need to work out the logic how this could know it's a new requirement - check session or add a new key/value flag.

app/lib/flash-notifications.lib.js
Cruikshanks, line 1: I know this is still in draft, but just a heads up. Could we move this function to `general.lib.js`? I'm not sure it warrants its own module unless there was other 'flash'-related functionality we needed to add.

jonathangoulding: done

jonathangoulding:
Screenshot 2024-06-07 at 15 54 55

The banner was in place before this code and the text is bigger/bolder than required on the ticket - https://eaflood.atlassian.net/jira/software/c/projects/WATER/boards/96?quickFilter=347&selectedIssue=WATER-4386

I have not updated the styling as i assume the previous use passed test and PO review.

Cruikshanks (approved):

PR #1080, opened 2024-06-06
Remove blipp plugin

Jozzey:
https://github.com/DEFRA/water-abstraction-team/issues/111

We and it seems most other Defra services add this as standard to projects. But we've been (rightly) challenged on what value this adds. As the project progresses the number of routes is going to expand so that routing table is going to get very noisy. It's an extra dependency and we even have untested logic in `server.js` to only add the plugin when in non-prod environments.

Keeping to our ethos of minimal and simple, why do we have this dependency?!

Having discussed it no one could make a case for keeping it. So, we think it is time to say goodbye to **Blipp** 😢👋


Cruikshanks (changes requested):
One of the reasons we were challenged was the overhead of trying to ensure each route has a meaningful `description:` property.

This is only used by Blipp. If Blipp is gone, you'll never see the route description.

Can you see where I'm going? 😜 

If we're doing this, we need to go through all the route files as well and remove the `description:` properties.

Sorry mate! 😬

Jozzey:
> One of the reasons we were challenged was the overhead of trying to ensure each route has a meaningful `description:` property.
> 
> This is only used by Blipp. If Blipp is gone, you'll never see the route description.
> 
> Can you see where I'm going? 😜
> 
> If we're doing this, we need to go through all the route files as well and remove the `description:` properties.
> 
> Sorry mate! 😬

No worries, I wasn't aware of that. It will give me something nice and easy to do on a Friday.

Jozzey:
> One of the reasons we were challenged was the overhead of trying to ensure each route has a meaningful `description:` property.
> 
> This is only used by Blipp. If Blipp is gone, you'll never see the route description.
> 
> Can you see where I'm going? 😜
> 
> If we're doing this, we need to go through all the route files as well and remove the `description:` properties.
> 
> Sorry mate! 😬

@Cruikshanks The description properties are quite useful for quickly being able to see what a route is supposed to be doing. I find them useful anyway. If we are going to remove the `description` from the route I think it would be useful to still have the description text but maybe move it into a comment. What do you think?

Cruikshanks (approved):

PR #1085, opened 2024-06-10
Persist the returns requirements

jonathangoulding:


app/services/return-requirements/submit-check.service.js
```diff
@@ -28,6 +29,8 @@ async function go (sessionId) {
 
   await _validateLicence(session.licence.id)
 
+  await PersistReturnRequirementsService.go(session)
```
jonathangoulding, line 32: We will need to guard against the user navigating to the previous check screen. As currently, if they go back and approve again, this will be a double insertion (a new record for each return requirement). 

We can add the id's in the session and do an update / upsert. I am not aware of the current practices in the system to know which way to go.

app/services/return-requirements/check/persist-return-requirements.service.js
```diff
@@ -0,0 +1,76 @@
...
+function _buildReturnVersion (startDateDay, startDateYear, startDateMonth, licence, reason) {
+  const startDate = startDateDay
+    ? new Date(`${startDateYear}-${startDateMonth}-${startDateDay}`)
+    : licence.currentVersionStartDate
+
+  return {
+    version: 1,
+    licenceId: licence.id,
+    startDate,
+    status: 'current',
```
jonathangoulding, line 38: What should this default be ?

app/services/return-requirements/check/persist-return-requirements.service.js
```diff
@@ -0,0 +1,76 @@
...
+  await _persistData(returnRequirements, returnVersion)
+}
+
+function _buildReturnVersion (startDateDay, startDateYear, startDateMonth, licence, reason) {
+  const startDate = startDateDay
+    ? new Date(`${startDateYear}-${startDateMonth}-${startDateDay}`)
+    : licence.currentVersionStartDate
+
+  return {
+    version: 1,
```
jonathangoulding, line 35: What should this default be ?

app/services/return-requirements/submit-check.service.js
jonathangoulding, line 32: Do we need to prevent a user from approving if there are no return requirements ? (This has now been asked on the ticket) 

Should we create a requirement version on it's own if there are no returnRequirements. I can not see where someone can select not required for return ?

app/services/return-requirements/check/persist-return-requirements.service.js
```diff
@@ -0,0 +1,76 @@
...
+  return {
+    summer: requirement.returnsCycle === 'summer',
+    siteDescription: requirement.siteDescription,
+    abstraction_period_end_day: requirement.abstractionPeriod['end-abstraction-period-day'],
+    abstraction_period_end_month: requirement.abstractionPeriod['end-abstraction-period-month'],
+    abstraction_period_start_day: requirement.abstractionPeriod['start-abstraction-period-day'],
+    abstraction_period_start_month: requirement.abstractionPeriod['start-abstraction-period-month'],
+    reportingFrequency: requirement.frequencyReported,
+    collectionFrequency: requirement.frequencyCollected,
+    returns_frequency: 'year' // need to know this
```
jonathangoulding, line 65: Is this a default ?

app/services/return-requirements/check/persist-return-requirements.service.js
```diff
@@ -0,0 +1,76 @@
...
+    startDate,
+    status: 'current',
+    reason
+  }
+}
+
+function _buildReturnRequirements (requirements, returnVersion) {
+  return requirements
+    // Strip empty objects off array - investigate bug
+    .filter((requirement) => { return Object.keys(requirement).length > 0 })
```
jonathangoulding, line 46: Is this a bug at the start of the journey. 

We are seeing an empty object at the start of the array `[{}, { "this would have the journey data" }]`

Cruikshanks:
Superseded by https://github.com/DEFRA/water-abstraction-system/pull/1137

PR #1087, opened 2024-06-11
Amend no returns message in view licence Returns tab

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4461

A licence can be linked to both return versions and return submissions. However, currently, the two are not generated at the same time. Plus, it is valid that a licence has neither.

We want to make the messaging more informative so this change amends the message we show when there are no returns.

- If a licence has neither - _No requirements for returns have been set up for this licence._
- If a licence has return versions but no submissions - _No returns for this licence._
- If a licence has both - (show the table of return submissions)


app/views/licences/tabs/returns.njk
```diff
@@ -43,12 +43,15 @@
       </tr>
     {% endfor %}
     </tbody>
-  {% else %}
-    <p>No returns found</p>
+
+  {% elseif hasReturns == false and hasRequirements == false %}
```
jonathangoulding, line 47: Certainly willing to change this if it can be done clearer ?

app/services/licences/fetch-licence-has-requirements.service.js
Cruikshanks, line 1: It is up to you but I think it would be fine to just have this as a function in `view-licence-returns.service.js`.

Else I would rename the service as it is not fetching data as such but making a determination. So, for example, `DetermineLicenceHasReturnVersionsService`. 🤷

app/services/licences/fetch-licence-has-requirements.service.js
```diff
@@ -0,0 +1,34 @@
...
+ * Fetches requirements for a licence and returns a boolean if the licence has requirements for returns
+ *
+ * @param {string} licenceId - The UUID for the licence id to fetch
+ *
+ * @returns {Promise<Boolean>} the result of check if a licence has requirements for returns
+ */
+async function go (licenceId) {
+  const requirement = await _fetch(licenceId)
+
+  return !!requirement
```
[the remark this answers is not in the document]
Cruikshanks: If we leave this as a `Fetch` service then I would move it to the presenter. If we go with my suggestion to either make the call in `view-licence-returns.service.js` or rename the service then yes it's fine.

app/services/licences/fetch-licence-has-requirements.service.js
```diff
@@ -0,0 +1,34 @@
...
+}
+
+async function _fetch (licenceId) {
+  return ReturnVersionModel.query()
+    .select([
+      'id'
+    ])
+    .where('licenceId', licenceId)
+    .first()
```
Cruikshanks, line 29: [Objection.js first()](https://vincit.github.io/objection.js/api/query-builder/other-methods.html#first) doesn't actually limit the number of results from the query, it is just a helper to return the first one.

In these cases, I find it more explicit to highlight we only want one result by including a [limit()](https://vincit.github.io/objection.js/api/query-builder/find-methods.html#limit).

```suggestion
  return ReturnVersionModel.query()
    .select([
      'id'
    ])
    .where('licenceId', licenceId)
    .limit(1)
    .first()
```

app/views/licences/tabs/returns.njk
Cruikshanks: I think we can if we take a different approach. Currently, we are passing through to flags that tell the view whether the licence has returns and has requirements. We are then asking the view to work out what to show.

In my opinion this should be moved to the presenter to try and keep the view as 'dumb' as possible.

The presenter could be changed to work out what the `noReturnsMessage` should be using these flags (means we can then test it easier as well).

The only logic the view then needs is 

```
  {% if noReturnsMessage %}
    // No returns stuff
  {% else %}
    // Display returns table
  {% endif %
```

test/controllers/licences.controller.test.js
Cruikshanks, line 1: This is why I try and move as much logic out of the views as possible. It then means the only way to test stuff is with controller tests.

I think if we can implement my suggestion about where the message is defined, we can remove a number of these tests.

Cruikshanks (changes requested):
I'm going to stop here as I've suggested some things that might impact a number of the tests.

Let me know if you need to catch me about any of this.

app/views/licences/tabs/returns.njk
```diff
@@ -21,7 +21,7 @@
 
 <table class="govuk-table">
   <h2 class="govuk-heading-l">Returns</h2>
-  {% if returns.length > 0 %}
+  {% if hasReturns %}
     <thead class="govuk-table__head">
```
jonathangoulding, line 25: thought from the other pr. 

See if this can use the njk component. 

Will mark as draft whilst working on this.

app/services/licences/fetch-licence-has-requirements.service.js
jonathangoulding: In the service is fine by me.

test/controllers/licences.controller.test.js
jonathangoulding: As a result of moving this logic i think it's best to keep the has requirement logic in it's own service to ensure it returns true / false. 

Otherwise it's a bigger service test using the presenter which is currently stubbed.

Cruikshanks: 👍

app/views/licences/tabs/returns.njk
```diff
@@ -19,36 +19,56 @@
...
+      {
+        text: 'Status'
+      }
+    ],
+    rows: returnRows
+  }) }}
+{% endif %}
 
-{% if returns and pagination.numberOfPages > 1 %}
+{% if hasReturns and pagination.numberOfPages > 1 %}
```
Cruikshanks, line 71: Thinking about it, if the licence doesn't have any returns then `pagination.numberOfPages` will equal 0. So, we don't actually need the double check.

```suggestion
{% if pagination.numberOfPages > 1 %}
```

test/controllers/licences.controller.test.js
```diff
@@ -403,9 +403,15 @@ describe('Licences controller', () => {
       }
     })
 
-    describe('when a request is valid and has returns', () => {
+    describe('when a request is valid and has returns and the licence has requirements', () => {
```
Cruikshanks, line 406: We are fine with just this test. The controller tests are there to confirm we have the route and controller hooked up correctly in Hapi and that they respond with the status codes we expect.

This is why you'll more often see in our tests just

```javascript
describe('when a request is valid', () => {
  // ...
})

describe('when a request is invalid', () => {
  // ...
})
```

```suggestion
    describe('when a request is valid', () => {
```

test/controllers/licences.controller.test.js
```diff
@@ -421,21 +427,21 @@ describe('Licences controller', () => {
       })
     })
 
-    describe('when a request is valid and has NO returns', () => {
+    describe('when a request is valid and has requirements but NO returns', () => {
```
Cruikshanks, line 430: We can drop this test.

test/services/licences/determine-licence-has-return-versions.service.test.js
```diff
@@ -0,0 +1,42 @@
...
+
+      expect(result).to.be.true()
+    })
+
+    it('returns false no return versions for licence', async () => {
+      const result = await FetchLicenceHasRequirementsService.go('ed3b9b1a-94e0-480c-8ad6-60e05f5fa9f4')
+
+      expect(result).to.be.false()
+    })
+  })
```
Cruikshanks, line 41: The 'returns false' test needs to be outside of the describe block because it currently reads as; "when the licence has return versions data, it returns false no return versions for licence".

I'd also say that the data setup makes things a little confusing. I'd suggest a slight tweak might make it clearer.

```suggestion
  const licenceId = 'e004c0c9-0316-42fc-a6e3-5ae9a271b3c6'

  beforeEach(async () => {
    await DatabaseSupport.clean()
  })

  describe('when the licence has return versions', () => {
    beforeEach(async () => {
      await ReturnVersionHelper.add({ licenceId })
    })

    it('returns true', async () => {
      const result = await FetchLicenceHasRequirementsService.go(licenceId)

      expect(result).to.be.true()
    })
  })

  describe('when the licence does not have return versions', () => {
    it('returns false', async () => {
      const result = await FetchLicenceHasRequirementsService.go('ed3b9b1a-94e0-480c-8ad6-60e05f5fa9f4')

      expect(result).to.be.false()
    })
  })
```

Cruikshanks (changes requested):

app/services/licences/determine-licence-has-return-versions.service.js
```diff
@@ -0,0 +1,34 @@
...
+
+  return !!requirement
+}
+
+async function _fetch (licenceId) {
+  return ReturnVersionModel.query()
+    .select([
+      'id'
+    ])
+    .where('licenceId', licenceId)
```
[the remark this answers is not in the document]
jonathangoulding: Missed it when i put this back into it's own file.

app/views/licences/tabs/returns.njk
jonathangoulding: Good spot i missed then when moving the message into the presenter.

Cruikshanks (approved):

PR #1106, opened 2024-06-14
Feature view requirements for returns

jonathangoulding:
Ticket - https://eaflood.atlassian.net/browse/WATER-4422

Adds a new page to allow users to view the requirements for returns.

A user can click a view link on the licence setup tab under the requirements for returns.

This page is similar to previous work done on the requirements for returns check page.


Cruikshanks (changes requested):
Started adding specific comments but realised there was going to be quite a few!

But I think its because the needs for the return requirement setup journey have gotten mixed in with this. Whereas I think we can simplify things when just taking the completed record and presenting it.

This will be quicker and easier to pair on and discuss. So, we'll arrange that offline from here and progress there.

app/services/return-requirements/fetch-requirements-for-returns.service.js
```diff
@@ -0,0 +1,164 @@
...
+async function go (returnVersionId) {
+  const returnVersion = await _fetch(returnVersionId)
+
+  return {
+    ...returnVersion,
+    returnRequirements: _transformForReturnRequirements(returnVersion.returnRequirements)
+  }
+}
+
+function _agreementExceptions (returnRequirement) {
```
jonathangoulding, line 31: move all these out

jonathangoulding (commented):
Comments to address based on conversation

jonathangoulding:
New PR created based on changes required - https://github.com/DEFRA/water-abstraction-system/pull/1118

PR #1116, opened 2024-06-18
Fix message text in two-part tariff review screen

Jozzey:
On the Review screen when there are licences in review we show a message letting users know how many there are. Currently when there is only 1 licence in review the text displays as `1 licences` whereas it should be singular and display `1 licence`.

This PR corrects that text.

### DIFF
--- a/app/views/bill-runs/review.njk
+++ b/app/views/bill-runs/review.njk
@@ -101,19 +101,21 @@
   </div>
 
   {# Dynamic message either telling the user they have issues to deal with or that they can generate bills #}
-  {% if numberOfLicencesToReview > 0 %}
-    <section class="govuk-!-margin-bottom-9">
-      {{ govukInsetText({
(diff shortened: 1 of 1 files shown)


Cruikshanks (changes requested):
Sorry, I know I'm pushing a quick fix to be something bigger.

But if we pushed this logic into the presenter, not only would we make the view simpler we could also add a unit test to cover this.

I'd push for the `review-bill-run.presenter.js` to be updated to generate the message. The view can then be completely 'dumb' and just display whatever message is determined by the presenter.

Jozzey:
> Sorry, I know I'm pushing a quick fix to be something bigger.
> 
> But if we pushed this logic into the presenter, not only would we make the view simpler we could also add a unit test to cover this.
> 
> I'd push for the `review-bill-run.presenter.js` to be updated to generate the message. The view can then be completely 'dumb' and just display whatever message is determined by the presenter.

That was an option I was considering but didn't bother as it was supposed to be a quick fix and now it isn't :grin: Still shouldn't take long.

Cruikshanks (approved):

PR #1118, opened 2024-06-18
Feature-requirements-for-returns-view-page

jonathangoulding:

The requirements for return journey has been built and the check page has been finished. This showed a page with requirements for returns allowing users to make changes to requirements for returns.

This piece of work will replicate the requirements for returns check page but will not allow users to make changes (the card and information is identical, apart from the title and a return reference).

There will be a link "View" on the licence set up page that links to this page. That is added as part of this change. Work for this was done here - https://github.com/DEFRA/water-abstraction-system/pull/1054

### DIFF
--- a/app/lib/general.lib.js
+++ b/app/lib/general.lib.js
@@ -153,31 +153,40 @@ function generateAbstractionPointDetail (pointDetail) {
 }
 
 /**
- * Generate a string that represents an abstraction point based on the new data (obs change this in PR)
+ * Generate a string that represents an abstraction point based on the assumption the points have already been merged
  *
+ * When abstracting water the point at which this is done can be described in several ways depending on the number of
+ * Nation Grid References are saved against the abstraction point. This function checks for these references and builds
+ * a string that defines the details of the abstraction point.
+ *
+ * This follows the same out put as generateAbstractionPointDetail but the points have already been merged
+ *
+ * @param {Object} pointDetail - Object containing all the details for the point
+ *
+ * @returns {String} a description of the abstraction point
  */
-function generatePointDetail (point) {
+function generatePointDetail (pointDetail) {
   let abstractionPoint = null
 
-  if (point.ngr4) {
-    const point1 = point.ngr1
-    const point2 = point.ngr2
-    const point3 = point.ngr3
-    const point4 = point.ngr4
+  if (pointDetail.ngr4) {
+    const point1 = pointDetail.ngr1
+    const point2 = pointDetail.ngr2
+    const point3 = pointDetail.ngr3
+    const point4 = pointDetail.ngr4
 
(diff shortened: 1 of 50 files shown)


app/lib/general.lib.js
```diff
@@ -152,6 +152,36 @@ function generateAbstractionPointDetail (pointDetail) {
   return abstractionPoint
 }
 
+/**
+ * Generate a string that represents an abstraction point based on the new data (obs change this in PR)
+ *
+ */
+function generatePointDetail (point) {
+  let abstractionPoint = null
```
jonathangoulding, line 160: Add test for this

app/views/return-requirements/view.njk
```diff
@@ -0,0 +1,194 @@
...
+    <hr class="govuk-section-break  govuk-!-margin-bottom-2 govuk-section-break--visible">
+  </div>
+
+{% if notes %}
+  <div class="govuk-!-margin-bottom-9">
+    <h2 class="govuk-heading-l govuk-!-margin-bottom-4">Notes</h2>
+    <hr class="govuk-section-break  govuk-!-margin-bottom-2 govuk-section-break--visible">
+    {{ govukSummaryList({
+      classes: 'govuk-!-margin-bottom-2',
+      rows: [
```
jonathangoulding, line 64: Currently awaiting a decision on the user email. 

Currently this is not stored nor does it exists in the legacy data. As this affect many records a redesign might be required.

app/lib/general.lib.js
```diff
@@ -152,6 +152,45 @@ function generateAbstractionPointDetail (pointDetail) {
...
+ * Nation Grid References are saved against the abstraction point. This function checks for these references and builds
+ * a string that defines the details of the abstraction point.
+ *
+ * This follows the same out put as generateAbstractionPointDetail but the points have already been merged
+ *
+ * @param {Object} pointDetail - Object containing all the details for the point
+ *
+ * @returns {String} a description of the abstraction point
+ */
+function generatePointDetail (pointDetail) {
```
jonathangoulding, line 168: Might need to work on this name and docs.

My current understanding is limiting me to explain it better.

app/views/return-requirements/check.njk
```diff
@@ -75,34 +75,20 @@
...
-                href: "delete-note",
-                text: "Delete"
-              } if note
-            ]
-          }
-        }
-      ]
-    }) }}
+    <div class="govuk-summary-list govuk-!-margin-bottom-1">
+      <p class="govuk-body">{{ note.text | escape }}</p>
```
jonathangoulding, line 79: Old logic was - ` text: note | escape | nl2br`

Is this still needed ?

Cruikshanks (approved):

PR #1119, opened 2024-06-19
Spike for two-part tariff annual billing

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4383

We've built the first part of the two-part tariff annual billing process: matching and allocating. Once users have reviewed the results and made any changes, we need to generate the actual bills.

We'e previously done that with SROC annual and supplementary and we know we can make use of a lot of it. But two-part tariff is different in that we already know which licences to include and key details need to come from the review results not the charging information.

So, this is a spike; hack together a 'working' billing engine that generates the bills using the review data.

We never intend to merge this change. But we'll use it to direct and inform the changes we do need to make.


Cruikshanks:
Although there were other PR's involved, this was the final key step in implementing 2PT annual billing https://github.com/DEFRA/water-abstraction-system/pull/1172

So, this spike can now be closed.

PR #1130, opened 2024-06-20
Add feature flag to enable the system licence view

jonathangoulding:
There are currently links directing the pages to the legacy licence view pages, built previously in the water-abstraction-ui repo. This service is slowly being replaced.

We have rebuilt these pages (with some tweaks) in this repo with the intention of using the water-abstraction-system to view and manage licences.

This change will allow us to navigate to the water-abstraction-system licence view pages instead of the 'legacy' licence view pages in a controlled manner.

This change will be mirrored in water-abstraction-ui.

https://eaflood.atlassian.net/browse/WATER-4466


app/plugins/views.plugin.js
```diff
@@ -10,9 +10,10 @@
  * @module ViewsPlugin
  */
 
-const path = require('path')
 const Nunjucks = require('nunjucks')
 const Vision = require('@hapi/vision')
+const path = require('path')
```
Cruikshanks, line 15: `path` being where it was is intentional, as per [Top of .js Files](https://github.com/DEFRA/water-abstraction-team/blob/main/coding_conventions.md#top-of-js-files)

If you were following that convention there would be a line break after `Vision` then your config 😁

app/plugins/views.plugin.js
```diff
@@ -100,6 +101,9 @@ function context (request) {
       scope: request.auth.credentials?.scope,
       permission: request.auth.credentials?.permission
     },
+    featureFlag: {
```
Cruikshanks, line 104: This would/could be a handy place to put any feature flag so there could be more. So, I'd suggest

```suggestion
    featureFlags: {
```

Cruikshanks (changes requested):
Just a couple of tweaks and a request: don't forget `env.example`! 😁

Cruikshanks (approved):

PR #1137, opened 2024-06-24
Persist a new return version and its requirements

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4519

Having completed the return requirements setup journey (returns-required), we need to persist both the return version (start date, reason and notes) along with its return requirements to the database.

When "Approve" is clicked in the `/check` page the data in the session needs to be persisted to:

- water.return_versions
- water.return_requirements
- water.return_requirement_points
- water.return_requirement_purposes


app/services/return-requirements/generate-return-version-requirements.service.js
```diff
@@ -0,0 +1,133 @@
...
+  const { lastLegacyId } = await ReturnRequirementModel.query()
+    .max('legacyId as lastLegacyId')
+    .first()
+
+  if (lastLegacyId) {
+    return lastLegacyId + 1
+  }
+
+  return 1
+}
```
Cruikshanks, line 129: I'm afraid it's more complicated than this.

`legacy_id` Is an incremental ID by region. In order to figure it out you need to

- Select all return requirements for the same region
- Select the MAX(legacy ID) from the results
- increment the result by 1

Selecting by region can be achieved with a

```suggestion
async function _nextLegacyId (naldRegionId) {
  const { lastLegacyId } = await ReturnRequirementModel.query()
    .max('legacyId as lastLegacyId')
    .where('externalId, 'like', `${naldRegionId}:%`)
    .first()

  if (lastLegacyId) {
    return lastLegacyId + 1
  }

  return 1
}
```

Also, it's been our convention (where we spot it!) _not_ to prefix functions with `_get` or `_set` hence the tweak to the method name.

Cruikshanks (changes requested):
I think we can simplify the tests a little. But before we get into that I thought I should flag this change first.

robertparkinson (approved):

app/services/return-requirements/persist-return-version.service.js
```diff
@@ -0,0 +1,87 @@
+'use strict'
+
+/**
+ * Persists the data required to create a new return version
+ * @module PersistReturnVersionService
+ */
+
+const ReturnVersionModel = require('../../models/return-version.model.js')
```
Beckyrose200, line 8: NIT 
```suggestion
const ReturnRequirementModel = require('../../models/return-requirement.model.js')
const ReturnRequirementPointModel = require('../../models/return-requirement-point.model.js')
const ReturnRequirementPurposeModel = require('../../models/return-requirement-purpose.model.js')
const ReturnVersionModel = require('../../models/return-version.model.js')
```

PR #1138, opened 2024-06-24
Fix default licence view createdBy

jonathangoulding:

When a return version does not have a user linked to the record the default string should be "Migrated from NALD".

The assumption is a user can not create a return unless they are logged in. When the return is saved the user will be captured and stored in the createdBy column. Therefore, if the created by field is null then it is safe to assume the data has been migrated from NALD.


Cruikshanks (approved):
Only thing to point out is to do with the PR description

- there is a simplified JIRA link that can be used (not affected by changes to boards) https://eaflood.atlassian.net/browse/WATER-4422
- you have a typo; `createBty`

PR #1141, opened 2024-06-25
Fix licence set up links

jonathangoulding:

During a refactor the links for the page where wrapped up into a single object.

The previous links where missed. This change adds the link to the href block of the gov uk buttons.

The links are used in the if condition and work / where done as part of the original refactor.

### DIFF
--- a/app/views/licences/tabs/set-up.njk
+++ b/app/views/licences/tabs/set-up.njk
@@ -134,7 +134,7 @@
   <p>
     {{ govukButton({
       text: "Set up a new charge",
-      href: setupNewCharge,
+      href: links.chargeInformation.setupNewCharge ,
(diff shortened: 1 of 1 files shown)


app/views/licences/tabs/set-up.njk
```diff
@@ -134,7 +134,7 @@
   <p>
     {{ govukButton({
       text: "Set up a new charge",
-      href: setupNewCharge,
+      href: links.chargeInformation.setupNewCharge ,
```
Cruikshanks, line 137: The NIT'est of NITs!

```suggestion
      href: links.chargeInformation.setupNewCharge,
```

Cruikshanks (changes requested):

Cruikshanks (approved):
Just so you know, also added the link to the ticket into your PR description.

PR #1144, opened 2024-06-26
Spike allowing test to be independent of cleanup.

jonathangoulding:
Currently, tests need to run await DatabaseSupport.clean() which we assume is taking considerable time for us to investigate making the tests more performant. 

Proposal:

> When `npm test` is run have a script run prior to the test to clean the database once.
>
> Update tests to manage themselves with unique ids. Use the id created by the DB and pass it around the test or use our generateUUID function where appropriate to allow the test to generate their own id's/

### DIFF
--- a/test/services/bill-runs/determine-blocking-bill-run.service.test.js
+++ b/test/services/bill-runs/determine-blocking-bill-run.service.test.js
@@ -9,28 +9,33 @@ const { expect } = Code
 
 // Test helpers
 const BillRunHelper = require('../../support/helpers/bill-run.helper.js')
-const DatabaseSupport = require('../../support/database.js')
-const { determineCurrentFinancialYear } = require('../../../app/lib/general.lib.js')
 const RegionHelper = require('../../support/helpers/region.helper.js')
+const { determineCurrentFinancialYear } = require('../../../app/lib/general.lib.js')
+const { generateUUID } = require('../../../app/lib/general.lib.js')
 
 // Thing under test
 const DetermineBlockingBillRunService = require('../../../app/services/bill-runs/determine-blocking-bill-run.service.js')
 
 describe('Determine Blocking Bill Run service', () => {
   let batchType
+  let billRunIdOne
+  let billRunIdTwo
   let financialEndYear
   let regionId
   let season
   let toFinancialYearEnding
 
   beforeEach(async () => {
-    await DatabaseSupport.clean()
-
     const { endDate } = determineCurrentFinancialYear()
+
     toFinancialYearEnding = endDate.getFullYear()
 
     const region = await RegionHelper.add()
+
     regionId = region.id
+
+    billRunIdOne = generateUUID()
+    billRunIdTwo = generateUUID()
   })
 
   describe('when the user is setting up an annual bill run', () => {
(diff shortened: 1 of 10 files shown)


test/services/idm/fetch-user-roles-and-groups.service.test.js
```diff
@@ -26,9 +26,11 @@ describe('Fetch User Roles And Groups service', () => {
   let testGroup
 
   beforeEach(async () => {
-    await DatabaseSupport.clean()
-
-    testUser = await UserHelper.add()
+    testUser = await UserHelper.add(
+      {
+        username: `${generateUUID()}@test.com`
```
jonathangoulding, line 31: Was tempted to use faker but this does the trick. We just need a unique name but don't want to replicate the bfe multiple times. Could look to nest / create a simple random text func

test/services/licences/fetch-licence-bills.service.test.js
```diff
@@ -8,23 +8,27 @@ const { describe, it, beforeEach } = exports.lab = Lab.script()
...
-  const licenceId = '96d97293-1a62-4ad0-bcb6-24f68a203e6b'
+
+  let billId
+  let billingAccountId
+  let billRunId
+  let licenceId
 
   beforeEach(async () => {
-    await DatabaseSupport.clean()
+    billId = generateUUID()
```
jonathangoulding, line 28: Simpler solution then going into the helpers. 

This would be a faster / simpler approach to changing the helpers and the potential impact on other tests.

test/services/bill-runs/fetch-live-bill-runs.service.test.js
```diff
@@ -126,15 +122,14 @@ describe('Fetch Live Bill Runs service', () => {
           const results = await FetchLiveBillRunsService.go(regionId, financialYearEnding, false)
 
           expect(results).to.have.length(1)
-          expect(results[0].id).to.equal('1021c5bc-673c-48fa-98dd-733b46c84f90')
+          expect(results[0].id).to.equal(billRun.id)
```
jonathangoulding, line 125: I think this is a necessary `evil` in the test to all this to be dynamic and not static. 

I think this is safe as it will fail if the id does not match. Worth a discussion / clarification.

jonathangoulding:
Just a memo to self, potential obvious observation that the DB in the CI will be clean compared to locally so this clean up may have different affects locally and in the CI.

test/services/bill-runs/determine-blocking-bill-run.service.test.js
```diff
@@ -22,15 +22,18 @@ describe('Determine Blocking Bill Run service', () => {
...
   beforeEach(async () => {
-    await DatabaseSupport.clean()
-
     const { endDate } = determineCurrentFinancialYear()
     toFinancialYearEnding = endDate.getFullYear()
 
     const region = await RegionHelper.add()
     regionId = region.id
+
+    billRunId = generateUUID()
```
jonathangoulding, line 35: I think this is a good example / case for generating the  uuid in the code where they were before. A minimal touch has fixed this test.

jonathangoulding:
Some tests just need a clean DB and don't need to have the id's changed. Need to see about the effort /  need to change these tests.

jonathangoulding:
Based on conversations the following process will be followed to refactor the test as minimally as possible. 

If a record is being returned from a helper func to a variable then the id from the record should be used to assert / help set up tests. - example https://github.com/DEFRA/water-abstraction-system/pull/1144/files#diff-37590134309bd72556ab2c600be74a59fd3900f0fb02f255777e43af513b48a2R40

If the id is hard coded and the record is not returning to a variable then replace the hard coded uuid using the generate UUID function and move the setting of the variable into the before each. example - https://github.com/DEFRA/water-abstraction-system/pull/1144/files#diff-d737f2d69286c9b96744b7153bbe95a5ae88010af96eb2a7f5bf16e8180a307dR35

Cruikshanks (approved):

PR #1147, opened 2024-06-28
Stop using version in package.json

sujithvg:
https://github.com/DEFRA/water-abstraction-team/issues/107

Currently, we have each app report details about itself on a `/health/info` endpoint. For each app this returns the current commit hash and the version recorded in `package.json`.

All this then gets shown in our main `/service-status` endpoint (or `/health/info` in [water-abstraction-system](https://github.com/DEFRA/water-abstraction-system) if you want to go direct!)

Normally, when we cut a release we're ok to just take whatever is on `main` and tag that as the next version. We run `npm version [minor|patch]` and both the tagging and the updating of the version in `package.json` is done for us.

But in a recent release `main` in a couple of the repos contained changes we weren't yet ready to include. No problem, we just have to check out an earlier commit and create the [git tag](https://git-scm.com/book/en/v2/Git-Basics-Tagging) there. This does mean though we can't run `npm version` and increment the version in `package.json`.

The issue came when our QA went to `/service-status` to double-check the versions now running. The commit hashes were all correct; we were running the version of the code we'd tagged for release. But the versions displayed were now out of sync with the git tags.

There are only 3 solutions as far as we can see

- wait until everything's been tested - just not practical
- go back to using git-flow and development/release branches - not on your life!
- stop using the version in `package.json` - 👈 😁

Whenever we 'cut' a release we create a tag at the commit we intend to push to production. The tag is our version, not what's in the `package.json`. So, if all our `/health/info` endpoints returned the result of `git describe --always --tags` instead they would be

- more accurate
- avoid this confusion


Cruikshanks (changes requested):
Ahhhh! Welcome to our primary repo (the one we like 😁).

So far, you have been making changes to the legacy repos where we try to do the minimum necessary when changes are needed.

In **water-abstraction-system** we're happy to go the extra mile to try and keep things ship shape and Bristol fashion! 🚢✨

I've purposefully not highlighted it in previous PR's because we don't really care. But we have a convention (admittedly still [waiting to be written up](https://github.com/DEFRA/water-abstraction-team/issues/92)) for function name ordering that places them alphabetically.

So, in this case, `_gitTagReference()` would go beneath `_getCommitHash()`.

If you are new to pull request reviews, it is more common to get a 'request changes' than an immediate `approve'. And this is a primary channel for communicating the 'team-way' of doing things, especially for our new members. So, you're going to get many more of these in the future, and they are _never_ an admonishment or to be seen as having done something wrong 😁 .

If you have any problems or concerns though, do please give me a ping.

---

On a practical level, have a look at the feedback. If it makes sense and you agree, simply press on and make the changes and then push them as you have been doing.

When you think something is ready to be reviewed again, locate the reviewers section in the top right of the GitHub page for the PR. You'll see my name and a refresh icon next to it. Click it, and that will tell me the PR is ready to be re-reviewed.

PR #1167, opened 2024-07-05
Spike building a licence history page

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4541

We've started talking about the NALD 'mod log'. Our current understanding is that whenever a user creates a new return, licence, or charge version NALD pops up an 'event' where the user is expected to enter a reason and a note. Essentially, why the change is being made.

They then have reports that combine these events. They use these to ensure the 3 legs of water abstraction are in sync. (They often use the metaphor of a stool, with licence (abs), returns and charging the 3 legs).

When viewed as a report, this essentially just seems to be what other systems would call history i.e. the history of changes to a licence.

There is a concern that moving away from the NALD also means losing the 'mod log' and therefore how would they ensure things are in sync.

This 'spike' is to build an example licence history page that combines licence, return and charge versions into one place to see if any of these concerns can be alleviated.

water-4566-initial-spike


Cruikshanks:
This has served its purpose. We now have https://github.com/DEFRA/water-abstraction-system/pull/1182 approved and ready to be merged. So, bye-bye spike!

PR #1171, opened 2024-07-08
Update display of points in return requirements

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4567

Currently, when we display the points for a return requirement in both the /check page and view page we simply list them. However, we know there are a number of licences which can be linked to more than 5 points. The most extreme is 36! It is a common pattern in the service that when we need to display more than 5 items, we’ll use a [GOV.UK details component](https://design-system.service.gov.uk/components/details/).

This PR is focused on amending both return requirement views to use a details component if a return requirement is linked to more than 5 points. If the requirement is linked to 5 or less, the points should be listed out (no change).


app/views/return-requirements/check.njk
```diff
@@ -103,6 +104,31 @@
...
       {% for requirement in requirements %}
+        {# Generate the points cell #}
+        {# When a requirement has more than 5 points it is displayed as a GOVUK Details component which displays the count #}
+        {# of points as the summary. Users must expand the block to get the full list. #}
+        {% set pointCell %}
+          {% set pointsList %}
+            {% set count = requirement.points | length %}
+            {% for point in requirement.points %}
+              {% set count = loop.index %}
+              {{ point }}<br>
```
Cruikshanks, line 115: It's my fault @rvsiyad ! I know you would have copied this from something I did in the bill runs.

I shouldn't have used `<br>` to separate the elements. 🤦

Fortunately, @jonathangoulding added a macro to this view (look at the top of the file) specifically for listing out items (that was the `newLine(requirement.points)` call this change has replaced).

Could you have a crack at incorporating that method rather than reproducing my `<br>` hack?

Cruikshanks (changes requested):
The comment I added will apply to the view template as well.

Sorry! 😳

Cruikshanks (approved):

PR #1173, opened 2024-07-09
Spike migrating import licence to system

jonathangoulding:
https://eaflood.atlassian.net/jira/software/c/projects/WATER/boards/96?quickFilter=347&selectedIssue=WATER-4536

We need to explore and establish how we could migrate the import licence process from the import service.

Questions have been raised around packages used by the import service and wether or not we will keep them or rewrite that functionality. For example, we will replace moment.js with node date time.

A mural board has been made to capture the process, dependencies and order in which we would need to do this work.

### DIFF
--- a/app/import/helpers/charge-merge.js
+++ b/app/import/helpers/charge-merge.js
@@ -1,0 +1,65 @@
+'use strict'
+
+// TODO: this will need to use built in date
+const moment = require('moment')
+// TODO: you might not need lodash...
+const { last, omit, isEqual } = require('lodash')
+
+const DATE_FORMAT = 'YYYY-MM-DD'
+
+/**
+ * Checks whether 2 objects have adjacent date ranges.
+ * Object a must have an endDate that is either null, or 1 day before
+ * the startDate property of object b
+ * @param {Object} a
+ * @param {Object} b next row
+ * @return {Boolean}
+ */
+const isAdjacentDateRange = (a, b) => {
+  const expectedEndDate = moment(b.startDate).subtract(1, 'day').format(DATE_FORMAT)
+
+  return [null, expectedEndDate].includes(a.endDate)
+}
+
+/**
+ * Default implementation of equality test between two objects
+ * for history merging.
+ * Checks if the objects are equal when taking into account all
+ * properties except startDate and endDate
+ * @param {Object} a
+ * @param {Object} b
+ * @return {Boolean}
+ */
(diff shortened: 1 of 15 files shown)


jonathangoulding:
This branch shows how we can port the existing code into this repo and break it down. 

The sql used to get the data from the import table can and should be used with minmal changes. 

The transformers should be broken out into

app/import/helpers/charge-merge.js
```diff
@@ -0,0 +1,65 @@
+'use strict'
+
+// TODO: this will need to use built in date
+const moment = require('moment')
```
jonathangoulding, line 4: It has been determined that moment will not bee used when the code is moved into the system as we do no use a date library.

app/import/helpers/charge-merge.js
```diff
@@ -0,0 +1,65 @@
...
+// TODO: you might not need lodash...
+const { last, omit, isEqual } = require('lodash')
```
jonathangoulding, line 6: Lodash - will need to be removed / replace with js funcs

app/import/licence/services/fetch-licence-from-import.service.js
```diff
@@ -0,0 +1,96 @@
...
+  return {
+    licence,
+    section130Agreements,
+    tptAgreements
+  }
+}
+
+async function _getTwoPartTariffAgreements (regionCode, licenceId) {
+  const query =
+    `  SELECT a.*,
```
jonathangoulding, line 29: The sql used to get the data from the import table can and should be used with minmal changes. 

Below is an example.

knexfile.application.js
```diff
@@ -24,7 +24,7 @@ const { legacyDbSnakeCaseMappers } = require('./app/lib/legacy-db-snake-case-map
 const knexfile = require('./knexfile')
 
 for (const environment in knexfile) {
-  Object.assign(knexfile[environment], legacyDbSnakeCaseMappers({ underscoreBeforeDigits: true }))
```
jonathangoulding, line 27: We need to work out how this won't apply to the import table querys.

test/import/licence/fixtures/licences/agreements.js
```diff
@@ -0,0 +1,37 @@
+'use strict'
+
+module.exports = {
```
jonathangoulding, line 3: Fixtures allow us to build a map of data from the import service to use across the import testing suite.

PR #1174, opened 2024-07-09
Update tear down service

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4450

Following on from writing the acceptance test for the two-part tariff review processed, we realised our tear-down service doesn't include the new tables created for the two-part tariff review. This PR is updating the tear-down service to add in those tables. This means all the data created from running the acceptance tests gets deleted.

### DIFF
--- a/app/services/data/tear-down/public-schema.service.js
+++ b/app/services/data/tear-down/public-schema.service.js
@@ -68,22 +68,6 @@ async function _deleteAllTestData () {
 
   DELETE
   FROM
-    "public"."review_charge_elements" AS "rce"
-      USING "public"."review_charge_references" AS "rcr",
-    "public"."review_charge_versions" AS "rcv",
-    "public"."review_licences" AS "rl",
-    "public"."bill_runs" AS "br",
-    "public"."regions" AS "r"
-  WHERE
-    "r"."nald_region_id" = 9
-    AND "rce"."review_charge_reference_id" = "rcr"."id"
-    AND "rcr"."review_charge_version_id" = "rcv"."id"
-    AND "rcv"."review_licence_id" = "rl"."id"
-    AND "rl"."bill_run_id" = "br"."id"
-    AND "br"."region_id" = "r"."id";
-
-  DELETE
-  FROM
     "public"."review_charge_references" AS "rcr"
       USING "public"."review_charge_versions" AS "rcv",
     "public"."review_licences" AS "rl",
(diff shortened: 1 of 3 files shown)


app/services/data/tear-down/public-schema.service.js
```diff
@@ -0,0 +1,132 @@
...
+    "public"."regions" AS "r"
+  WHERE
+  "r"."nald_region_id" = 9
+    AND "rce"."review_charge_reference_id" = "rcr"."id"
+    AND "rcr"."review_charge_version_id" = "rcv"."id"
+    AND "rcv"."review_licence_id" = "rl"."id"
+    AND "rl"."bill_run_id" = "br"."id"
+    AND "br"."region_id" = "r"."id";
+
+  DELETE
```
Jozzey, line 69: Copy & paste error here I think. The `review_charge_elements` table is already deleted previously on line 53

Jozzey (changes requested):

app/services/data/tear-down/public-schema.service.js
Beckyrose200: Good spot

app/services/data/tear-down/public-schema.service.js
```diff
@@ -0,0 +1,116 @@
...
+}
+
+async function _deleteAllTestData () {
+  return db.raw(`
+  ALTER TABLE public.review_returns DISABLE TRIGGER ALL;
+  ALTER TABLE public.review_charge_elements_returns DISABLE TRIGGER ALL;
+  ALTER TABLE public.review_charge_elements DISABLE TRIGGER ALL;
+  ALTER TABLE public.review_charge_references DISABLE TRIGGER ALL;
+  ALTER TABLE public.review_charge_versions DISABLE TRIGGER ALL;
+  ALTER TABLE public.review_licences DISABLE TRIGGER ALL;
```
Cruikshanks, line 21: My two-cents!

We need to disable these triggers on the legacy tables because they set up foreign key constraints. This means we can't delete record `x` because it's linked to `y` in a different table.

Unless I missed something, we didn't complicate things by adding any foreign key constraints so you _shouldn't_ need to disable and re-enable any triggers.

Cruikshanks (changes requested):
I think we can drop all the trigger stuff.

app/services/data/tear-down/public-schema.service.js
Beckyrose200: That's cool, Ive dropped those now. Are you happy with me using the test region to identify the test data rather than adding an 'is_test' column to the reviewLicences table?

Jozzey (approved):

Cruikshanks (approved):

PR #1182, opened 2024-07-10
Build new licence history page

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4566

Billing & Data rely on the NALD ‘mod log’ to be able to see that all 3 legs are in sync when managing a licence. For example, when a licence is varied (a new licence version is created), new charge and return versions are also created (if applicable).

This PR is focused on building a new 'History' page to replicate ‘mod log’ within our system. It should display all licence, charge and return versions for the licence in a single table (no pagination), ordered by created date in descending order.


app/services/licences/fetch-licence-history.service.js
```diff
@@ -0,0 +1,89 @@
...
+      db.raw("'return-version' as entry_type"),
+      'returnVersions.id as entry_id',
+      'returnVersions.createdAt as created_at'
+    )
+    .modify('history')
+    .orderBy([
+      { column: 'version', order: 'desc' }
+    ])
+    .where('returnVersions.licenceId', licenceId)
+}
```
Cruikshanks, line 85: A few things in this.

## Renaming fields with snake case

Renaming fields, for example, `createdAt` to `created_at` is doing nothing. You can see this in your presenter where you are still referencing everything as `createdAt`. 

We reference a view for all existing tables we use rather than the table directly. Doing so allows us to

- ignore fields that are never used or always populated with the same value
- rename fields to be consistent, for example, `date_created` to `created_at`

All **Objection.js** and **Knex** queries (`db('licences).withSchema('public).select(*)` for example) then have Knex snake case mappers applied (see `knexfile.application.js`). This is what converts the PostgreSQL snake_case field names into JavaScript friendly camelCase.

My source query bypassed the snake case mappers by using `db.raw()` fully (not as a partial within an existing query). Hence, fields would have come back as snake_case.

## Renaming to be consistent across queries

The fields were renamed in my source query to use generic names because the query used a [UNION ALL](https://www.w3schools.com/sql/sql_ref_union_all.asp) to bring the results from across 3 tables into one.

Now the service is firing these queries individually via the Objection.js models, there is no point in doing this.

## Only one query needed

The service is currently fetching the licence, then fetching the charge, licence and return versions in separate queries. But as these are linked to the `LicenceModel` via the relationships we have defined in the model, they could all be obtained with a single Objection.js query to `LicenceModel` and then `withGraphFetched()` for each of the version types.

The service could be fetching a single licence with all its associated versions and their mod logs attached as a single instance to the presenter to them be formatted ready for the view.

Cruikshanks (changes requested):
I've focussed on the Fetch service because there are some changes we'll need to make. Those changes then might have knock on changes to the downstream services so there seemed little point in going through those now.

Any questions or problems give me a ping and we can always pair on getting this PR ready! 😁

Cruikshanks (approved):

PR #1193, opened 2024-07-16
Add region test seeder

jonathangoulding:

The public.regions table is what we call reference data. We assume this table will not change frequently, if ever. 

As part of the work to replace the legacy import service we have decided to seed any data we have deemed reference data. 

We will use these seeders in testing. We need to add this now as the nald region id when randomly generated is between 1-8. This was causing random test failures when using the normal RegionHelper.add(). 

This was found because we have moved away from clearing the database before each test and having a pre test clean the database instead. (This is previous work to allow us to potentially speed up the tests and change our test framework)

### DIFF
--- a/db/clean.js
+++ b/db/clean.js
@@ -18,6 +18,7 @@ async function run () {
     await Database.clean()
     console.log('Database cleaned')
   } catch (error) {
+    console.log(error)
     process.exit(1)
   } finally {
     Database.closeConnection()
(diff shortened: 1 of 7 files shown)


jonathangoulding:
I have tested against the acceptance tests and they are working. 

They will either create there own new region or update the region.

robertparkinson (approved):

PR #1195, opened 2024-07-18
Import Licence versions

jonathangoulding:

We will leave the import service running to add the NALD data in the import tables, but we will slowly replace the import-licence functionality and move this into the system repo.

This change will add the licence versions to the import logic.

The licence versions rely on the created / existing licence id.

This work needs to be done first to allow us to remove the licence and licence versions from the old import logic.


app/services/import/legacy-import/licence-versions.mapper.js
```diff
@@ -0,0 +1,83 @@
...
+    return mapLicencePurpose(purpose)
+  })
+}
+
+module.exports = {
+  go
+}
+
+const mapLicencePurpose = (purpose) => {
+  // double check no longer needed because of update sql
```
jonathangoulding, line 62: // double check no longer needed because of update sql
  // issue: purpose.AABV_ISSUE_NO,
  //   increment: purpose.AABV_INCR_NO,

app/validators/import/licence-versions.validator.js
```diff
@@ -0,0 +1,61 @@
...
+    timeLimitedEndDate: Joi.date().iso().allow(null),
+    notes: Joi.string().allow(null),
+    annualQuantity: Joi.number().allow(null),
+    externalId: Joi.string().required(),
+    instantQuantity: Joi.number().allow(null),
+    hourlyQuantity: Joi.number().allow(null),
+    dailyQuantity: Joi.number().allow(null)
+  })
+
+  const schema = Joi.array().items({
```
jonathangoulding, line 42: what if a licence does not have versions ?

app/validators/import/licence-versions.validator.js
```diff
@@ -0,0 +1,61 @@
...
+ * Checks that the data for inserting/updating the public.licence_versions
+ * and public.licence_versions_purposes table is valid
+ *
+ * @param {ImportLicenceVersionType[]} data The data to be validated
+ *
+ * Throws an error if anything fails
+ *
+ */
+function go (data) {
+  const purposeSchema = Joi.object({
```
jonathangoulding, line 24: does a licence version have to have a purpose ?

jonathangoulding:
Add test for create and updated - assert the updated / not updated response. They should be different ?

jonathangoulding (commented):
.

app/services/import/persist-licence-versions.service.js
```diff
@@ -0,0 +1,100 @@
...
+      ...purpose,
+      primaryPurposeId: primaryPurpose.id,
+      secondaryPurposeId: secondaryPurpose.id,
+      purposeId: purposeUse.id,
+      createdAt: new Date().toISOString(),
+      updatedAt: new Date().toISOString()
+    })
+    .onConflict('externalId')
+    .merge([
+      'abstractionPeriodEndDay',
```
jonathangoulding, line 61: Why does it only update some ?

app/validators/import/licence-versions.validator.js
jonathangoulding: This affects he validation service test as well

test/services/import/legacy-licence.service.test.js
```diff
@@ -5,58 +5,187 @@ const Lab = require('@hapi/lab')
...
+            abstractionPeriodStartDay: 1,
+            abstractionPeriodStartMonth: 4,
+            abstractionPeriodEndDay: 31,
+            abstractionPeriodEndMonth: 3,
+            timeLimitedStartDate: null,
+            timeLimitedEndDate: null,
+            notes: null,
+            instantQuantity: null,
+            dailyQuantity: 1500.2,
+            hourlyQuantity: 140.93,
```
jonathangoulding, line 167: What is rounding ?

test/services/import/_fixtures/import-licence-versions.fixture.js
```diff
@@ -0,0 +1,46 @@
...
+
+function importLicenceVersionPurpose () {
+  return {
+    abstractionPeriodEndDay: 31,
+    abstractionPeriodEndMonth: 3,
+    abstractionPeriodStartDay: 1,
+    abstractionPeriodStartMonth: 4,
+    annualQuantity: 545520,
+    dailyQuantity: 1500.2,
+    externalId: `${randomInteger(1, 9)}:${randomInteger(10000004, 99999999)}`,
```
jonathangoulding, line 13: These would have to be functions otherwise the generate data will be the same for each test unless we put the require in the bfe. 

We could create / follow the pattern of the model helpers and have it export a single func to generate the data. The issue with this only the effort to desctruce the data from the array formats in this case. 

Often we only want the one purpose in `licenceVersion.purposes[0]`

app/validators/import/licence.validator.js
```diff
@@ -9,37 +9,37 @@ const Joi = require('joi')
...
     throw new Error(result.error.details[0].message)
   }
 }
 
+const _schema = Joi.object({
+  expiredDate: Joi.string().allow(null).custom(CustomDateValidator.isValidISODate),
+  lapsedDate: Joi.string().allow(null).custom(CustomDateValidator.isValidISODate),
+  licenceRef: Joi.string().required(),
+  naldRegionId: Joi.number().required(),
+  regions: Joi.object({
```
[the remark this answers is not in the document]
jonathangoulding: All region codes are present in the import and so it is assumed that all fields are required.

app/services/import/persist-licence-versions.service.js
```diff
@@ -0,0 +1,105 @@
...
+const PurposeModel = require('../../models/purpose.model.js')
+
+/**
+ * Saves the licence versions, purposes and conditions
+ *
+ * @param {ImportLicenceVersionType[]} licenceVersions
+ * @param {string} licenceId
+ */
+async function go (licenceVersions, licenceId) {
+  return Promise.all(licenceVersions.map(async (version) => {
```
jonathangoulding, line 21: I don't think this needs to return anything. 

We can test the DB entires. 

The .returning statement does not seem to limit the response either.

Cruikshanks (changes requested):
This is a _big_ PR so it is going to take me a number of passes. So, this is the first tranche of feedback.

## @typedef comments

I've had a mull and decided I don't think this is a feature I want to start implementing. Its functionality seems to be closely tied to TypeScript, especially in VSCode and I'm wary 'type panic'.

If the purpose is to inform a future dev what properties an object is expected to have, then [@param](https://jsdoc.app/tags-param) is sufficient. `@typedef` would save duplicating that info if it is expected elsewhere, but then the tools let us down.

If it worked out of the box (my research suggests you can hack VSCode to get it to work across files, but with caveats) it might be worth following up on. But even then, there's the overhead of reviewing the existing code and maintaining this info going forward.

It's hard enough just trying to get folks to be consistent with the standard JSDoc comments (hint-hint, you may want to check some of your spacing 😉 ).

## Background - Foreground

The controller needs to follow the pattern most clearly seen in `jobs.controller.js`.

```javascript
async function licence (request, h) {
  const { licenceRef } = request.payload

  LegacyImportLicenceService.go(licenceRef)

  return h.response().code(204)
}
```

This means the client is not awaiting the import to complete, and the work can be done in the background.

This also requires a change in `LegacyImportLicenceService`.

> BTW the file name is wrong. It needs to be `app/services/import/legacy-import-licence.service.js`

```javascript
  try {
    // ..
    calculateAndLogTimeTaken(startTime, 'Import licence complete', { licenceRef })
  } catch (error) {
    global.GlobalNotifier.omfg('Licence import failed', { licenceRef }, error)
  }
```

Boom is intended for relaying HTTP error responses. With this pattern, we've already responded to the client, so Boom is redundant.

We don't throw the error, we swallow it to prevent the app from crashing. Also note my tweak to `calculateAndLogTimeTaken()` to keep messages consistent.

## console.debug()

These are actually adding noise to the unit tests at the moment. They are also something I feel are only needed when trying to track down an issue. We have no power to switch the logging level in production without an RfC. But in our non-prod environments, we can deploy any branch we like and trace the log output.

Therefore, it's easy to add them temporarily to just those places they are needed when they are needed.

So, I think we can drop them from this PR.

## Declaring query in const first

We need to avoid generating queries where params are injected using string interpolation. Though unlikely, this opens the door to SQL injection attacks. The Knex.js `raw()` function supports [paramter binding](https://knexjs.org/guide/raw.html#raw-parameter-binding) which we have used in other places. (I find positional binding the easiest).

Also, as we have no intent to use the query elsewhere, we opt to declare it directly in the call.

```javascript
  const { rows: [row] } = await db.raw(`
    SELECT
      licence."AREP_AREA_CODE"
      -- Other select lines
    FROM import."NALD_ABS_LICENCES" licence
      WHERE licence."LIC_NO" = ?
  `, [licenceref])
```

---

Breaking for coffee. We'll leave this first pass with you. Ping me if any queries/challenges or when you are ready for me to take another look.

jonathangoulding:
Parameter binding updated.

jonathangoulding:
console debug removed.

test/services/import/legacy-licence.service.test.js
jonathangoulding: investiagte

app/services/import/persist-licence-versions.service.js
jonathangoulding: remove from here update test.

jonathangoulding:
js docs updated

app/services/import/legacy-import/licence.mapper.js
```diff
@@ -56,8 +54,8 @@ const _regions = (licenceData) => {
  *
  * It is assumed one of these will always exist
  *
- * @param {LegacyLicenceType} licence
- * @param {LegacyLicenceVersionsArray} licenceVersions
+ * @param {object} licence - the legacy licence data
+ * @param {object[]} licenceVersions - the legacy licence versions
```
jonathangoulding, line 58: The sort in this func has come up on sonar look into it

Cruikshanks (approved):

PR #1200, opened 2024-07-22
Some returns not received error incorrectly applies 2pt

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4605

During testing for the two-part tariff review pages, the issue of `Some returns not received` has been picked up as being incorrectly used. When the status of a return is due, an issue flag gets marked against that return saying 'No returns received.' If this return matches a charge element an additional flag gets marked on the charge element saying 'Some returns not received'. Whilst technically this is correct based on the initial requirements, the business has since asked for this to be changed. 

The changes are as follows:
* For a return with a status of due, that matches a charge element and the charge element only matches the one return, then the issue 'No returns received' will be persisted against the return AND the charge element. The charge element will not have the issue 'Some returns not received' (because it only matched one return that wasn't received, the text 'some' here does not make sense.)The implications of this mean that when all the issues on the returns and the charge elements are added together to make a list for the licence because it is the same issue on the return and charge element it will be deduplicated and therefore the licence won't get additionally marked as having 'Multiple issues'. Another implication is when a user filters the licence by 'No returns received' they will see a list of that issue from the return and charge elements. However the business is ok with this, and it would likely not cause any issues. 

* For a return with a status of due, that matched a charge element and the charge element has also matched other returns, then the issue 'No returns received' will be flagged against the return and the current issue of 'Some returns not received' will continue to flag on the charge element. This is no different to the current behaviour.

* Lastly, for a charge element that has matched multiple returns and all the returns have a status of due, this will be like scenario one again where the return and the charge element get flagged as 'No returns received'.


app/services/bill-runs/two-part-tariff/determine-licence-issues.service.js
```diff
@@ -220,14 +239,37 @@ function _returnLogIssues (returnLog, licence) {
...
+    noReturnsReceived = matchingReturnLogs.every((matchingReturnLog) => {
+      return matchingReturnLog.status === 'due'
+    })
+  } else {
+    noReturnsReceived = false
+  }
+
+  // If someReturnsNotReceived is true it means either some of the returnLogs matched to the element have a status of
+  // due, or all of them
+  const someReturnsNotReceived = matchingReturnLogs.some((matchingReturnLog) => {
```
robertparkinson, line 265: I think this could be reworked. if you inverted the test so look for responses that are not equal to 'due' you know for certain that noReturnsReceived is false and can get rid of the if statement above I think? and if you know that at least one of them is not equal to 'due' there's no need to do the every until after that? do you still need the every after that?

robertparkinson (changes requested):

app/services/bill-runs/two-part-tariff/determine-licence-issues.service.js
Beckyrose200: Screenshot 2024-07-23 at 17 09 43
I managed to get it to work with the `.some` function, but when `some` takes an empty array it returns false, this was being flipped with the bang operator causing the 'noReturnsRecieved' to be true when it shouldn't be. This meant I needed to add the if statement back in, for it to only execute when we know there are returnLogs. As you can see from the photo the only difference between the two working versions is the `.some` or the `.every`. I decided to go with the `every` as I felt it was more readable and you don't have to get your head around any switching on the true/false with the bang operators

robertparkinson (approved):

PR #1201, opened 2024-07-22
Replace view licence contact details page

sujithvg:
https://eaflood.atlassian.net/browse/WATER-4559

**Context**

We recently rebuilt the legacy view licence page, implementing the new version in our shiny new codebase water-abstraction-system. Substantial changes needed to be made to support the return requirements, so this was a great opportunity to also deal with the tech debt of the legacy view licence page.  One of the pages that is linked to from the summary tab is the view licence contacts page. If we move that into our code base, we make another small dent in moving away from the legacy code.

water-4559_legacy_page

**Task**
The page is currently accessed from the view licence summary tab (search for any licence on main page, select licence, summary tab will be shown).

Screenshot 2024-07-04 at 09 32 56

The task is to ‘spike’ rebuilding the page as is, but following the conventions and patterns used in water-abstraction-system.


Cruikshanks:
This was superseded by https://github.com/DEFRA/water-abstraction-system/pull/1238 (I think! 😬 ) so we can say goodbye to this PR.

PR #1206, opened 2024-07-23
https://eaflood.atlassian.net/browse/WATER-4579

jonathangoulding:
The public.purposes, public.primary_purpose and public.secondary_purpose tables are what we call reference data. We assume these tables will not change frequently, if ever.

As part of the work to replace the legacy import service we have decided to seed any data we have deemed reference data.

We will use these seeders in testing.

This change will add a seeder for the reference data for purposes, primary purpose and secondary purpose. It will export its data to be used in tests. 

Where test have failed they have been updated and where possible the db clean function has been removed from the bfe.

### DIFF
--- a/test/models/licence-version-purpose.model.test.js
+++ b/test/models/licence-version-purpose.model.test.js
@@ -8,27 +8,34 @@ const { describe, it, beforeEach } = exports.lab = Lab.script()
 const { expect } = Code
 
 // Test helpers
-const DatabaseSupport = require('../support/database.js')
-const LicenceVersionModel = require('../../app/models/licence-version.model.js')
 const LicenceVersionHelper = require('../support/helpers/licence-version.helper.js')
+const LicenceVersionModel = require('../../app/models/licence-version.model.js')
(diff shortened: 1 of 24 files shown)


robertparkinson (approved):

test/models/licence-version-purpose.model.test.js
```diff
@@ -231,14 +217,24 @@ describe('Licence Version Purposes model', () => {
...
+        return secondaryPurpose.legacy_id === 'ELC'
+      })
+      validPurpose = PurposesSeeder.data.find((purpose) => {
+        return purpose.legacy_id === '200'
+      })
 
-      validPrimaryPurpose = await PrimaryPurposeHelper.add({ legacyId: 'P' })
-      validSecondaryPurpose = await SecondaryPurposeHelper.add({ legacyId: 'ELC' })
-      validPurpose = await PurposeHelper.add({ legacyId: '200' })
+      console.log('i', invalidPrimaryPurpose, invalidSecondaryPurpose, invalidPurpose)
```
robertparkinson, line 237: console.log here

robertparkinson (changes requested):

robertparkinson (approved):

PR #1213, opened 2024-07-30
Persist a new return version for 'no returns required'

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4619

Having completed the return requirements setup journey (no-returns-required), we need to persist a return version (start date, reason and notes) to the database. There will be no return requirements in this scenario.

When "Approve" is clicked in the `/check` page the data in the session needs to be persisted to the `water.return_versions` table.


Cruikshanks (approved):

app/services/return-requirements/generate-return-version.service.js
```diff
@@ -21,7 +21,7 @@ const ReturnVersionModel = require('../../models/return-version.model.js')
  */
 async function go (sessionData, userId) {
   const returnVersion = await _generateReturnVersion(sessionData, userId)
-  const returnRequirements = await GenerateReturnVersionRequirementsService.go(sessionData.licence.id, sessionData.requirements)
+  const returnRequirements = await _generateReturnRequirements(sessionData)
```
Beckyrose200, line 24: I think the top comment might need updating to explain that there is a no-returns required route, but other than that all good

Jozzey: I've explained that on line 43 when it comes to generating the requirements. I can't see any need to add it to the top comment as well

PR #1217, opened 2024-07-31
Change 'Transfer licence' reason for 'No returns needed' journey

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4612

A decision has been made to change the 'Transfer licence' option for the 'no returns needed' page to 'Licence conditions do not require returns'. This also includes changing the return version reason being saved in the database as 'transfer-licence' to 'licence-conditions-do-not-require-returns'.

This PR is focused on amending the option from 'Transfer licence' to 'Licence conditions do not require returns'.


Cruikshanks (approved):
Just a tweak to the PR and merged title as the current one makes it sound like you are making all the reasons changes in this one change.

PR #1219, opened 2024-07-31
Add insert on conflict to seeders

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4535

As we move to using seeders for tests we can look to add functionality to allow us to seed the actually database.

This change adds ON CONLICT resoultion based on unique constraints that are not the primary id's. This will allow the seeders to seed data and not affect existing data.

This change is part of ongoing work to with the import service and test improvements to allow us to move to parallel test runs.


jonathangoulding:
uidx_code_subcode seems to have missed during the migration. Need to discuss with the team.

Cruikshanks (approved):
Let's get these in and look at regions separately.

robertparkinson (approved):

PR #1220, opened 2024-07-31
Deleted agreements shouldn't show in the setup tab

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4604

Currently, when an agreement is deleted within the licence set up tab, the agreement still remains. This is because the agreement is being 'soft deleted' rather than being removed from the database completely by adding the current time and date to the 'deleted_at' column in the licence agreements table. The logic for pulling licence agreements is missing instructions to only show licence agreements where the 'deleted_at' is null.

This PR is focused on not showing licence agreements in the licence set up tab where the licence has the 'deleted_at' column populated.


test/services/licences/fetch-agreements.service.test.js
```diff
@@ -30,24 +30,42 @@ describe('Fetch Agreements service', () => {
...
+          financialAgreement: { id: '970168ce-06c3-4823-b84d-9da30b742bb8', code: 'S127' }
+        }, { skip: ['id'] })
       })
     })
 
-    it('returns the matching agreements data', async () => {
-      const results = await FetchAgreementsService.go(licenceRef)
+    describe('and the agreement has been deleted', () => {
+      beforeEach(async () => {
+        await LicenceAgreementHelper.add(licenceAgreementData.deletedAt = new Date('2024-07-31 13:06:27.666'))
```
Jozzey, line 61: I wouldn't bother hard coding a date here as it doesn't really matter what the date is, just that it's not null.
```suggestion
        await LicenceAgreementHelper.add(licenceAgreementData.deletedAt = new Date())
```

test/services/licences/fetch-agreements.service.test.js
```diff
@@ -30,24 +30,42 @@ describe('Fetch Agreements service', () => {
...
+          financialAgreement: { id: '970168ce-06c3-4823-b84d-9da30b742bb8', code: 'S127' }
+        }, { skip: ['id'] })
       })
     })
 
-    it('returns the matching agreements data', async () => {
-      const results = await FetchAgreementsService.go(licenceRef)
+    describe('and the agreement has been deleted', () => {
+      beforeEach(async () => {
+        await LicenceAgreementHelper.add(licenceAgreementData.deletedAt = new Date())
```
Jozzey, line 61: Sorry, I should have spotted this the first time round. This test is not using your dummy `licenceAgreementData`. What you are actually doing here is using the default helper data and just adding an `endDate` to it. The test is passing because the record you are creating has a different licence reference. I would just put separate `licenceAgreementData` in the `beforeEach` for the "deleted" and " not deleted" tests.

Jozzey (changes requested):

test/services/licences/fetch-agreements.service.test.js
Jozzey: Actually I don't think this is even adding an `endDate`

rvsiyad: Ahh makes sense, thanks 👍🏾

Jozzey (changes requested):
Spotted another issue you have with the tests on this. We are no longer adding `await DatabaseSupport.clean()` to any of our unit tests unless we really need to so this needs to be removed from your tests. When you do that your tests are going to break. This is because the `financialAgreement` id is hardcoded. There is no need to hardcode any of the `FinancialAgreementHelper` data, it will generate random data for this which is fine for these tests.

You also have the `licenceRef` hardcoded. You should let the`LicenceAgreementHelper` generate this and have it declared as a `let` rather than a `const` and then update the `licenceRef` variable when you generate the record with the helper.

Let me know if you want me to go through this with you.

rvsiyad:
> Spotted another issue you have with the tests on this. We are no longer adding `await DatabaseSupport.clean()` to any of our unit tests unless we really need to so this needs to be removed from your tests. When you do that your tests are going to break. This is because the `financialAgreement` id is hardcoded. There is no need to hardcode any of the `FinancialAgreementHelper` data, it will generate random data for this which is fine for these tests.
> 
> You also have the `licenceRef` hardcoded. You should let the`LicenceAgreementHelper` generate this and have it declared as a `let` rather than a `const` and then update the `licenceRef` variable when you generate the record with the helper.
> 
> Let me know if you want me to go through this with you.

Think i've got it working fine now, let me know if i should change the names for the variables to have a Data prefix for clarity e.g. `let financialAgreementData`

test/services/licences/fetch-agreements.service.test.js
```diff
@@ -8,46 +8,62 @@ const { describe, it, beforeEach } = exports.lab = Lab.script()
...
-  const licenceRef = '01/12/34/1000'
-
-  beforeEach(async () => {
-    await DatabaseSupport.clean()
-  })
+  let licenceAgreement
+  let financialAgreement
 
   describe('when the licence has agreements data', () => {
     beforeEach(async () => {
```
Jozzey, line 22: Having this as a `beforeEach` will cause it to create 2 records, one for each test. You only really need one Financial Agreement record for these tests so you can change this to a `before`. You'll also need to add `before` to line 7.
```suggestion
    before(async () => {
```

test/services/licences/fetch-agreements.service.test.js
```diff
@@ -8,46 +8,62 @@ const { describe, it, beforeEach } = exports.lab = Lab.script()
...
-      const financialAgreement = await FinancialAgreementHelper.add({
-        id: '970168ce-06c3-4823-b84d-9da30b742bb8',
-        code: 'S127'
+      financialAgreement = await FinancialAgreementHelper.add()
+    })
+
+    describe('and the agreement has not been deleted', () => {
+      beforeEach(async () => {
+        licenceAgreement = await LicenceAgreementHelper.add({
+          endDate: new Date('2040-05-01'),
```
Jozzey, line 29: `endDate`, `startDate` & `signedOn` are reused in all the tests. You could declare these all as consts above the lets on line 18.  You can then update your code where these are used so that it looks more like this:
`        licenceAgreement = await LicenceAgreementHelper.add({
          endDate,
          financialAgreementId: financialAgreement.id,
          startDate,
          signedOn
        })`

Jozzey (changes requested):
Sorry, Spotted another couple of things while on a call. That should be it this time though!

Up to you if you want to rename the variables to things like `let financialAgreementData`. I'm fine with them as they are.

Jozzey (approved):

PR #1222, opened 2024-08-01
Remove DatabaseSupport.clean() from bill-runs

Cruikshanks:

We had a pattern of calling `await DatabaseSupport.clean()` in a `beforeEach()` in any unit test file that depended on adding data to support its tests.

The problem is that the project is becoming so large that it is no longer sustainable. It is also a blocker to moving to a test framework that supports [ECMAScript Modules](https://nodejs.org/api/esm.html), [the official standard format](https://tc39.github.io/ecma262/#sec-modules) to package JavaScript code. Unit test frameworks other than [Hapi Lab](https://hapi.dev/module/lab/) run tests in parallel by default. This means you can't have one test running DB clean while another is trying to load data.

But there are always those where it'll be some time before we have cause to look at them again. So, rather than waiting, this change handles updating all tests in `test/services/bill-runs` to drop their dependence on `DatabaseSupport.clean()`.

### DIFF
--- a/test/services/bill-runs/annual/fetch-billing-accounts.service.test.js
+++ b/test/services/bill-runs/annual/fetch-billing-accounts.service.test.js
@@ -4,7 +4,7 @@
 const Lab = require('@hapi/lab')
 const Code = require('@hapi/code')
 
-const { describe, it, beforeEach } = exports.lab = Lab.script()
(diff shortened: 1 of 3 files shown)


Cruikshanks:
Superseded by #1327

PR #1224, opened 2024-08-01
Return version can be on/after licence start date

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4599

Currently, when a user reaches the start date page for either the 'no returns required' or 'set up new requirements' journey, and enters a start date before the return version start date, the system errors. A fix needs to be put in place to allow users to enter a start date that is on or after the licence start date.

This PR will change the validation to allow users to enter a start that can proceed the return version start date but is on or after the licence start date.


app/validators/return-requirements/start-date.validator.js
```diff
@@ -62,12 +62,12 @@ function _validateAnotherStartDate (payload, licenceStartDate, licenceEndDate) {
...
       .format(['YYYY-MM-DD'])
       .required()
-      .greater(licenceStartDate)
+      .min(licenceStartDate)
       .less(licenceEndDate || '9999-12-31')
       .messages({
         'date.base': 'Enter a real start date',
         'date.format': 'Enter a real start date',
-        'date.greater': 'Start date must be after the original licence start date',
+        'date.min': 'Start date must be after the original licence start date',
```
Cruikshanks, line 70: I know it is not in the ticket. But could we also update the error message to reflect the change.

```suggestion
        'date.min': 'Start date must be on or after the original licence start date',
```

Cruikshanks (changes requested):
Just a tweak to the error message.

app/validators/return-requirements/start-date.validator.js
rvsiyad: I knew I should've just made that change. Note to self: next time just do the common sense thing 🗒️ 🖊️

Cruikshanks (approved):

PR #1229, opened 2024-08-02
Use seeded purpose IDs in rtn req purpose helper

robertparkinson:

As part of the work to generate return logs we need to be able to write unit tests that cover actual scenarios. This change refactors the return requirement purposes helper to use the seeder reference data rather than generating random UUIDs. This will help build our confidence in the unit tests by using actual values.


Cruikshanks (changes requested):
I'm fine with the code changes. But could we improve the title and correct the link?

So, it needs to be an actual link to the ticket, https://eaflood.atlassian.net/browse/WATER-4543, and have the correct reference.

Also, the title makes it sound like we are doing general structural changes, whereas what we are actually doing is moving from random IDs to seeded ones.

So, perhaps

> Use seeded purpose IDs in rtn req purpose helper

(within the 50 char limit and is capitilised)

Then just need to ensure this title and corrected link are used when the squash merge happens

Cruikshanks (approved):

PR #1233, opened 2024-08-05
Fix region seeder name

jonathangoulding:
A naught fella has been left in the code. This should be removed

### DIFF
--- a/test/support/seeders/regions.seeder.js
+++ b/test/support/seeders/regions.seeder.js
@@ -21,7 +21,7 @@ async function seed () {
               UPDATE public.regions
               SET charge_region_id = '${data[index].chargeRegionId}',
                   nald_region_id   = '${data[index].naldRegionId}',
-                  name             = 'fella',
+                  name             = '${data[index].name}',
                   display_name     = '${data[index].displayName}',
                   updated_at       = now()
               WHERE charge_region_id = '${data[index].chargeRegionId}'


Cruikshanks:
Unless it is urgent, this was spotted and will be resolved by https://github.com/DEFRA/water-abstraction-system/pull/1230

jonathangoulding:
> Unless it is urgent, this was spotted and will be resolved by #1230

This should be breaking tests - but i think they are testing the display name and not the name. So maybe not urgent if the next work is going in.

Beckyrose200 (commented):
Can we change it for the test region, so it sets the `isTest` column to true? The tear-down service for the acceptance tests utilises the test region `isTest` property to know what data to delete and with the seeder test region the tear-down is breaking

jonathangoulding:
> Can we change it for the test region, so it sets the `isTest` column to true? The tear-down service for the acceptance tests utilises the test region `isTest` property to know what data to delete and with the seeder test region the tear-down is breaking

You would need to update the data file and columns. Maybe set the default to false allow the data file to override. Might be worth a check on the PR mentioned if that's considered this.

Cruikshanks:
Parking this as we have https://github.com/DEFRA/water-abstraction-system/pull/1230 incoming.

It also deals with @Beckyrose200 question about flagging the region as `is_test`.

Cruikshanks:
Closing as covered by #1230

PR #1236, opened 2024-08-06
Removing licence from annual 2pt flags for supp billing

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4590

When a licence is removed from an annual two-part tariff bill run, it must be flagged to be included in the supplementary two-part bill run. This PR sets the foundation for flagging a licence. Currently, on the licence table, we have a supplementary bill run flag used for pre-sroc supplementary bill runs. However, with a 2pt sroc one we need to be able to store the year that the licence needs to be included.

The solution is a new table called licenceSupplementaryYears. This table will contain details of the `licenceId`, `billRunId` (if its been picked up by a billRun), the `financialYearEnd` and if it is 2pt. We have the 2pt column to allow us in the future to change existing supplementary flagging for non-2pt bill runs. This PR is specifically looking at when a licence is removed from an annual bill run, and the details then get persisted in our new table.

It also looks at the scenario for when a licence is removed from an annual 2pt bill run but then the bill run is cancelled and created again.


app/services/bill-runs/two-part-tariff/submit-remove-bill-run-licence.service.js
```diff
@@ -27,7 +28,9 @@ const ReviewLicenceModel = require('../../../models/review-licence.model.js')
 async function go (billRunId, licenceId, yar) {
   await RemoveReviewDataService.go(billRunId, licenceId)
 
-  const licenceRef = await _flagForSupplementaryBilling(licenceId)
+  await _flagForSupplementaryBilling(licenceId, billRunId)
+
+  const licenceRef = await _fetchLicenceRef(licenceId)
```
Cruikshanks, line 33: Can we move this down into the `if (!allLicencesRemoved) {` check?

There is no point fetching it if we are not going to use it. Plus it makes it clearer that this information is needed for the banner message.

app/services/bill-runs/two-part-tariff/submit-remove-bill-run-licence.service.js
```diff
@@ -53,15 +56,27 @@ async function _allLicencesRemoved (billRunId) {
...
 
   return licence.licenceRef
 }
 
+async function _flagForSupplementaryBilling (licenceId, billRunId) {
+  const financialYearEnd = await BillRunModel.query()
+    .findById(billRunId)
+    .returning('toFinancialYearEnding')
+
+  await LicenceSupplementaryYearModel.query()
```
Cruikshanks, line 72: As this is awaited above and there are no statements that follow it in the function, we should instead

```suggestion
  return LicenceSupplementaryYearModel.query()
```

Cruikshanks (changes requested):
There are just a couple of tweaks I'd suggest in the code. Other than that, this should get things rolling for 2PT supplementary!

Cruikshanks (approved):

PR #1239, opened 2024-08-07
Create SROC two-part tariff flagging endpoint for new charge versions

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4622
https://eaflood.atlassian.net/browse/WATER-4588

This is a part of the work for two-part tariff supplementary billing. A licence can be added to a two-part tariff supplementary bill run in 5 ways. Editing a return, adding a historic charge version, removing a licence from the annual two-part tariff bill run, recalculating a bill and if a licence is lapsed or revoked. 
Out of these 5 ways, 4 of them are in the legacy code base. To keep the new code for the supplementary billing within our system code base, we have decided to create a new endpoint that the legacy code can hit to calculate if the changed licence needs to be added to a supplementary bill run. This PR sets up the endpoint and flags a licence in need of a supplementary bill run when a historic charge version is added.


app/controllers/licences.controller.js
```diff
@@ -91,9 +92,16 @@ async function viewReturns (request, h) {
   })
 }
 
+async function submitSupplementaryFlag (request, h) {
+  await CheckSupplementaryBillingFlagService.go(request.payload)
+
+  return h.response().code(204)
+}
```
Cruikshanks, line 99: It would be worth treating this as a background process.

By this, I mean the legacy service does not have to `await` the flagging to complete. For an example, check the controllers (and their respective backing services) in `/jobs`. Or ping me if you need further info.

app/controllers/licences.controller.js
```diff
@@ -91,9 +92,16 @@ async function viewReturns (request, h) {
   })
 }
 
+async function submitSupplementaryFlag (request, h) {
```
Cruikshanks, line 95: We've been reserving `submit` for when a web form is submitted to us rather than for _all_ `POST` requests. So, we can rename this to match the endpoint directly, `supplementaryFlag()`.

app/services/licences/check-supplementary-billing-flag.service.js
```diff
@@ -0,0 +1,72 @@
...
+
+    const years = await SupplementaryBillingYearsService.go(startDate, endDate)
+
+    if (years) {
+      await FlagSupplementaryBillingService.go(licence, years)
+    }
+  }
+}
+
+async function _getChargeVersion (chargeVersionId) {
```
Cruikshanks, line 45: Our convention has been to name these `_fetch`.

```suggestion
async function _fetchChargeVersion (chargeVersionId) {
```

app/services/licences/check-supplementary-billing-flag.service.js
```diff
@@ -0,0 +1,72 @@
...
+ * in the charge version.
+ *
+ * If there are years that have been affected, these are then passed to our `FlagSupplementaryBillingService`, which
+ * checks if a bill run has already been sent for that affected year, and if it has it will then persist the flag in the
+ * `LicenceSupplementaryYear` table.
+ *
+ * @param {Object} payload - The payload from the request to be validated
+ */
+async function go (payload) {
+  if (payload.chargeVersionId) {
```
Cruikshanks, line 29: If we flip this, we can immediately return in the guard clause style.

```javascript
  if (!payload.chargeVersionId) {
    return
  }
```

app/services/licences/flag-supplementary-billing.service.js
```diff
@@ -0,0 +1,64 @@
...
+    await _persistSupplementaryBillingYears(licence.id, yearsForSupplementaryBilling)
+  }
+}
+
+/**
+ * We need to verify which years annual two-part tariff bill runs have been sent. A year shouldn't be flagged for a
+ * supplementary bill run if the annual bill run hasn't been sent yet, as any licence changes will be handled in the
+ * annual run.
+ */
+async function _getSupplementaryBillingYears (regionId, years) {
```
Cruikshanks, line 32: Our convention is to try and avoid prefixing functions with `get`, `set` and `is`. If a function if returning a thing we can just name it as that.

```suggestion
async function _supplementaryBillingYears (regionId, years) {
```

app/services/licences/flag-supplementary-billing.service.js
```diff
@@ -0,0 +1,64 @@
...
+      .where('batchType', 'two_part_tariff')
+      .where('status', 'sent')
+      .where('toFinancialYearEnding', year)
+
+    if (annualTwoPartTariff.length > 0) {
+      annualTwoPartTariffYears.push(year)
+    }
+  }
+
+  return annualTwoPartTariffYears
```
Cruikshanks, line 48: It's a minor performance tweak but still worth making IMHO.

We should be able to fetch all bill runs using `.whereIn()` in one go. I think if you then amend the select to be a `distinct('toFinancialYearEnding')` the result can be directly used in place of `annualTwoPartTariffYears`.

app/services/licences/supplementary-billing-years.service.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  return years
+}
+
+/**
+ * When flagging a licence for the supplementary bill run, we need to consider which financial years have been
+ * impacted by the change. We only care about the financial year ends. So if the startDate for a new chargeVersion is
+ * `2022-05-31`, the financial year end is considered to be `2023` since the financial years run April to March. Same
+ * goes for if a charge versions endDate is `2024-04-05`, the financial year end is `2025`.
```
Cruikshanks, line 47: This is a helpful comment!

My suggestion, though, would be to alter the second example to `2023-03-05`. I think it helps reinforce your point that the financial year is April to March.

app/services/licences/supplementary-billing-years.service.js
```diff
@@ -0,0 +1,61 @@
+'use strict'
+
+/**
+ * Works out which financial year ends are between the date ranges
+ * @module SupplementaryBillingYearsService
+ */
+
+const MONTHS_UNTIL_FINANCIAL_YEAR_START = 3
```
Cruikshanks, line 8: It's my own personal opinion. But we're testing if the month of a date is equal to or greater than April (assuming I'm not mistaken).

So, if we renamed this to `APRIL` do you think it would make where we use it a little clearer?

```javascript
  if (date.getMonth() >= APRIL) {
    // ..
  }
```

Cruikshanks (changes requested):
Let me know if there are any issues, questions or challenges to these suggestions. All are welcome!

Cruikshanks (changes requested):
Next tranche of feedback! 😁

app/services/licences/process-supplementary-billing-flag.service.js
```diff
@@ -0,0 +1,78 @@
...
+      return
+    }
+
+    const years = await DetermineSupplementaryBillingYearsService.go(startDate, endDate)
+
+    if (years) {
+      await FlagSupplementaryBillingService.go(licence, years)
+    }
+  } catch (error) {
+    global.GlobalNotifier.omfg('Supplementary Billing Flag failed', null, error)
```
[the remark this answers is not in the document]
Beckyrose200: Possibly not because we fetch the licenceId ourselves it isn't passed to us so it could break before we've fetched it?

app/services/licences/process-supplementary-billing-flag.service.js
```diff
@@ -0,0 +1,78 @@
...
+
+    if (!twoPartTariffIndicator) {
+      return
+    }
+
+    const years = await DetermineSupplementaryBillingYearsService.go(startDate, endDate)
+
+    if (years) {
+      await FlagSupplementaryBillingService.go(licence, years)
+    }
```
Cruikshanks, line 45: I think to be consistent with the 2 guard clauses above, we could tweak this section.

```suggestion
    if (!years) {
      return
    }

    await FlagSupplementaryBillingService.go(licence, years)
```

app/services/licences/process-supplementary-billing-flag.service.js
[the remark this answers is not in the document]
Cruikshanks: @Beckyrose200 is right, if the fetch fails we'll have nothing to pass through. _But_ we do have the payload. I would suggest passing that through as the `data` arg to omfg(). That way we'll know the details of what 'thing' to start looking at should it be a data issue.

test/services/licences/determine-supplementary-billing-years.service.test.js.js
```diff
@@ -0,0 +1,121 @@
...
+      clock.restore()
+    })
+
+    describe('and no end date', () => {
+      beforeEach(() => {
+        testDate = new Date('2024-03-31')
+        clock = Sinon.useFakeTimers(testDate)
+
+        endDate = null
+      })
```
Cruikshanks, line 39: We repeat this same section in the next section of tests. So, in effect it applies to _all_ unit tests in this file.

On that basis, I'd suggest moving this to a top-level `before()`, and the `clock.restore()` along with the `Sinon.restore()` into a top-level `after()`.

app/services/licences/process-supplementary-billing-flag.service.js
```diff
@@ -0,0 +1,78 @@
...
+      return
+    }
+
+    const years = DetermineSupplementaryBillingYearsService.go(startDate, endDate)
+
+    if (years) {
+      await FlagSupplementaryBillingService.go(licence, years)
+    }
+  } catch (error) {
+    global.GlobalNotifier.omfg('Supplementary Billing Flag failed', error)
```
Cruikshanks, line 47: This will break the notifier. `omfg()` expects 3 args. error is the last one, so if we want to pass it through we _must_ provide a second arg for data (else error is treated as the data arg by the function).

What I would actually suggest is you pass the `payload` as the second arg. You were right in that we couldn't use whatever the fetch returned because it might be the fetch erroring. But the payload is always available and having the details of _what_ triggered the call would be extremely useful when trying to resolve any future error/issue.

Cruikshanks (changes requested):

test/services/licences/supplementary/determine-charge-version-years.service.test.js
```diff
@@ -0,0 +1,77 @@
...
+          expect(result.twoPartTariff).to.equal(false)
+        })
+      })
+
+      describe('and a charge reference with two-part tariff indicators', () => {
+        beforeEach(async () => {
+          await ChargeReferenceHelper.add({ chargeVersionId: chargeVersion.id, adjustments: { s127: true } })
+        })
+
+        it('flagForBilling and twoPartTariff as true', async () => {
```
Cruikshanks, line 55:
```suggestion
        it('returns flagForBilling and twoPartTariff as true', async () => {
```

test/services/licences/supplementary/process-billing-flag.service.test.js
```diff
@@ -0,0 +1,144 @@
...
+              id: licence.id,
+              regionId: licence.regionId
+            },
+            startDate: chargeVersion.startDate,
+            endDate: chargeVersion.endDate,
+            twoPartTariff: true,
+            flagForBilling: true
+          })
+
+          await BillRunHelper.add({ batchType: 'two_part_tariff', status: 'sent', regionId: licence.regionId })
```
Cruikshanks, line 66: Rather than add a bill run to the DB, I would suggest stubbing `DetermineBillRunsSentService` (or whatever it ends up being called!)

Cruikshanks (changes requested):
Hopefully just some minor tweaks plus the change we discussed.

Cruikshanks (approved):

PR #1240, opened 2024-08-07
Create `chargeVersionNotes` table model, helper and view

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4566

While creating the tests for the [Licence History page](https://github.com/DEFRA/water-abstraction-system/pull/1182), we noticed we are missing the model and helper for the notes table. This is being referenced when extracting charge versions for a licence and any relating notes created. As a result, while creating the tests for the `fetch-licence-history.service` database errors occurred as a result of the unknown relation between the notes and charge versions table.

As well as adding the model and helper for the chargeVersionNotes table, the unit tests will also require the creation of the chargeVersionNotes table in the test database. This also gives us the opportunity to add the view for the chargeVersionNotes table to the public schema.

This PR will add the model, helper and the view for the `notes` table in the water schema.


test/models/charge-version-note.model.test.js
```diff
@@ -0,0 +1,62 @@
...
+        await UserHelper.add({ id: testRecord.userId })
+      })
+
+      it('can successfully run a related query', async () => {
+        const query = await ChargeVersionNoteModel.query()
+          .withGraphFetched('user')
+
+        expect(query).to.exist()
+      })
+    })
```
Cruikshanks, line 60: We seem to be missing the eager load tests ?? 😕 

Ping me if you are unsure what I mean by this (or check one of the other model tests).

Cruikshanks (changes requested):
Just seem to be some tests got overlooked/missed.

test/models/charge-version-note.model.test.js
```diff
@@ -0,0 +1,92 @@
...
+        expect(result.chargeVersion).to.equal(testChargeVersion)
+      })
+    })
+
+    describe('when linking to user', () => {
+      let testUser
+
+      beforeEach(async () => {
+        testUser = await UserHelper.add({ id: testRecord.userId })
+      })
```
Cruikshanks, line 70: We should be able to avoid creating a new user record by instead `select()` an existing one. Shout if you need any help/extra info about how to go about this.

Cruikshanks (changes requested):

test/models/charge-version-note.model.test.js
rvsiyad: Ahh completely forgot about that, implemented it now 👍🏾

Cruikshanks (approved):

PR #1241, opened 2024-08-07
Remove more DatabaseSupport.clean() from tests

jonathangoulding:

This change like all things initially started small changing  bills and licences services tests. As i moved on the return requirements most of them were straight forwards. However some complexity was added with the licence-abstraction-data.seeder.js. Some temporary code has been introduced to create the financial agreements if they don't exists (they are queried and used if they exists). This is necessary as some test are still cleaning the database. 

We had a pattern of calling `await DatabaseSupport.clean()` in a `beforeEach()` in any unit test file that depended on adding data to support its tests.

The problem is that the project is becoming so large that it is no longer sustainable. It is also a blocker to moving to a test framework that supports [ECMAScript Modules](https://nodejs.org/api/esm.html), [the official standard format](https://tc39.github.io/ecma262/#sec-modules) to package JavaScript code. Unit test frameworks other than [Hapi Lab](https://hapi.dev/module/lab/) run tests in parallel by default. This means you can't have one test running DB clean while another is trying to load data.

But there are always those where it'll be some time before we have cause to look at them again. So, rather than waiting, this change handles updating all tests in `test/services/bill-licences` to drop their dependence on `DatabaseSupport.clean()`.

### DIFF
--- a/test/services/bills/fetch-bill-summary.service.test.js
+++ b/test/services/bills/fetch-bill-summary.service.test.js
@@ -8,14 +8,13 @@ const { describe, it, beforeEach } = exports.lab = Lab.script()
 const { expect } = Code
 
 // Test helpers
-const BillingAccountHelper = require('../../support/helpers/billing-account.helper.js')
-const BillingAccountAddressHelper = require('../../support/helpers/billing-account-address.helper.js')
 const BillHelper = require('../../support/helpers/bill.helper.js')
 const BillLicenceHelper = require('../../support/helpers/bill-licence.helper.js')
 const BillRunHelper = require('../../support/helpers/bill-run.helper.js')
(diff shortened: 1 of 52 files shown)


jonathangoulding:
Tested with the seed and clean commented out to ensure the test will work generically.

jonathangoulding:
Biggest chages are in the licence-abstraction-data.seeder.js

[licence-abstraction-data.seeder.js](https://github.com/DEFRA/water-abstraction-system/pull/1241/files#diff-bbb9a220992041204cfcafc08b40739850cb144497082abf6b1963f66ceb1ea8)

Cruikshanks (approved):

Cruikshanks:
Sorry @jonathangoulding . I merged on your behalf because both this and my PR #1230 touch a lot of the same files. Having this merged means I can crack on and deal with the conflicts.

PR #1244, opened 2024-08-08
Generate return logs from requirements

robertparkinson:

As part of the work to replace the legacy system NALD we need to generate return logs. This PR creates a job that when run twice a year will create the return logs for the appropriate cycle, either summer or winter/all year.


app/routes/jobs.routes.js
```diff
@@ -58,6 +58,20 @@ const routes = [
         crumb: false
       }
     }
+  },
+  {
+    method: 'POST',
+    path: '/jobs/return-logs',
```
Cruikshanks, line 64: TBH, I'm a little uncomfortable with a feature such as this having a default cycle. Ideally, I'd like what cycle to be run to be explicit in the request.

Pragmatically, if I can avoid a payload when it comes to the Jenkins job that would also be handy.

So, I know this is a change to how I said to go, but could we change the route to be `jobs/return-logs/${cycle}`?

That way if someone pings a request without one it will hit a 404, which acts as an alternate form of validation. It also acts as a means of documenting that the job is cycle based.

We'll keep the `licenceRef` in the payload for now. Once we're happying this job is working correctly we'll be dropping it anyway.

Cruikshanks (changes requested):
To help keep things moving I'll flit between the two 🦣 PR's with quick bits of feedback and suggested changes.

So, I still need to go through the rest. But figured this is something that could be looked at whilst I take a chunk out of @jonathangoulding 's PR!

app/services/jobs/return-logs/process-return-logs.service.js
```diff
@@ -5,19 +5,22 @@
...
   try {
+    const startTime = currentTimeInNanoseconds()
     const returnLogs = await FetchReturnLogsService.go(isSummer, licenceReference)
 
     if (returnLogs.length > 0) {
       await CreateReturnLogsService.go(returnLogs)
     }
+    calculateAndLogTimeTaken(startTime, 'Process return logs complete', { isSummer, licenceReference })
   } catch (error) {
     global.GlobalNotifier.omfg('Create return logs job failed', null, error)
```
Cruikshanks: Ideally, the log messages should be the same. Using the pattern in other services.

```javascript
    calculateAndLogTimeTaken(startTime, 'Create return logs job complete', { isSummer, licenceReference })
  } catch (error) {
    global.GlobalNotifier.omfg('Create return logs job failed', { isSummer, licenceReference }, error)
```

app/services/jobs/return-logs/create-return-logs.service.js
Cruikshanks, line 1: I'm all for splitting responsibilities between services and saying they should focus on one thing.

But in this case, I think this logic could be moved into `ProcessReturnLogsService` as a function without impugning that idea or making the tests much more complicated to manage.

app/services/jobs/return-logs/process-return-logs.service.js
```diff
@@ -5,19 +5,22 @@
  * @module CreateReturnLogsService
```
Cruikshanks, line 5: The module comment needs to be updated to match the service name.

Cruikshanks (changes requested):
Next round of feedback. Good news is I've got through the whole PR now so this should be it!

app/services/jobs/return-logs/process-return-logs.service.js
```diff
@@ -1,31 +1,39 @@
...
+    const isSummer = cycle === 'summer'
     const startTime = currentTimeInNanoseconds()
     const returnLogs = await FetchReturnLogsService.go(isSummer, licenceReference)
 
     if (returnLogs.length > 0) {
-      await CreateReturnLogsService.go(returnLogs)
+      await _createReturnLogs(returnLogs)
     }
-    calculateAndLogTimeTaken(startTime, 'Process return logs complete', { isSummer, licenceReference })
+    calculateAndLogTimeTaken(startTime, 'Create return logs complete', { isSummer, licenceReference })
```
Cruikshanks, line 24: To be consistent it should be the following message. And ideally we'd log the the cycle so we can see what was requested rather than what was inferred.

```javascript
    calculateAndLogTimeTaken(startTime, 'Create return logs job complete', { cycle, licenceReference })
```

app/services/jobs/return-logs/process-return-logs.service.js
```diff
@@ -1,31 +1,39 @@
...
   } catch (error) {
     global.GlobalNotifier.omfg('Create return logs job failed', null, error)
```
Cruikshanks, line 26: Again, being consistent but also so we can diagnose if it was the cause, we'd log out the arguments here as well.

```javascript
    global.GlobalNotifier.omfg('Create return logs job failed', { cycle, licenceReference }, error)
```

test/services/jobs/return-logs/process-return-logs.service.test.js
```diff
@@ -25,7 +25,7 @@ describe('Process return logs service', () => {
   const allYearEndDate = new Date(new Date().getFullYear() + 1, 2, 31).toISOString().split('T')[0]
   const allYearStartDate = new Date(new Date().getFullYear(), 3, 1).toISOString().split('T')[0]
 
-  describe('Is summer is true and a licence reference is provided', () => {
+  describe('cycle is summer is true and a licence reference is provided', () => {
```
Cruikshanks, line 28: Perhaps

```javascript
    describe('cycle is "summer" and a licence reference is provided',
```

test/services/jobs/return-logs/process-return-logs.service.test.js
```diff
@@ -65,9 +65,9 @@ describe('Process return logs service', () => {
     })
   })
 
-  describe('Is summer is false and a licence reference is provided but there is no matching return requirements', () => {
+  describe('All year cycle and a licence reference is provided but there is no matching return requirements', () => {
```
Cruikshanks, line 68: Perhaps

```javascript
    describe('cycle is "all-year" and a licence reference is provided but there is no matching return requirements'
```

Cruikshanks (changes requested):
Sorry, there are a few more tweaks I can suggest.

app/services/jobs/return-logs/process-return-logs.service.js
```diff
@@ -11,19 +11,20 @@ const ReturnLogModel = require('../../../models/return-log.model.js')
 
 /**
  * Creates the return logs for the next cycle
+ * @param {string} cycle - the return cycle
+ * @param {string} licenceReference - The licence reference to have return logs created
```
Cruikshanks, line 15: I should have spotted it earlier but we should also update the `go()` signature to indicate that `licenceref` is optional

```javascript
async function go (cycle, licenceReference = null) {
```

```suggestion
 * The return requirement is the information held against the licence that defines how and when an abstractor needs to
 * submit their returns.
 *
 * The return log is the 'header' record generated each return cycle from the requirement that an abstractor submits
 * their returns against.
 *
 * When users make changes to return requirements, the service will determine if any new return logs need to be
 * created depending on the current cycle.
 *
 * But if no changes are ever made to a licence's return requirements, and this job didn't exist, no new return logs
 * would be created.
 *
 * So, this job will run twice yearly: once for each cycle. The job determines which return requirements need a return
 * log generated for the selected cycle and then creates them.
 *
 * > Because the job creates _all_ return logs in a cycle, it makes it difficult to test what it is generating is
 * > correct. So, to support testing and validation, we can pass a licence ref in the job request to limit the creation
 * > to just a single licence.
 * @param {string} cycle - the return cycle to create logs for (summer or all-year)
 * @param {string} [licenceReference] - An optional argument to limit return log creation to just the specific licence
```

Cruikshanks (changes requested):

Cruikshanks (approved):

PR #1246, opened 2024-08-08
Remove more DatabaseSupport.clean() from services/job

jonathangoulding:

We had a pattern of calling await DatabaseSupport.clean() in a beforeEach() in any unit test file that depended on adding data to support its tests.

The problem is that the project is becoming so large that it is no longer sustainable. It is also a blocker to moving to a test framework that supports ECMAScript Modules, the official standard format to package JavaScript code. Unit test frameworks other than Hapi Lab run tests in parallel by default. This means you can't have one test running DB clean while another is trying to load data.

But there are always those where it'll be some time before we have cause to look at them again. So, rather than waiting, this change handles updating all tests in test/services/bill-licences to drop their dependence on DatabaseSupport.clean().


jonathangoulding:
Appreciate this is a small one but the others in here where quite complex to tackle and i am going back to working on the import licence system tasks.

PR #1247, opened 2024-08-08
Mark Licence for Supplementary 2PT bill run after creating a historic return version

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4621

When a new return version is created we need to check to see if it has impacted any financial years that have already been included in a two-part tariff bill run.

If any financial years are impacted, then the licence needs to be flagged for a two-part tariff supplementary bill run for the years that a 2PT bill run exists.


Jozzey:
It has been realised that updating a return version should not add the licence to supplementary billing. This PR is therefore no longer required and can be closed.

PR #1248, opened 2024-08-08
Alan's Update prev. ret. vers. post new version

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4568

A couple of suggested alternative solutions to the one presented in [Update existing return versions when a new one is added](https://github.com/DEFRA/water-abstraction-system/pull/1216).

1. It avoids the need to first extract _all_ return versions and then identify which previous versions to update 'in code'. The checks are done within the queries sent to the database. In most cases, a single query will be fired at the database. But there is a possibility of making 4 calls to the database

2. As currently, it extracts _all_ return versions from the database and then identifies which version to update 'in code'. But it breaks the work down rather than doing it all in one large method.


Cruikshanks:
We went with solution 2 and made the change in the original PR. This one can now be closed.

PR #1255, opened 2024-08-13
Amend import licence endpoint to work in background

Cruikshanks:

> Part of the work to replace the import service logic to import a licence from NALD

This change updates the endpoint and backing service to allow the import to happen in the background.

It follows the same pattern we have used for other background tasks.


Cruikshanks:
Do not believe we need this now

PR #1256, opened 2024-08-14
Add create licence supplementary years service

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4588
https://github.com/DEFRA/water-abstraction-system/pull/1236

The new two-part tariff supplementary bill run requires licences to be recorded in our new LicenceSupplementaryYears table, along with the years they affect. The first ticket we picked up for this new process was amending the existing annual two-part tariff review pages. A licence can be removed from the bill run there and this should be flagged. This was added to the existing review page journey. Since starting work on the other flagging tickets it became clear that the process of adding the licence into our LicenceSupplementaryYears table is going to be repeated if we don't pull this out into its own service. So that is what this PR is for, making a service just for creating the licence record that can be shared amongst the various ways a licence can be flagged.

### DIFF
--- a/app/services/licences/supplementary/create-licence-supplementary-year.service.js
+++ b/app/services/licences/supplementary/create-licence-supplementary-year.service.js
@@ -1,0 +1,50 @@
+'use strict'
+
+/**
+ * Creates a licenceSupplementaryYears record based on the provided licence data
+ * @module CreateLicenceSupplementaryYearService
+ */
+
+const LicenceSupplementaryYearModel = require('../../../models/licence-supplementary-year.model.js')
+
+/**
+ * Creates a licenceSupplementaryYears record based on the provided licence data
+ *
+ * @param {module:LicenceModel} licenceId - The UUID of the licence to be persisted
+ * @param {Object[]} years - An array of the years to be persisted as individual records
+ * @param {Boolean} twoPartTariff - If there are any two-part tariff indicators on the licence
+ */
+async function go (licenceId, years, twoPartTariff) {
+  for (const year of years) {
+    const existingLicenceSupplementaryYears = await _fetchExistingLicenceSupplementaryYears(
+      licenceId, year, twoPartTariff
+    )
+
+    // Create a new record only if no existing record matches all the provided properties, and where 'billRunId' is null
+    if (existingLicenceSupplementaryYears.length === 0) {
+      await _persistSupplementaryBillingYearsData(licenceId, year, twoPartTariff)
+    }
+  }
+}
+
+async function _fetchExistingLicenceSupplementaryYears (licenceId, year, twoPartTariff) {
+  return LicenceSupplementaryYearModel.query()
+    .select('id')
+    .where('licenceId', licenceId)
+    .where('financialYearEnd', year)
+    .where('twoPartTariff', twoPartTariff)
+    .where('billRunId', null)
+}
+
+async function _persistSupplementaryBillingYearsData (licenceId, year, twoPartTariff) {
+  return LicenceSupplementaryYearModel.query()
+    .insert({
+      licenceId,
+      financialYearEnd: year,
+      twoPartTariff
+    })
+}
+
+module.exports = {
+  go
+}
--- a/app/services/licences/supplementary/create-licence-supplementary-year.service.js
+++ b/app/services/licences/supplementary/create-licence-supplementary-year.service.js
@@ -11,36 +11,38 @@ const LicenceSupplementaryYearModel = require('../../../models/licence-supplemen
  * Creates a licenceSupplementaryYears record based on the provided licence data
  *
  * @param {module:LicenceModel} licenceId - The UUID of the licence to be persisted
- * @param {Object[]} years - An array of the years to be persisted as individual records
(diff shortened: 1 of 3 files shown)


Beckyrose200:
Just a note, on the service I have resorted to creating a private `fetch` function to look for any existing records with the same properties that we want to persist. We did discuss using a where clause on the query insert, but after some research, I don't think this is possible due to the where clause only being used for 'selects', 'updates', and 'deletes'. I could not find any examples in our code that have a where clause on an insert.

I also looked into the option of using `onConflict` however, this again was a no-go due to the fact the database itself needs to have constraints on the table for the `onConflict` to be triggered. Any suggestions on this let me know :)

Cruikshanks:
> Just a note, on the service I have resorted to creating a private `fetch` function to look for any existing records with the same properties that we want to persist. We did discuss using a where clause on the query insert, but after some research, I don't think this is possible due to the where clause only being used for 'selects', 'updates', and 'deletes'. I could not find any examples in our code that have a where clause on an insert.
> 
> I also looked into the option of using `onConflict` however, this again was a no-go due to the fact the database itself needs to have constraints on the table for the `onConflict` to be triggered. Any suggestions on this let me know :)

My bad. It can be done in raw SQL or by dropping down to Knex, but it would be pretty messy. We don't need to optimise this, so don't worry about it!

app/services/licences/supplementary/create-licence-supplementary-year.service.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+}
+
+async function _fetchExistingLicenceSupplementaryYears (licenceId, year, twoPartTariff) {
+  return LicenceSupplementaryYearModel.query()
+    .select('id')
+    .where('licenceId', licenceId)
+    .where('financialYearEnd', year)
+    .where('twoPartTariff', twoPartTariff)
+    .where('billRunId', null)
```
Cruikshanks, line 36: We can simplify things a bit where the result is used by making the query only return a single `result` rather than `[result]`.

You could then do something like this in your `go()`.

```javascript
    const match = await _fetchExistingLicenceSupplementaryYears(
      licenceId, year, twoPartTariff
    )

    // Skip if an existing record for this licence and year (and no bill run ID) exists 
    if (match) {
      continue
    }

    await _persistSupplementaryBillingYearsData(licenceId, year, twoPartTariff)
```

```suggestion
  return LicenceSupplementaryYearModel.query()
    .select('id')
    .where('licenceId', licenceId)
    .where('financialYearEnd', year)
    .where('twoPartTariff', twoPartTariff)
    .where('billRunId', null)
    .limit(1)
    .first()
```

app/services/licences/supplementary/create-licence-supplementary-year.service.js
```diff
@@ -0,0 +1,50 @@
...
+}
+
+async function _persistSupplementaryBillingYearsData (licenceId, year, twoPartTariff) {
```
Cruikshanks, line 39: Not a hill I'm going to die on. But if the arg is actually the financial year end, would it help to just call it that?

```javascript
async function _persistSupplementaryBillingYearsData (licenceId, financialYearEnd, twoPartTariff) {
  return LicenceSupplementaryYearModel.query()
    .insert({
      licenceId,
      financialYearEnd,
      twoPartTariff
    })
}
```

test/services/licences/supplementary/create-licence-supplementary-year.service.test.js
```diff
@@ -0,0 +1,93 @@
...
+    })
+
+    describe('that does not already exist', () => {
+      it('persists the data', async () => {
+        await CreateLicenceSupplementaryYearService.go(licenceId, years, twoPartTariff)
+
+        const result = await _fetchLicenceSupplementaryYears(licenceId)
+
+        expect(result).to.have.length(1)
+        expect(result[0].licenceId).to.equal(licenceId)
```
Cruikshanks, line 37: I would suggest checking all the fields. Else, we have nothing confirming the year got persisted (I think) in these tests.

```suggestion
        expect(result[0]).to.equal({
          licenceId,
          billRunId: null,
          financialYearEnd: years[0],
          twoPartTariff
        }, { skip: ['id'] })
```

Cruikshanks (changes requested):
Just some suggested tweaks.

app/services/licences/supplementary/create-licence-supplementary-year.service.js
```diff
@@ -0,0 +1,52 @@
...
+ * @module CreateLicenceSupplementaryYearService
+ */
+
+const LicenceSupplementaryYearModel = require('../../../models/licence-supplementary-year.model.js')
+
+/**
+ * Creates a licenceSupplementaryYears record based on the provided licence data
+ *
+ * @param {module:LicenceModel} licenceId - The UUID of the licence to be persisted
+ * @param {Object[]} years - An array of the financial year ends to be persisted as individual records
```
Cruikshanks, line 14: Ooh, really minor one!

```suggestion
 * @param {Object[]} financialYearEnds - An array of the financial year ends to be persisted as individual records
```

Cruikshanks (changes requested):

Cruikshanks (approved):

PR #1262, opened 2024-08-16
Trigger 2PT Supplementary Bill Run - Pt2

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4631

As we treat 2PT supplementary as a separate bill run we will need a new journey to trigger it.

The first part of this work was carried out in PR https://github.com/DEFRA/water-abstraction-system/pull/1259

In this PR the remainder of the journey to enable the triggering of a 2PT supplementary will be completed. This will include selecting the required financial year and triggering the bill run.


app/presenters/bill-runs/setup/year.presenter.js
```diff
@@ -2,21 +2,71 @@
...
+    { text: '2020 to 2021', value: '2021', checked: selectedYear === '2021' }
+  ]
+}
+
+async function _supplementaryFinancialYearsData (regionId, selectedYear) {
+  const supplementaryFinancialYearsData = []
+  const tptSupplementaryYears = await _tptSupplementaryYears(regionId)
+
+  if (tptSupplementaryYears.length > 0) {
+    tptSupplementaryYears.forEach((tptSupplementaryYear) => {
```
Beckyrose200, line 51: Can I make 1 small suggestion to make this a bit easier on the eyes 
```suggestion
    tptSupplementaryYears.forEach((tptSupplementaryYear) => {
      const { financialYearEnd } = tptSupplementaryYear

      supplementaryFinancialYearsData.push({
        text: `${financialYearEnd - 1} to ${financialYearEnd}`,
        value: financialYearEnd.toString(),
        checked: selectedYear === financialYearEnd.toString()
      })
    })
  }
```

app/presenters/bill-runs/setup/year.presenter.js
```diff
@@ -2,21 +2,71 @@
...
+        value: tptSupplementaryYear.financialYearEnd.toString(),
+        checked: selectedYear === tptSupplementaryYear.financialYearEnd.toString()
+      })
+    })
   }
+
+  return supplementaryFinancialYearsData
+}
+
+async function _tptSupplementaryYears (regionId) {
```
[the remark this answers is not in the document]
Jozzey: I didn't think we were supposed to be prefixing things like this with "fetch", I could be mistaken though. @Cruikshanks are we doing the "fetch" thing?

app/presenters/bill-runs/setup/year.presenter.js
```diff
@@ -2,23 +2,75 @@
...
+        checked: selectedYear === financialYearEnd.toString()
+      })
+    })
+  }
+
+  return supplementaryFinancialYearsData
+}
+
+async function _tptSupplementaryYears (regionId) {
+  return LicenceSupplementaryYearModel.query()
```
Beckyrose200, line 66: This is more of a question than a change but we don't fetch data in any of our other presenters, we do that in a separate service normally and then pass the fetched data to the presenter, is this the exception to that rule?

Jozzey: I have to use it in both the GET & POST requests which is why I popped it in the presenter 🤠 I'm going to move it out of the presenter into its own service.

app/services/bill-runs/setup/year.service.js
```diff
@@ -20,7 +21,12 @@ const SessionModel = require('../../../models/session.model.js')
  */
 async function go (sessionId) {
   const session = await SessionModel.query().findById(sessionId)
-  const formattedData = YearPresenter.go(session)
+
+  const regionId = session.region
+  const twoPartTariff = session.type.startsWith('two_part')
```
Beckyrose200, line 26: If the bill run types currently using this are
'two_part_tariff' (annual)
and
'two_part_tariff_supplementary"

And once the two_part_tariff annual bill runs are all caught up this bill run type won't be using the years service anymore. 
And the plan is for supplementary bill runs to use this eventually.

So eventually it's just:
'supplementary'
and
'two_part_tariff_supplementary"

Can we change this logic to be like this:

```suggestion
  const twoPartTariffSupplementary = session.type === 'two_part_supplementary'
```

app/services/bill-runs/setup/year.service.js
```diff
@@ -20,7 +21,12 @@ const SessionModel = require('../../../models/session.model.js')
...
+  const licenceSupplementaryYears = await FetchLicenceSupplementaryYearsService.go(regionId, twoPartTariff)
```
Beckyrose200, line 27:
```suggestion
  const licenceSupplementaryYears = await FetchLicenceSupplementaryYearsService.go(regionId, twoPartTariffSupplementary)
```

app/services/bill-runs/setup/submit-year.service.js
```diff
@@ -37,10 +38,20 @@ async function go (sessionId, payload) {
...
+    if (session.type === 'two_part_supplementary') {
+      return { goBackToBillRuns: true }
+    }
+
     return { setupComplete: ['2024', '2023'].includes(session.year) }
   }
 
-  const formattedData = YearPresenter.go(session)
+  const regionId = session.region
+  const twoPartTariff = session.type.startsWith('two_part')
```
Beckyrose200, line 51: Same comment for this as above

Beckyrose200 (commented):
Looks good some small comments

Beckyrose200 (approved):

PR #1269, opened 2024-08-20
Lint JSDOC's

jonathangoulding:
https://github.com/DEFRA/water-abstraction-team/issues/125

We have moved most of the code conventions into eslint. This change implements linting on our jsdocs to meet our coding standards and conventions.

This change will add an initial set of rules to help enforce our standards.

As the impact of these changes do no necessarily affect code quality they have been set to warn to give us time to gradually fix the issues raised.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -44,7 +44,8 @@ module.exports = {
     'jsdoc/newline-after-description': 'off', // does not work with 'use strict'
     'jsdoc/check-indentation': 'warn',
     'jsdoc/lines-before-block': 'warn',
-    'jsdoc/check-types': 'warn'
+    'jsdoc/check-types': 'warn',
+    'jsdoc/require-hyphen-before-param-description': 'warn'
   }
 }
 
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -43,7 +43,8 @@ module.exports = {
     'jsdoc/check-alignment': 'warn',
     'jsdoc/newline-after-description': 'off', // does not work with 'use strict'
     'jsdoc/check-indentation': 'warn',
-    'jsdoc/lines-before-block': 'warn'
+    'jsdoc/lines-before-block': 'warn',
+    'jsdoc/check-types': 'warn'
   }
 }
 
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,11 +2,13 @@
 
 module.exports = {
   extends: 'standard', // Maintain Standard.js rules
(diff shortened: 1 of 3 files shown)


.eslintrc.js
```diff
@@ -33,7 +35,15 @@ module.exports = {
...
-    strict: ['error', 'global']
+    strict: ['error', 'global'],
+    'jsdoc/require-description': 'warn',
+    'jsdoc/require-param': 'warn',
+    'jsdoc/require-returns': 'warn',
+    'jsdoc/check-tag-names': 'warn',
+    'jsdoc/check-alignment': 'warn',
+    'jsdoc/newline-after-description': 'off', // does not work with 'use strict'
+    'jsdoc/check-indentation': 'warn',
+    'jsdoc/lines-before-block': 'warn'
```
jonathangoulding, line 46: I don't think is necessary. 

Where we have something like this: 

```
  return {
      /**
       * currentVersion modifier fetches only the current licence version record for this licence
       */
```

So might be work leaving out.

.eslintrc.js
```diff
@@ -33,7 +35,16 @@ module.exports = {
...
+    strict: ['error', 'global'],
+    'jsdoc/require-description': 'warn',
+    'jsdoc/require-param': 'warn',
+    'jsdoc/require-returns': 'warn',
+    'jsdoc/check-tag-names': 'warn',
+    'jsdoc/check-alignment': 'warn',
+    'jsdoc/newline-after-description': 'off', // does not work with 'use strict'
+    'jsdoc/check-indentation': 'warn',
+    'jsdoc/lines-before-block': 'warn',
+    'jsdoc/check-types': 'warn'
```
jonathangoulding, line 47: This rule alone finds 744 problems. 

```
 Invalid JSDoc @param "licenceName" type "String"; prefer: "string" 
 Invalid JSDoc @returns type "Object"; prefer: "object"
Invalid JSDoc @param "index" type "Number"; prefer: "number" 
```

We can discuss the is this rule is needed or not.

jonathangoulding:
✖ 1546 problems (0 errors, 1546 warnings)
  0 errors and 1347 warnings potentially fixable with the `--fix` option.

After a fix there will be 216 problems to manually fix. With only 5 manual ERRORS that will need to be fixed.

.eslintrc.js
jonathangoulding: Discussed - this can be an inline comment.

Cruikshanks (approved):

PR #1273, opened 2024-08-20
Fix JSDOC's lint issues

jonathangoulding:
Previous work https://github.com/DEFRA/water-abstraction-system/pull/1269

Building on from previous work of using eslint to handle out code conventions this change will use the auto fix feature of eslint and fix any errors in linting that may occur.

Some of the more complex changes such as adding returns and parameters to the jsdocs will be left to be done as when the file is touched or when time permits addressing tech debt.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -39,13 +39,15 @@ module.exports = {
     'jsdoc/require-description': 'warn',
     'jsdoc/require-param': 'warn',
     'jsdoc/require-returns': 'warn',
-    'jsdoc/check-tag-names': 'warn',
+    'jsdoc/check-tag-names': ['warn', {
+      definedTags: ['json']
+    }],
     'jsdoc/check-alignment': 'warn',
     'jsdoc/newline-after-description': 'off', // does not work with 'use strict'
-    'jsdoc/check-indentation': 'warn',
     'jsdoc/lines-before-block': 'warn',
     'jsdoc/check-types': 'warn',
-    'jsdoc/require-hyphen-before-param-description': 'warn'
+    'jsdoc/require-hyphen-before-param-description': 'warn',
+    'jsdoc/check-indentation': ['warn', { excludeTags: ['json', 'example'] }]
   }
 }
 
--- a/app/controllers/import.controller.js
+++ b/app/controllers/import.controller.js
@@ -7,6 +7,8 @@ const LegacyImportLicenceService = require('../services/import/legacy-licence.se
 
 /**
  * Controller for /import
+ * @param request
+ * @param h
  * @module ImportController
  */
 async function licence (request, h) {
--- a/app/controllers/import.controller.js
+++ b/app/controllers/import.controller.js
@@ -1,16 +1,15 @@
 'use strict'
 
+/**
+ * Controller for /import
+ * @module ImportController
+ */
+
 const Boom = require('@hapi/boom')
 
 const FeatureFlags = require('../../config/feature-flags.config.js')
 const LegacyImportLicenceService = require('../services/import/legacy-licence.service.js')
 
-/**
- * Controller for /import
- * @param request
- * @param h
- * @module ImportController
- */
 async function licence (request, h) {
   try {
     const { licenceRef } = request.payload
--- a/app/services/idm/fetch-user-roles-and-groups.service.js
+++ b/app/services/idm/fetch-user-roles-and-groups.service.js
@@ -20,10 +20,10 @@ const UserModel = require('../../models/user.model.js')
  *
  * @param {number} userId - The user id to get roles and groups for
  *
- * @returns {Promise<object>} result The resulting roles and groups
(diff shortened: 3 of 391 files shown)


app/controllers/import.controller.js
```diff
@@ -7,6 +7,8 @@ const LegacyImportLicenceService = require('../services/import/legacy-licence.se
 
 /**
  * Controller for /import
+ * @param request
```
jonathangoulding, line 10: Fix - this is missing a jdco for the controller

jonathangoulding:
Screenshot 2024-08-20 at 15 00 20

Potential issue with the indentation.

jonathangoulding:
Eslint fix has added the params where they might not exist for example - https://github.com/DEFRA/water-abstraction-system/pull/1273/files#diff-ca8af47304ddb7f1146579a0039ef262ff33c60cc545f883eafb95409cf4f4c7R54

This is only applied where jsdocs are present.

.eslintrc.js
```diff
@@ -39,13 +39,15 @@ module.exports = {
...
+      definedTags: ['json']
+    }],
     'jsdoc/check-alignment': 'warn',
     'jsdoc/newline-after-description': 'off', // does not work with 'use strict'
-    'jsdoc/check-indentation': 'warn',
     'jsdoc/lines-before-block': 'warn',
     'jsdoc/check-types': 'warn',
-    'jsdoc/require-hyphen-before-param-description': 'warn'
+    'jsdoc/require-hyphen-before-param-description': 'warn',
+    'jsdoc/check-indentation': ['warn', { excludeTags: ['json', 'example'] }]
```
jonathangoulding, line 50: This custom tag is breaking the docs due to this being set - 
```
    // We want JSDoc to error if it encounters an unknown tag
    allowUnknownTags: false,
```

We can either use the default @example tag everywhere or turn this on. 

The other option is to write a pulgin for jsdoc which is not necessary i feel.

app/services/idm/fetch-user-roles-and-groups.service.js
```diff
@@ -18,12 +18,12 @@ const UserModel = require('../../models/user.model.js')
...
  * of, along with `userFound` to explicitly indicate whether or not the user id exists.
  *
- * @param {Number} userId The user id to get roles and groups for
+ * @param {number} userId - The user id to get roles and groups for
  *
- * @returns {Promise<Object>} result The resulting roles and groups
- * @returns {UserModel} result.user Returns the UserModel representing the user, or `null` if the user is not found
- * @returns {RoleModel[]} result.roles An array of RoleModel objects representing the roles the user has
- * @returns {GroupModel[]} result.groups An array of GroupModel objects representing the groups the user is a member of
+ * @returns {Promise<object>} result - A promise that resolves to an object containing:
```
jonathangoulding, line 23: The linter enforce single returns.

This is the proposed soultion that is not a typedef.

Cruikshanks:
> Screenshot 2024-08-20 at 15 00 20
> 
> Potential issue with the indentation has been resovled by adding a tag called json.
> 
> Where this tag is used in jsdocs the rule where that tag applies is turned off. This is useful when we have defined json blocks inside the js docs but still want to maintain a geenral consistency across the code base with indentation.
> 
> Defined - https://github.com/DEFRA/water-abstraction-system/pull/1273/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5R43
> 
> Use - https://github.com/DEFRA/water-abstraction-system/pull/1273/files#diff-a4cc560e262bab2dfa7d2c2d2c2325e990ae57be714ea5d2e7cd6393c152f2cfR59
> 
> The example tag is a default that is also used, this has been used for return statements in the js doc - https://github.com/DEFRA/water-abstraction-system/pull/1273/files#diff-5ffd5b1dbdd1bacda168f372826e401555e476b8250a6b7df4538ef165afc337R66

I don't think this needs a custom tag. I was able to resolve the alignment issues without it. I've got that file down to this without making any changes to any linting rules.

Screenshot 2024-08-20 at 21 13 11

Happy to demonstrate when you are free (just ping me).

Cruikshanks:
I'd also suggest adding the following to `eslintrc.js`.

```javascript
  ,
  settings: {
    jsdoc: {
      mode: 'jsdoc',
      ignorePrivate: true
    }
  }
```

The plugin's default mode is Typescript. This ensures it knows we are just following JSDoc conventions.

It picked up issues with `app/services/bill-runs/supplementary/pre-generate-billing-data.service.js` that were in private methods. It's not that the issues are invalid. But I wanted to properly document only the public methods.

I want to encourage people to add comments to private methods, especially when they are gnarly. But I don't want them to be shouted at for missing params and return statements. To make this work, we'll just have to add `@private` to them in the future.

jonathangoulding:
Work has been done on separate branches to make the PR's smaller.

jonathangoulding:
All issues addressed on other PR's.

PR #1279, opened 2024-08-21
Fix JSDOC's lint required params and returns

jonathangoulding:
https://github.com/DEFRA/water-abstraction-team/issues/125

Previous work https://github.com/DEFRA/water-abstraction-system/pull/1269

Building on from previous work of using eslint to handle out code conventions this change will use the auto fix feature of eslint and fix any errors relating to required paramas and returns for private methods.

Some of the more complex changes such as adding returns and parameters to the jsdocs will be left to be done as when the file is touched or when time permits addressing tech debt.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -40,12 +40,12 @@ module.exports = {
     //   publicOnly: true
     // }],
     // 'jsdoc/require-description': 'warn',
-    // 'jsdoc/require-param': ['warn', {
-    //   exemptedBy: ['private']
-    // }],
-    'jsdoc/require-returns': ['warn', {
-      publicOnly: true
+    'jsdoc/require-param': ['warn', {
+      exemptedBy: ['private']
     }]
+    // 'jsdoc/require-returns': ['warn', {
+    //   publicOnly: true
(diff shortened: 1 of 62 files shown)


.eslintrc.js
```diff
@@ -36,12 +36,19 @@ module.exports = {
     'arrow-body-style': ['error', 'always'],
     'import/extensions': ['error', 'always'],
     strict: ['error', 'global'],
+    'jsdoc/require-jsdoc': ['warn', {
```
jonathangoulding, line 39: Makes sense to add this if we are adding no requirement for params / returns for private methods. 

Unfortunately params needs to use the @private tag but we can use  publicOnly: true on the required returns which is much easier.

Cruikshanks:
Does this help you @jonathangoulding ? https://github.com/DEFRA/water-abstraction-system/pull/1280

jonathangoulding:
> Does this help you @jonathangoulding ? https://github.com/DEFRA/water-abstraction-system/pull/1280 

That'll do it for the params.

Cruikshanks (approved):

PR #1285, opened 2024-08-23
Remove more DatabaseSupport.clean()

jonathangoulding:

We had a pattern of calling await DatabaseSupport.clean() in a beforeEach() in any unit test file that depended on adding data to support its tests.

The problem is that the project is becoming so large that it is no longer sustainable. It is also a blocker to moving to a test framework that supports ECMAScript Modules, the official standard format to package JavaScript code. Unit test frameworks other than Hapi Lab run tests in parallel by default. This means you can't have one test running DB clean while another is trying to load data.

But there are always those where it'll be some time before we have cause to look at them again. So, rather than waiting, this change handles updating all tests in test/services/bill-licences to drop their dependence on DatabaseSupport.clean().


test/services/jobs/licence-updates/fetch-licence-updates.service.test.js
```diff
@@ -70,7 +67,9 @@ describe('Fetch Licence Updates service', () => {
       })
 
       it('returns no results', async () => {
-        const results = await FetchLicenceUpdatesService.go()
+        const data = await FetchLicenceUpdatesService.go()
+
+        const results = data.filter((l) => { return l.licenceId === licence.id })
```
jonathangoulding, line 72: I am not too happy with this solution. But it does work. 

Maybe the tests descriptions could change to reflect that we are only concerned with the licence we have added for the test ?

Cruikshanks: 🎯 I think that is part of the change we need to make in our testing as it will only get harder to avoid clashes as we add more functionality.

When we cleaned before every test it made sense to have tests like, "I added some junk and when I run my service nothing is returned proving it works".

I think we need to move more towards an approach which is more like, "I added some junk and when I run my service _the junk is not returned_ proving it works". Or something! 😁

Cruikshanks (approved):

Cruikshanks:
First run 3 mins 32 secs. Last run 2 mins 51 seconds. (unit tests)

Every little helps!

PR #1294, opened 2024-08-29
Create return logs for newly imported licences

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4545

As part of the work to replace the legacy system NALD we need to generate return logs. This PR updates the logic of return log creation to allow for new licences to be imported and have their return logs created


app/services/jobs/return-logs/fetch-licence-return-requirements.service.js
```diff
@@ -0,0 +1,128 @@
...
+ * Given the licence reference this service returns the return requirements to be turned into return logs.
+ *
+ * @param {string} licenceReference - if provided only do the return log for that licence
+ *
+ * @returns {Promise<Array>} the array of return log payloads to be created in the database
+ */
+async function go (licenceReference) {
+  const returnRequirements = await _fetchReturnRequirements(licenceReference)
+
+  return returnRequirements
```
Cruikshanks, line 25: We can return directly here

```suggestion
  return _fetchReturnRequirements(licenceReference)
```

app/services/jobs/return-logs/fetch-licence-return-requirements.service.js
```diff
@@ -0,0 +1,128 @@
...
+  return externalIdsArray
+}
+
+async function _fetchReturnRequirements (licenceReference) {
+  // pass in false to ge the all year start and end date to make sure we cover the entire year
+  const _cycleEndDate = cycleEndDate(false)
+  const _cycleStartDate = cycleStartDate(false)
+  const externalIds = await _fetchExternalIds(licenceReference)
+
+  return await ReturnRequirementModel.query()
```
Cruikshanks, line 47: We can drop the await here

```suggestion
  return ReturnRequirementModel.query()
```

app/services/jobs/return-logs/fetch-return-requirements.service.js
```diff
@@ -0,0 +1,149 @@
...
+ *
+ * @param {boolean} summer - are we running summer cycle or all year
+ * @param {string} licenceReference - if provided only do the return log for that licence reference
+ *
+ * @returns {Promise<Array>} the list of return requirement ids
+ */
+async function go (summer, licenceReference) {
+  const returnRequirements = await _fetchReturnRequirements(summer, licenceReference)
+
+  return returnRequirements
```
Cruikshanks, line 26: We can return directly here.

```suggestion
  return _fetchReturnRequirements(summer, licenceReference)
```

app/services/jobs/return-logs/fetch-return-requirements.service.js
```diff
@@ -0,0 +1,149 @@
...
+
+  return externalIdsArray
+}
+
+async function _fetchReturnRequirements (summer, licenceReference) {
+  const _cycleEndDate = cycleEndDate(summer)
+  const _cycleStartDate = cycleStartDate(summer)
+  const externalIds = await _fetchExternalIds(_cycleStartDate)
+
+  const results = await ReturnRequirementModel.query()
```
Cruikshanks, line 48: We can return directly here.

```javascript
return ReturnRequirementModel.query()
```

app/services/jobs/return-logs/generate-return-logs.service.js
```diff
@@ -0,0 +1,165 @@
...
+/**
+ * Generates the payload for submission to the returns table.
+ *
+ * @param {Array} returnRequirements - the return requirements to be turned into return logs
+ *
+ * @returns {Promise<Array>} the array of return log payloads to be created in the database
+ */
+async function go (returnRequirements) {
+  const returnLogs = returnRequirements.map(async (requirements) => {
+    const startDate = _getStartDate(requirements.summer, requirements.returnVersion)
```
Cruikshanks, line 27: Rename to drop the `get` prefix.

```javascript
    const startDate = _startDate(requirements.summer, requirements.returnVersion)
```

app/services/jobs/return-logs/generate-return-logs.service.js
```diff
@@ -0,0 +1,165 @@
...
+    const endDate = _getLicenceEndDate(requirements.summer, requirements.returnVersion)
```
Cruikshanks, line 28: Rename to drop the `getLicence` prefix.

```javascript
    const endDate = _endDate(requirements.summer, requirements.returnVersion)
```

app/services/jobs/return-logs/generate-return-logs.service.js
```diff
@@ -0,0 +1,165 @@
...
+    const id = _createReturnLogId(requirements, startDate, endDate)
```
Cruikshanks, line 29: Rename to drop the `createReturnLog` prefix.

```javascript
    const id = _id(requirements, startDate, endDate)
```

app/services/jobs/return-logs/generate-return-logs.service.js
```diff
@@ -0,0 +1,165 @@
...
+    const metadata = await _createMetaData(requirements.summer, endDate, requirements)
```
Cruikshanks, line 30: Rename to drop the `create` prefix.

```javascript
    const metadata = await _metadata(requirements.summer, endDate, requirements)
```

app/services/jobs/return-logs/generate-return-logs.service.js
```diff
@@ -0,0 +1,165 @@
...
+      periodEndDay: requirements.abstractionPeriodEndDay.toString(),
+      periodEndMonth: requirements.abstractionPeriodEndMonth.toString()
+    },
+    points: _createPointsMetaData(requirements.returnRequirementPoints),
+    purposes: _createPurposesMetaData(requirements.returnRequirementPurposes),
+    version: 1
+  }
+}
+
+function _createPointsMetaData (returnRequirementPoints) {
```
Cruikshanks, line 76: Rename to `_metadataPoints`.

app/services/jobs/return-logs/generate-return-logs.service.js
```diff
@@ -0,0 +1,165 @@
...
+      name: returnRequirementPoint.description,
+      ngr1: returnRequirementPoint.ngr1,
+      ngr2: returnRequirementPoint.ngr2,
+      ngr3: returnRequirementPoint.ngr3,
+      ngr4: returnRequirementPoint.ngr4
+    }
+  })
+}
+
+function _createPurposesMetaData (returnRequirementPurposes) {
```
Cruikshanks, line 88: Rename to `_metadataPurposes`.

app/services/jobs/return-logs/generate-return-logs.service.js
```diff
@@ -0,0 +1,165 @@
...
+    returnVersion.endDate]
+    .filter((date) => { return date !== null })
+    .map((date) => { return new Date(date) })
+
+  if (dates.length === 0) {
+    return cycleEndDate(summer)
+  }
+
+  dates.map((date) => { return date.getTime() })
+  const _earliestEndDate = new Date(Math.min(...dates))
```
Cruikshanks, line 128: This section can be moved into its own function to determine the earliest end date.

app/services/jobs/return-logs/generate-return-logs.service.js
```diff
@@ -0,0 +1,165 @@
...
+
+  if (summer) {
+    const _endOfSummerCycle = endOfSummerCycle()
```
Cruikshanks, line 131: This is an example of what I was trying to convey with the date helpers. We have

```javascript
endOfSummerCycle ()
endOfWinterAndAllYearCycle()

cycleEndDate(summer)
```

The first type we specifically call after checking the flag `summer` in this function.

```javascript
if (summer) {
    const _endOfSummerCycle = endOfSummerCycle()
    // ...
}
```

If the first two followed the pattern of `cycleEndDate()`, not only would things be consistent, I think we could remove a chunk of the code in this function.

But to support that I think it might be useful to rename the existing helpers. for example

- `cycleEndDate(summer)` -> `cycleEndDateAsISO(summer)`
- `endOfSummerCycle() & endOfWinterAndAllYearCycle()` -> `cycleEndDate(summer)`

Then the code becomes.

```javascript
const cycleEndDate = cycleEndDate(summer)

if (earliestEndDate < cycleEndDate) {
  // ...
}
```

Cruikshanks (changes requested):
First pass of some things I think we can tweak and update.

app/services/jobs/return-logs/generate-return-logs.service.js
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  return cycleEndDateAsISO(summer)
+}
+
+function _earliestDate (summer, returnVersion) {
+  const dates = [returnVersion.licence.expiredDate,
+    returnVersion.licence.lapsedDate,
+    returnVersion.licence.revokedDate,
+    returnVersion.endDate]
```
Cruikshanks, line 130:
```suggestion
  const dates = [
    returnVersion.licence.expiredDate,
    returnVersion.licence.lapsedDate,
    returnVersion.licence.revokedDate,
    returnVersion.endDate
  ]
```

app/services/jobs/return-logs/generate-return-logs.service.js
```diff
@@ -0,0 +1,161 @@
...
+    .filter((date) => { return date !== null })
```
Cruikshanks, line 131: Returning the value should work in the same way. We also avoid putting these as one-liners. It means we are always consistent, and IMHO, the one-liner fools us into thinking something is more straightforward than it might be.

```suggestion
    .filter((date) => {
      return date
    })
```

app/services/jobs/return-logs/generate-return-logs.service.js
```diff
@@ -0,0 +1,161 @@
...
+    .map((date) => { return new Date(date) })
```
Cruikshanks, line 132:
```suggestion
    .map((date) => {
      return new Date(date)
    })
```

app/services/jobs/return-logs/generate-return-logs.service.js
```diff
@@ -0,0 +1,161 @@
...
+
+  if (dates.length === 0) {
+    return cycleEndDateAsISO(summer)
+  }
+
+  dates.map((date) => { return date.getTime() })
```
Cruikshanks, line 138:
```suggestion
  dates.map((date) => {
    return date.getTime()
  })
```

app/services/jobs/return-logs/generate-return-logs.service.js
Cruikshanks, line 1: Also just need to review the order of the private methods to ensure they are alphabetical.

test/lib/dates.lib.test.js
```diff
@@ -86,4 +86,100 @@ describe('Dates lib', () => {
...
+      expect(result).to.equal(summerDueDate)
+    })
+
+    it('should return the due date of the winter and all year cycle  with summer as false', () => {
+      const allYearDueDate = new Date(new Date().getFullYear() + 1, 3, 28)
+      const result = DateLib.cycleDueDate(false)
+
+      expect(result).to.equal(allYearDueDate)
+    })
+  })
```
Cruikshanks, line 104: Just highlighting this one, but the example applies to the rest of the new date tests.

First, the aim is that the test output we see reads as much as possible as a comprehensive sentence.

So, currently these will read as

```text
cycleDueDate
  [it] should return the due date of the summer cycle with summer as true
  [it] should return the due date of the winter and all year cycle with summer as false
```

Ideally, I'd be aiming for something like this.

```text
#cycleDueDate
  when the requested cycle is "summer"
    it] should return the due date for the current summer cycle
  when the requested cycle is not "summer" (winter and all year)
    [it] should return the due date of the current winter and all year cycle
```

The next aim is to follow the AAA pattern (Arrange, Act, Assert) with the tests. To make this clear we try to always ensure that any setup (Arrange) is done outside of the `it()` in a `beforeEach()`. The first line in the `it()` then becomes our Act, followed by any assertions.

So, my suggested changes for this would be
```javascript
  // You'll want to place these at the top of the file after the initial describe()
  let expectedDate
  let summer

  describe('#cycleDueDate', () => {
    describe('when the requested cycle is "summer"', () => {
      beforeEach(() => {
        summer = true

        expectedDate = new Date(new Date().getFullYear() + 1, 10, 28)
      })

      it('should return the due date for the current summer cycle', () => {
        const result = DateLib.cycleDueDate(summer)
  
        expect(result).to.equal(expectedDate)
      })
    })
    
    describe('when the requested cycle is "summer"', () => {
      beforeEach(() => {
        summer = false

        expectedDate = new Date(new Date().getFullYear() + 1, 3, 28)
      })
    })

    it('should return the due date of the current winter and all year cycle', () => {
      const result = DateLib.cycleDueDate(summer)

      expect(result).to.equal(expectedDate)
    })
  })
```

test/services/import/legacy/process-licence.service.test.js
```diff
@@ -83,6 +89,21 @@ describe('Import Legacy Process Licence service', () => {
...
   })
 
+  describe('when there is a valid NALD licence to import without an existing licence', () => {
+    beforeEach(() => {
+      Sinon.stub(TransformLicenceService, 'go').resolves({ naldLicenceId, regionCode, transformedLicence })
+      persistLicenceServiceStub = Sinon.stub(PersistLicenceService, 'go').resolves(licenceId)
+      processLicenceReturnLogsServiceStub = Sinon.stub(ProcessLicenceReturnLogsService, 'go').resolves()
+    })
+
+    it('saves the imported licence and creates the return logs', async () => {
```
Cruikshanks, line 99:
```suggestion
    it('saves the imported licence but does not process the return logs', async () => {
```

test/services/jobs/return-logs/process-licence-return-logs.service.test.js
```diff
@@ -0,0 +1,67 @@
...
+
+// Thing under test
+const ProcessLicenceReturnLogsService = require('../../../../app/services/jobs/return-logs/process-licence-return-logs.service.js')
+
+describe('Process licence return logs service', () => {
+  const allYearDueDate = new Date(new Date().getFullYear() + 1, 3, 28).toISOString().split('T')[0]
+  const allYearEndDate = new Date(new Date().getFullYear() + 1, 2, 31).toISOString().split('T')[0]
+  const allYearStartDate = new Date(new Date().getFullYear(), 3, 1).toISOString().split('T')[0]
+
+  describe('a valid licence reference is provided', () => {
```
Cruikshanks, line 28:
```suggestion
  describe('when a valid licence reference is provided', () => {
```

test/services/jobs/return-logs/process-licence-return-logs.service.test.js
```diff
@@ -0,0 +1,67 @@
...
+
+      // BaseRequest depends on the GlobalNotifier to have been set.
+      // This happens in app/plugins/global-notifier.plugin.js when the app starts up and the plugin is registered.
+      // As we're not creating an instance of Hapi server in this test we recreate the condition by setting
+      // it directly with our own stub
+      notifierStub = { omg: Sinon.stub(), omfg: Sinon.stub() }
+      global.GlobalNotifier = notifierStub
+    })
+
+    it('can successfully save a return log in the database', async () => {
```
Cruikshanks, line 51:
```suggestion
    it('generates and saves any returns logs required for the current cycle', async () => {
```

Cruikshanks (changes requested):

Cruikshanks (approved):

PR #1295, opened 2024-08-30
Fix return version tests

jonathangoulding:
As part of ongoing work to remove database cleans from the test suit we are getting intermittent tests failing mainly due to unique constraints where the range is too small to ensure uniqueness.

This change adds a generate random region code function explaining the issue.

The aim of this change is to encapsulate the issue and remove any need to explain the issue else where in the tests.

### DIFF
--- a/test/support/general.js
+++ b/test/support/general.js
@@ -113,6 +113,6 @@ function randomRegionCode () {
 module.exports = {
   postRequestOptions,
   randomInteger,
-  selectRandomEntry,
-  randomRegionCode
+  randomRegionCode,
+  selectRandomEntry
 }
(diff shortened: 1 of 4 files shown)


test/support/general.js
```diff
@@ -95,8 +95,24 @@ function selectRandomEntry (data) {
...
+function randomRegionCode () {
+  return randomInteger(1, 999999)
+}
+
 module.exports = {
   postRequestOptions,
   randomInteger,
-  selectRandomEntry
+  selectRandomEntry,
+  randomRegionCode
```
Cruikshanks, line 117: Just this NIT

```suggestion
  randomRegionCode,
  selectRandomEntry
```

Cruikshanks (changes requested):

robertparkinson (approved):

Cruikshanks (approved):

PR #1297, opened 2024-08-30
Fix charge category charge ref tests

jonathangoulding:

As part of ongoing work to remove database cleans from the test suit we are getting intermittent tests failing. In this case we are seeing tests interfering with each other when they use the generated charge reference in the ChargeCategoryHelper.

A charge version is made up of three parts. Some with a small range.

This change will broaden the range of the charge version but attempt to document the parts of the charge reference and why there is a defined range for the parts.

### DIFF
--- a/test/support/helpers/charge-category.helper.js
+++ b/test/support/helpers/charge-category.helper.js
@@ -74,11 +74,27 @@ function defaults (data = {}) {
   }
 }
 
+/**
+ * Generate Charge reference - e.g "4.6.42"
+ *
+ * A charge reference is built in three parts
+ *
+ * firstPart why - 4 (does it have a know range ? )
+ * secondPart why - (1, 6)
+ * thirdPart why - (1, 42)
+ *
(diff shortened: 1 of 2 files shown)


test/support/helpers/charge-category.helper.js
```diff
@@ -74,11 +74,27 @@ function defaults (data = {}) {
   }
 }
 
+/**
+ * Generate Charge reference - e.g "4.6.42"
+ *
+ * A charge reference is built in three parts
+ *
+ * firstPart why - 4 (does it have a know range ? )
```
jonathangoulding, line 82: Try and get answer for the why questions.

Cruikshanks: Charge Categories are a reference record, not transactional.

Rather than tweak the generate, these should be moved to being a seed and the helper updated to just provide a `select()` method.

On why, because someone in policy liked the numbers (and that's all you'll ever know!)

jonathangoulding: I will close this PR.

PR #1299, opened 2024-08-30
Only include the required information in the metadata field

robertparkinson:

As part of the work to replace the legacy system NALD we need to generate return logs. This PR refactors the metadata field so that it only includes the required information from the points and purposes tables.


app/services/jobs/return-logs/fetch-return-logs.service.js
```diff
@@ -115,7 +147,17 @@ async function _fetchReturnRequirements (isSummer, licenceReference) {
...
+      builder.select(['description',
+        'description',
+        'ngr1',
+        'ngr2',
+        'ngr3',
+        'ngr4'])
+    })
+    .withGraphFetched('returnRequirementPurposes.primaryPurpose')
+    .withGraphFetched('returnRequirementPurposes.secondaryPurpose')
+    .withGraphFetched('returnRequirementPurposes.purpose')
```
Cruikshanks, line 160: If you only need the legacy IDs and descriptions off these, then they should have `modifyGraph()` functions which specify this in the select.

Without it, Objection.js will do `SELECT *` pulling everything off the record, and for which I have a strict 🚫 rule.

test/services/jobs/return-logs/fetch-return-logs.service.test.js
```diff
@@ -9,8 +9,10 @@ const { expect } = Code
 
 // Test helpers
 const LicenceHelper = require('../../../support/helpers/licence.helper.js')
+const PrimaryPurposeModel = require('../../../../app/models/primary-purpose.model.js')
+const SecondaryPurposeModel = require('../../../../app/models/secondary-purpose.model.js')
+const PurposeModel = require('../../../../app/models/purpose.model.js')
```
Cruikshanks, line 14: Just need to sort the ordering of these requires, including the existing ones.

test/services/jobs/return-logs/fetch-return-logs.service.test.js
```diff
@@ -47,6 +49,10 @@ describe('Fetch return logs service', () => {
     it('should return one return log payload', async () => {
       const result = await FetchReturnLogsService.go(false, licence.licenceRef)
 
+      const primaryPurpose = await PrimaryPurposeModel.query().findById(returnRequirementPurpose.primaryPurposeId)
+      const secondaryPurpose = await SecondaryPurposeModel.query().findById(returnRequirementPurpose.secondaryPurposeId)
+      const tertiaryPurpose = await PurposeModel.query().findById(returnRequirementPurpose.purposeId)
+
```
Cruikshanks, line 55: These records are pre-seeded in the DB.

We should be able to use their helper `select()` functions to pick them, then assign those picked ones to your test return requirement purposes (just as done with `Region.select()`).

You can then use them again to assert the legacy IDs and descriptions are what you expected.

Cruikshanks (changes requested):

app/services/jobs/return-logs/fetch-return-logs.service.js
```diff
@@ -11,26 +11,29 @@ const ReturnVersionModel = require('../../../models/return-version.model.js')
...
 
-const endOfSummerCycle = new Date(new Date().getFullYear() + 1, summerEndMonth, summerEndDay)
-const endOfWinterAndAllYearCycle = new Date(new Date().getFullYear() + 1, allYearEndMonth, allYearEndDay)
+const endOfSummerCycle = new Date(new Date().getFullYear() + 1, SUMMER_END_MONTH, SUMMER_END_DAY)
+const endOfWinterAndAllYearCycle = new Date(new Date().getFullYear() + 1, ALL_YEAR_END_MONTH, ALL_YEAR_END_DAY)
 
 /**
  * Fetch all return requirements that need return logs created.
  *
+ * @param isSummer - are we running summer cycel or all year - boolean
```
Beckyrose200, line 34:
```suggestion
 * @param {boolean} isSummer - are we running summer cycle or all year
```

app/services/jobs/return-logs/fetch-return-logs.service.js
```diff
@@ -11,26 +11,29 @@ const ReturnVersionModel = require('../../../models/return-version.model.js')
...
+ * @param licenceReference - if provided only do the return log for that licence reference - string
```
Beckyrose200, line 35:
```suggestion
 * @param {string} licenceReference - if provided only do the return log for that licence reference
```

Cruikshanks (approved):

PR #1303, opened 2024-09-02
Fix licence agreements recalculate link bug

Beckyrose200:
https://github.com/DEFRA/water-abstraction-system/pull/1291

During a recent bug fix for the view-licence presenter, we removed some data from the presenter that we thought was redundant to it. This data is needed, so this PR is adding it back in thus fixing some broken functionality that was created by removing it (mainly the recalculate bills link showing for a licence agreement).


Cruikshanks (approved):
> approving the PR that fixes what you broke

PR #1304, opened 2024-09-03
Import company data for a licence

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4649

This change will use the nald licence id and region to update the company data. This means going forwards we will only import company when the licence is imported.

This change focuses on the company data and not address or contacts tables which will be done after this work.


app/presenters/import/legacy/companies.presenster.js
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * We will remove duplicates of the party id to reduce unnecessary inserts into the database.
+ *
+ * @param {ImportLegacyCompaniesType[]} companies - the legacy NALD companies
+ *
+ * @returns {object[]} the NALD companies data transformed into the WRLS format for a company
+ * ready for validation and persisting
+ */
+function go (companies) {
+  // need to remove duplicates
```
jonathangoulding, line 22: Remove duplicates

app/services/import/legacy/fetch-company.service.js
```diff
@@ -0,0 +1,103 @@
...
+ */
+async function go (regionCode, licenceId) {
+  const query = _query()
+
+  const { rows } = await db.raw(query, [regionCode, licenceId])
+
+  return rows
+}
+
+function _query () {
```
jonathangoulding, line 28: I think we actually need to get the parties from the lic version as well here. 

I can see locally that there is a party in the licence version table that is not in the licence roles table. ( the licence holder as we see it in the UI).

I might add it as another query have the transform query handle merging together the companies for persisting  and try and return the same data object shape to create the contacts and address from the `ACON_AADD_ID` and `ACON_APAR_ID`

app/services/import/legacy/transform-companies.service.js
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * But some of these companies will share the same party id. We can remove these duplicates
+ * and only persist the company once.
+ *
+ * @param {object[]} transformedCompanies - The transformed companies
+ *
+ * @param transformedCompanies
+ * @returns {object[]} -The transformed companies array with no duplicates
+ */
+function _removeDuplicateCompanies (transformedCompanies) {
```
jonathangoulding, line 52: I think we remove duplicates for now. 

We might be able to save the addresses etc based on the simplified list of companies. 

We have the party code, might need to get the region when it comes to it but i think it should work this way.

jonathangoulding: We will get all the party id's (company id's) and transform them into a set. 

This array can then be used to fetch the addresses / contacts etc in other transform services.

test/services/import/legacy/transform-companies.service.test.js
```diff
@@ -0,0 +1,87 @@
...
+      })
+    })
+  })
+
+  describe('when no matching valid legacy companies are found', () => {
+    beforeEach(() => {
+      Sinon.stub(FetchCompanyService, 'go').resolves([])
+    })
+
+    it('returns an empty array', async () => {
```
jonathangoulding: I think we might want to throw an error here instead if there are no companies ? 

A licence must have a company / contact.

jonathangoulding: Will be added in the overall validator.

Cruikshanks (approved):

PR #1309, opened 2024-09-05
Add external id to companies view.

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4659

As part of the ongoing to work to migrate the import service into the water-abstraction-system, we need to add the external id for the companies (know as parties in NALD).

When we import the companies (parties) we need to upsert the data. The current import service uses the external id to resolve the conflict.

This follows the same approach as the other import work.


db/migrations/public/20240905081225_alter-companies-view.js
```diff
@@ -0,0 +1,47 @@
...
+    .schema
+    .dropView(viewName)
+    .createView(viewName, (view) => {
+      // NOTE: We have commented out unused columns from the source table
+      view.as(knex(viewName).withSchema('crm_v2').select([
+        'companies.company_id AS id',
+        'companies.name',
+        'companies.type',
+        'companies.company_number',
+        'companies.external_id', // is used when importing data from nald
```
Cruikshanks, line 38: The down query needs to revert the change, which I believe in this case would be to comment out (drop) the external ID column.

```suggestion
        '// companies.external_id', // is used when importing data from nald
```

Cruikshanks (changes requested):

Cruikshanks (approved):

PR #1312, opened 2024-09-05
Create seeder for `charge_categories`

Jozzey:
Currently, with the way data is being inserted into `charge_categories` using the helper to create records during unit testing. We are getting intermittent failures when running the unit tests when the charge `reference` gets duplicated due to its lack of randomness.

We have therefore decided that since the data in `charge_categories` is reference data we shouldn't be creating new charge category records each time a unit test requires one. Instead, we should just be seeding the data once and then writing the unit tests to use this seeded data. The bulk of this work has been done for other tables in this PR https://github.com/DEFRA/water-abstraction-system/pull/1230

This PR will create the functions required to seed the `charge_categories` and then fix any code affected by this change.


robertparkinson (approved):

db/seeds/13-charge-categories.seed.js
```diff
@@ -0,0 +1,33 @@
+'use strict'
+
+const { timestampForPostgres } = require('../../app/lib/general.lib.js')
```
Beckyrose200, line 3: Just a nit, I think this needs to be below line 5

test/support/helpers/charge-category.helper.js
```diff
@@ -4,85 +4,32 @@
...
- * - `shortDescription` - Low loss, non-tidal, restricted water, up to and including 5,000 ML/yr, Tier 1 model
- * - `tidal` - false
- * - `lossFactor` - low
- * - `modelTier` - tier 1
- * - `restrictedSource` - true
- * - `minVolume` - 0
- * - `maxVolume` - 5000,
- * - `dateCreated` - Date.now()
+ * So, they are seeded automatically when tests are run. Tests that need to link to a record can use this method to
+ * select a specific entry, or have it it return one at random.
```
Beckyrose200, line 17:
```suggestion
 * select a specific entry, or have it return one at random.
```

Beckyrose200 (commented):
Looks good

PR #1316, opened 2024-09-09
Update view licence summary to use new points data

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4600
https://eaflood.atlassian.net/browse/WATER-4645

We've been extending and amending the import of return versions from NALD to WRLS as part of our work to switch from NALD to WRLS to manage them.

In [Update Return req set up journey to use new points](https://github.com/DEFRA/water-abstraction-system/pull/1301) we updated the return requirement setup journey to use the new `water.licence_version_purpose_points` table we've created and now populate thanks to [a change to the import](https://github.com/DEFRA/water-abstraction-system/pull/1301).

The one other place we are referencing the `permit.licence` table to retrieve point information is the view licence summary page. This change updates its logic to grab the points data using our new [points solution](https://github.com/DEFRA/water-abstraction-system/pull/1352): `water.licence_version_purposes` linked to `water.points` via `water.licence_version_purpose_points`.


Cruikshanks:
I hadn't spotted that the summary page displays "Source of supply". This is taken from the first point in the `permit.licence` JSONB blob.

We don't have that on the points we're importing, which means we need to do a bit more work around points before we can progress this PR.

Jozzey (approved):

PR #1317, opened 2024-09-09
Import contact data for a licence

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4667

This change will use the nald licence id and region to update the companies (parties) contact data. This means going forwards we will only import companies (parties) contact when the licence is imported.

This change focuses on the companies (parties) contact data and not address or contacts tables which will be done after this work.


app/services/import/legacy/fetch-company-contacts.service.js
```diff
@@ -0,0 +1,83 @@
...
+        ELSE np."FORENAME"
+        END
+      )  AS firstName,
+    (
+      CASE np."NAME"
+        WHEN 'null' THEN NULL
+        ELSE np."NAME"
+        END
+      )  AS lastName
+  FROM import."NALD_PARTIES" np
```
jonathangoulding, line 56: Need to add logic to ignore if the type is ORG. 

Old import logic - 

```
/**
 * Maps NALD party to CRM contact
 * @param {Object} NALD party
 * @return {Object}
 */
const mapContact = party => {
  if (party.APAR_TYPE === 'ORG') {
    return null
  }
  return {
    salutation: str.mapNull(party.SALUTATION),
    initials: str.mapNull(party.INITIALS),
    firstName: str.mapNull(party.FORENAME),
    lastName: str.mapNull(party.NAME),
    externalId: `${party.FGAC_REGION_CODE}:${party.ID}`,
    _nald: party
  }
}
```

app/services/import/legacy/fetch-company.service.js
```diff
@@ -29,24 +29,6 @@ function _query () {
...
-        WHEN 'null' THEN NULL
-        ELSE np."FORENAME"
-        END
-      )  AS firstName,
-    (
-      CASE np."NAME"
-        WHEN 'null' THEN NULL
-        ELSE np."NAME"
-        END
-      )  AS lastName,
```
jonathangoulding, line 49: Check this trim is actullay needed

app/services/import/legacy/fetch-company-contacts.service.js
jonathangoulding: One thing to note is that company only has a contact id it has a licence version and the type is not organisation.

app/services/import/legacy/fetch-company.service.js
jonathangoulding: I think this is needed. 

This was the previous code: 

```
const mapName = party => {
  const firstNameKey = party.FORENAME === 'null' ? 'INITIALS' : 'FORENAME'

  const parts = [
    party.SALUTATION,
    party[firstNameKey],
    party.NAME
  ]

  return parts
    .map(str.mapNull)
    .filter(value => value)
    .join(' ')
}
```

We can see in the database there are companies (parties) where the `APAR_TYPE = person` and in WRLS the name in the company table is the result of this logic.

app/services/import/legacy/fetch-contacts.service.js
```diff
@@ -0,0 +1,86 @@
...
+      CASE np."NAME"
+        WHEN 'null' THEN NULL
+        ELSE np."NAME"
+        END
+      )  AS last_name
+  FROM import."NALD_PARTIES" np
+    LEFT JOIN import."NALD_ABS_LIC_VERSIONS" nalv
+      ON nalv."ACON_APAR_ID" = np."ID"
+      AND nalv."FGAC_REGION_CODE" = np."FGAC_REGION_CODE"
+    LEFT JOIN import."NALD_LIC_ROLES" nlr
```
jonathangoulding, line 61: revert this

app/services/import/legacy/fetch-contacts.service.js
```diff
@@ -0,0 +1,86 @@
...
+      ON nlr."ACON_APAR_ID" = np."ID"
+      AND nlr."FGAC_REGION_CODE" = np."FGAC_REGION_CODE"
+  WHERE np."APAR_TYPE" != 'ORG'
```
jonathangoulding, line 64: keep this

app/services/import/legacy/fetch-company.service.js
```diff
@@ -29,24 +29,6 @@ function _query () {
...
     (
       CASE np."APAR_TYPE"
         WHEN 'PER' THEN 'person'
```
jonathangoulding, line 34: check if department is a field ?

app/services/import/persist-licence.service.js
```diff
@@ -156,6 +165,23 @@ async function _persistCompany (trx, updatedAt, company) {
...
+      'salutation',
+      'initials',
+      'firstName',
+      'lastName',
+      'updatedAt'
+    ])
+}
+
+async function _persistsCompanyContact (trx, updatedAt, contact) {
+  // raw query needed here to link the id's together
```
jonathangoulding, line 182: Add the knex query to link using the external id

app/services/import/persist-licence.service.js
```diff
@@ -156,6 +167,38 @@ async function _persistCompany (trx, updatedAt, company) {
     ])
 }
 
+async function _persistContact (trx, updatedAt, contact) {
+  return ContactModel.query(trx)
+    .insert({ ...contact, updatedAt })
+    .onConflict('externalId')
+    .merge([
+      'salutation',
```
jonathangoulding, line 175: Should this update the data source to nald ? what if this has changed in wrls. Might be a none issue as the test helper sets this as wrls.

app/services/import/persist-licence.service.js
```diff
@@ -156,6 +167,38 @@ async function _persistCompany (trx, updatedAt, company) {
...
+      'lastName',
+      'updatedAt'
+    ])
+}
+
+async function _persistsCompanyContact (trx, updatedAt, companyContact) {
+  const { externalId, startDate, licenceRoleId } = companyContact
+
+  return db.raw(`
+    INSERT INTO public."company_contacts" (company_id, contact_id, licence_role_id, start_date, "default", created_at, updated_at)
```
jonathangoulding, line 187: Issue with the on conflict not updating the existing record - it does not insert a new record so the conflict key must be working...

app/services/import/legacy/transform-contacts.service.js
```diff
@@ -0,0 +1,49 @@
+'use strict'
+
+/**
+ * Transforms NALD company data into a valid object that matches the WRLS structure
```
Cruikshanks, line 4:
```suggestion
 * Transforms NALD contact data into a valid object that matches the WRLS structure
```

app/services/import/legacy/transform-contacts.service.js
```diff
@@ -0,0 +1,49 @@
...
+ * @module ImportLegacyTransformContactsService
+ */
+
+const FetchContactsService = require('./fetch-contacts.service.js')
+const ContactPresenter = require('../../../presenters/import/legacy/contact.presenter.js')
+const CompanyContactPresenter = require('../../../presenters/import/legacy/company-contact.presenter.js')
+const ImportContactValidator = require('../../../validators/import/contact.validator.js')
+
+/**
+ * Transforms NALD company data into a validated object that matches the WRLS structure
```
Cruikshanks, line 14:
```suggestion
 * Transforms NALD contact data into a validated object that matches the WRLS structure
```

app/services/import/legacy/transform-contacts.service.js
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @param {string} regionCode - The NALD region code
+ * @param {string} licenceId - The NALD licence ID
+ * @param {object[]} transformedCompanies
+ *
+ * @returns {Promise<object>} an object representing an array of valid WRLS transformed companies and
+ * an array of companies from the db
```
Cruikshanks, line 21: My bad. Didn't spot on the `TransformCompaniesService` that there is a documented return when we don't actually have one, and no doubt this is here because of that.

```suggestion
```

app/services/import/legacy/transform-contacts.service.js
```diff
@@ -0,0 +1,49 @@
...
+  naldContacts.forEach((naldContact) => {
+    const matchingCompany = _matchingCompany(transformedCompanies, naldContact)
+
+    const transformedContact = ContactPresenter.go(naldContact)
+
+    ImportContactValidator.go(transformedContact)
+
+    matchingCompany.contact = { ...transformedContact, dataSource: 'nald' }
+
+    if (naldContact.licence_role_id && naldContact.start_date) {
```
Cruikshanks, line 35: We inner join to WRLS licence roles, so if there wasn't a match we'd have no result rather than a blank `licence_role_id`. And I don't think there are any NALD licence versions with no `EFF_ST_DATE` else I suspect `to_date()` would blow up and cause the query to crash.

So, I don't think we need this check here, but I could be wrong? 🤷

app/services/import/legacy/fetch-contacts.service.js
```diff
@@ -0,0 +1,88 @@
+'use strict'
+
+/**
+ * Fetches the company data from the import.NALD_PARTIES table for the licence ref
+ * @module FetchContactsService
+ */
+
+const { db } = require('../../../../db/db.js')
+
+const licenceHolderName = 'licenceHolder'
```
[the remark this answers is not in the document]
jonathangoulding: Have removed the const

app/services/import/legacy/transform-contacts.service.js
jonathangoulding: No you are correct. 

It would have been needed in the previous version. But it now redundant. 

Good spot

Cruikshanks (approved):

PR #1318, opened 2024-09-09
Add external id to contacts view.

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4667

As part of the ongoing to work to migrate the import service into the water-abstraction-system, we need to add the external id for the contacts table.

When we import the contact we need to upsert the data. The current import service uses the external id to resolve the conflict.

This follows the same approach as the other import work.

### DIFF
--- a/db/migrations/public/20240909132727_alter-contacts-view.js
+++ b/db/migrations/public/20240909132727_alter-contacts-view.js
@@ -1,0 +1,60 @@
+'use strict'
+
+const viewName = 'contacts'
+
+exports.up = function (knex) {
+  return knex
+    .schema
+    .dropViewIfExists(viewName)
+    .createView(viewName, (view) => {
+      // NOTE: We have commented out unused columns from the source table
+      view.as(knex(viewName).withSchema('crm_v2').select([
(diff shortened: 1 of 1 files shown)


db/migrations/public/20240909132727_alter-contacts-view.js
```diff
@@ -0,0 +1,60 @@
...
+    .dropView(viewName)
+    .createView(viewName, (view) => {
+      // NOTE: We have commented out unused columns from the source table
+      view.as(knex(viewName).withSchema('crm_v2').select([
+        'contacts.contact_id AS id',
+        'contacts.salutation',
+        'contacts.first_name',
+        'contacts.middle_initials',
+        'contacts.last_name',
+        // external_id is populated for contacts migrated from NALD and is used to resolve import conflicts
```
Cruikshanks, line 45:
```suggestion
        // 'external_id', // is populated for contacts migrated from NALD and is used to resolve import conflicts
```

Cruikshanks (changes requested):
Just a NIT

Cruikshanks (approved):

PR #1319, opened 2024-09-10
Add wlrs licence id to the nald licence import

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4545

Relevant changes - https://github.com/DEFRA/water-abstraction-system/pull/1294

As part of the ongoing work to migrate the legacy import jobs we need to establish if a licence already exists in the wrls database.

We check if the licence already exists, if it does then we want to do nothing. If it does not exist we want to generate return logs.

This change is added to highlight where we need to add this additional logic. And the service call will be added in another piece of work.

### DIFF
--- a/app/services/import/legacy/fetch-licence.service.js
+++ b/app/services/import/legacy/fetch-licence.service.js
@@ -66,11 +66,14 @@ function _query () {
           AND nalv."AABL_ID" = nal."ID"
           AND nalv."STATUS" <> 'DRAFT'
       ) AS earliest_version_start_date,
-      r.id AS region_id
+      r.id AS region_id,
+      l.id as wrls_licence_id
     FROM
       "import"."NALD_ABS_LICENCES" nal
     LEFT JOIN
       public.regions r ON r.nald_region_id = (nal."FGAC_REGION_CODE")::INTEGER
+    LEFT JOIN
+      public.licences l ON l.licence_ref = nal."LIC_NO"
     WHERE
       nal."LIC_NO" = ?;
   `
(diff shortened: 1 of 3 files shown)


Cruikshanks (changes requested):
Unless you know of any issue, we should be able to call the property `licenceId` because that is what it is on the transformed licence object (we can drop the wrls prefix).

And I think you'll need to drop it in the same way we drop `licenceVersions` in the persist service. Try updating `test/services/import/persist-licence.service.test.js` to include the new property: if it blows up you know you do!

On that point, we should also be updating the validator to define that we expect this parameter but null is acceptable.

Cruikshanks:
> > Unless you know of any issue, we should be able to call the property `licenceId` because that is what it is on the transformed licence object (we can drop the wrls prefix).
> > And I think you'll need to drop it in the same way we drop `licenceVersions` in the persist service. Try updating `test/services/import/persist-licence.service.test.js` to include the new property: if it blows up you know you do!
> > On that point, we should also be updating the validator to define that we expect this parameter but null is acceptable.
> 
> This is not used in the transformedLicence so there is no need to validate and exclude.
> 
> This is treated like the naldLicenceId and regionCode
> 
> Also calling it licenceId conflcts when descrutuing with code further down.
> 
> ```
> const licenceId = await PersistLicenceService.go(transformedLicence, transformedCompanies)
> ```

I still think we should be including it in the validator. The validator serves as one place where we define the object we expect to be passed to the persist service. And certainly the test for the validator should be updated to include the property. In fact, I'd be surprised if the test passes when you do because `licenceId` is not in the schema, and we haven't specified `.unknown(true)`.

On the final point, we only need the ID for the log message. It can simply be renamed as `const persistedLicenceId`. Alternatively, you could not set it using the destructure, but do it with a `let licenceId = transformedLicence.licenceId` instead 🤷

Cruikshanks (approved):
Well, I got completely the wrong end of the stick!

All good, carry on!

PR #1327, opened 2024-09-12
Remove `DatabaseSupport.clean()` from unit tests

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4677

We currently have some unit tests that fail if `DatabaseSupport.clean()` is not run before each test. This is usually due to duplicate data being added to the DB causing a constraint violation.

Since we have introduced "seeders" to populate some of our reference data. It should now be possible to fix any tests that currently fail without the DB clean. This will speed up the running of the unit tests as it won't have to rebuild the DB after running some of them.


test/services/bill-runs/two-part-tariff/fetch-charge-versions.service.test.js
```diff
@@ -22,17 +22,17 @@ const WorkflowHelper = require('../../../support/helpers/workflow.helper.js')
 
 // Thing under test
 const FetchChargeVersionsService = require('../../../../app/services/bill-runs/two-part-tariff/fetch-charge-versions.service.js')
+const regionHelper = require('../../../support/helpers/region.helper.js')
```
Cruikshanks, line 25: Think this just needs to be bumped up to the helpers section.

Cruikshanks (changes requested):
Just one little NIT. But this is _outstanding_ work!

test/services/bill-runs/two-part-tariff/fetch-charge-versions.service.test.js
Jozzey: I think that was caused by VSCode trying to be helpful and adding the `require` when I used a lowercase 'r' for the region helper on line 35. I've got the region helper in there twice, DOH!

PR #1328, opened 2024-09-12
Import company contacts data for a licence

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4668

This change will use the nald licence id and region to update the companies (parties) licence holder data. This means going forwards we will only import companies (parties) licence holder when the licence is imported.

This change focuses on the licence holder contact data (for a company) and will link a contact to a company if the contact is a person in the company contacts table.

### DIFF
--- a/app/services/import/legacy/fetch-licence-holders.service.js
+++ b/app/services/import/legacy/fetch-licence-holders.service.js
@@ -1,0 +1,61 @@
+'use strict'
+
+/**
+ * Fetches the Licence Holder data from the import.NALD_PARTIES table for the licence ref
+ * @module FetchLicenceHolderService
+ */
+
+const { db } = require('../../../../db/db.js')
+
+/**
+ * Fetches the licence holder data from the import.NALD_PARTIES table for the licence ref
+ *
+ * The licence holder is determined from the import.NALD_ABS_LIC_VERSIONS table. We get the earliest start date
+ *
+ * @param {string} regionCode - The NALD region code
+ * @param {string} licenceId - The NALD licence ID
+ *
+ * @returns {Promise<ImportLegacyLicenceHolderType[]>}
+ */
+async function go (regionCode, licenceId) {
+  const query = _query()
+
+  const { rows } = await db.raw(query, [regionCode, licenceId])
+
+  return rows
+}
+
+function _query () {
+  return `
+    SELECT DISTINCT ON (np."ID")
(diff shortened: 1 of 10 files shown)


app/services/import/persist-licence.service.js
```diff
@@ -156,6 +163,22 @@ async function _persistCompany (trx, updatedAt, company) {
     ])
 }
 
+async function _persistLicenceHolder (trx, updatedAt, licenceHolder) {
+  // return CompanyContactModel.query(trx)
```
jonathangoulding, line 167: Implement the persist. 

The on conflict needs a bit of work.

app/services/import/legacy/fetch-licence-holders.service.js
```diff
@@ -0,0 +1,61 @@
...
+  const query = _query()
+
+  const { rows } = await db.raw(query, [regionCode, licenceId])
+
+  return rows
+}
+
+function _query () {
+  return `
+    SELECT DISTINCT ON (np."ID")
```
jonathangoulding, line 30: This has been done separately to the previous contact work. 

Maybe it should use the previous contact data. But this functionality encompasses a licence holder (for the company contacts table) as far as i understand it.

So highlighting this is a different thing seemed right ?  

We could merge the two as we really only want to add the contact if we are actually going to link it to a company. ( I will investigate if there will be any edge cases to this understanding)

app/services/import/legacy/fetch-licence-holders.service.js
```diff
@@ -0,0 +1,61 @@
...
+      (concat_ws(':', np."FGAC_REGION_CODE", np."ID")) AS company_external_id,
+      (concat_ws(':', np."FGAC_REGION_CODE", np."ID")) AS contact_external_id,
+      to_date(nalc."EFF_ST_DATE", 'DD/MM/YYYY') as start_date,
```
jonathangoulding, line 33: Where is the returns communication type coming from - double check this is done elsewhere.

app/services/import/persist-licence.service.js
```diff
@@ -156,6 +164,21 @@ async function _persistCompany (trx, updatedAt, company) {
     ])
 }
 
+async function _persistLicenceHolder (trx, updatedAt, licenceHolder) {
```
jonathangoulding, line 167: Add tests

PR #1338, opened 2024-09-17
Licence import flagging for supplementary billing

Beckyrose200:

During the licence import process, we need to check if a licence has been lapsed/revoked or ended. A licence can end at any time, either in the past or the future. If the licence has ended in the past then the licence bills will be incorrect. This means the licence will need to be flagged for either pre-sroc supplementary billing, sroc supplementary billing or two-part tariff supplementary billing (depending on what is affected). This PR is building the service needed to check the licence lapsed/revoked or ended date and flagging if necessary.


Cruikshanks (changes requested):
I think there are a number of places we can simplify things. I reckon we can also avoid doing work in some others.

But what those are will be easier (and less time-consuming!) to go through on a call rather than comments added here.

So, when you are ready, let me know and we can work through it.

Beckyrose200:
NOTE:
For `app/services/licences/supplementary/process-imported-licence.service.js`
Function _earliestChangedDate 

I had some issues doing `Math.min` on the array of dates. 

When doing `Math.min` on the array you need to spread it out first so `Math.min(...filteredDates)`. Without doing this first you get NaN returned.

But if the array is empty and you try to do `Math.min(...[]`), then you get infinity returned. 
SO I needed to add in a check on the filteredDates to check the length of it before attempting `Math.min`.

Also doing `Math.min` on dates that weren't already in a timestamp format, returned the date as a timestamp hence why its wrapped in `new Date()`

app/services/import/determine-supplementary-billing-flags.service.js
```diff
@@ -0,0 +1,71 @@
...
+  try {
+    const licenceChanged = await _licenceChanged(naldLicence, wrlsLicenceId)
+
+    if (!licenceChanged) {
+      return
+    }
+
+    await ProcessImportedLicenceService.go(naldLicence, wrlsLicenceId)
+  } catch (error) {
+    global.GlobalNotifier.omfg('Supplementary Billing Flag failed for licence ', wrlsLicenceId, error)
```
Cruikshanks, line 31: Need to update this call.

First, the message can be changed to make it clearer where the error has occurred, for example, "Determine supplementary billing flags on import failed".

Second, we tend to stub the notifier in our tests and then just confirm they got called with the arguments we expected.

However, if you look at the notifier code you'll see this will cause an error. `omfg()` expects an object for the second arg. So, you'll need to call it as

```javascript
  global.GlobalNotifier.omfg('Supplementary Billing Flag failed for licence ', { licenceId }, error)
```

app/services/import/determine-supplementary-billing-flags.service.js
```diff
@@ -0,0 +1,71 @@
...
+ * Determines if an imported licence should be flagged for supplementary billing.
+ *
+ * This service is responsible for determining whether a licence imported from NALD should be flagged for
+ * supplementary billing. It compares the licences end dates (such as lapsed, revoked or expired dates) between WRLS
+ * and the transformed data, and if there is a change in the dates allows the licence to go on to determining the flags.
+ *
+ * @param {object} naldLicence - the legacy NALD licence
+ * @param {object} wrlsLicenceId - the WRLS licence data
+ */
+async function go (naldLicence, wrlsLicenceId) {
```
Cruikshanks, line 21: We've sat this outside the legacy folder so it theoretically could be used by imports from any system. So, we can't name it `naldLicence`.

I also don't believe there is any confusion as to what system the licence ID relates to.

So, the JSDoc comments and params all need to be updated. Hint, `ProcessImportedLicenceService` might have something you can use!

app/services/import/determine-supplementary-billing-flags.service.js
```diff
@@ -0,0 +1,71 @@
...
+
+  _whereClauses(query, naldLicence)
+
+  const result = await query
+
+  return result.length === 0
+}
+
+function _whereClauses (query, naldLicence) {
+  const { expiredDate, lapsedDate, revokedDate } = naldLicence
```
Cruikshanks, line 48: We need to change the param name (see previous comment)

```suggestion
function _whereClauses (query, importedLicence) {
  const { expiredDate, lapsedDate, revokedDate } = importedLicence
```

I also think it would be useful to add a comment that explains _why_ this method exists, i.e. the problem it is solving.

app/services/licences/supplementary/persist-supplementary-billing-flags.service.js
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * NOTE: Due to the column data type of the includeInPresrocBilling & includeInSrocBilling, one is a string value and
+ * one is a boolean.
+ *
+ * @param {object[]} twoPartTariffFinancialYears - The years that need persisting in the LicenceSupplementaryYears table
+ * @param {boolean} preSrocFlag - `true` or `false` depending on if the licence needs to be flagged for pre sroc billing
+ * @param {boolean} srocFlag - `true` or `false` depending on if the licence needs to be flagged for sroc billing
+ * @param {string} wrlsLicenceId - The UUID of the licence that needs the flags persisting for
+ */
+async function go (twoPartTariffFinancialYears, preSrocFlag, srocFlag, wrlsLicenceId) {
```
Cruikshanks, line 25: Need to rename `wrlsLicenceId` to `licenceId`. Also, could we not update these params to use the same names we've used in `ProcessImportedLicenceService`?

app/services/licences/supplementary/persist-supplementary-billing-flags.service.js
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Persists two-part tariff financial years in the LicenceSupplementaryYears table.
+ * @private
+ */
+async function _flagForLicenceSupplementaryYears (financialYears, wrlsLicenceId) {
+  if (financialYears.length > 0) {
+    const twoPartTariff = true
+
+    await CreateLicenceSupplementaryYearService.go(wrlsLicenceId, financialYears, twoPartTariff)
```
Cruikshanks, line 40: We can just return here

```suggestion
    return CreateLicenceSupplementaryYearService.go(wrlsLicenceId, financialYears, twoPartTariff)
```

app/services/licences/supplementary/persist-supplementary-billing-flags.service.js
```diff
@@ -0,0 +1,52 @@
...
+ * @param {object[]} twoPartTariffFinancialYears - The years that need persisting in the LicenceSupplementaryYears table
+ * @param {boolean} preSrocFlag - `true` or `false` depending on if the licence needs to be flagged for pre sroc billing
+ * @param {boolean} srocFlag - `true` or `false` depending on if the licence needs to be flagged for sroc billing
+ * @param {string} wrlsLicenceId - The UUID of the licence that needs the flags persisting for
+ */
+async function go (twoPartTariffFinancialYears, preSrocFlag, srocFlag, wrlsLicenceId) {
+  const includeInPresrocBilling = preSrocFlag ? 'yes' : 'no'
+
+  await _updateLicenceFlags(includeInPresrocBilling, srocFlag, wrlsLicenceId)
+  await _flagForLicenceSupplementaryYears(twoPartTariffFinancialYears, wrlsLicenceId)
```
Cruikshanks, line 29: We can just return this last call.

```suggestion
  return _flagForLicenceSupplementaryYears(twoPartTariffFinancialYears, wrlsLicenceId)
```

app/services/licences/supplementary/process-imported-licence.service.js
```diff
@@ -0,0 +1,125 @@
...
+  // to change anything on the flags
+  if (!earliestChangedDate) {
+    return
+  }
+
+  const flagForPreSrocSupplementary = _flagForPresrocSupplementary(existingLicenceDetails, earliestChangedDate)
+  const flagForSrocSupplementary = _flagForSrocSupplementary(existingLicenceDetails)
+  const twoPartTariffBillingYears = _flagForTwoPartTariffSupplementary(existingLicenceDetails, earliestChangedDate)
+
+  await PersistSupplementaryBillingFlagsService.go(
```
Cruikshanks, line 37: We can just return this.

```suggestion
  return PersistSupplementaryBillingFlagsService.go(
```

app/services/licences/supplementary/process-imported-licence.service.js
```diff
@@ -0,0 +1,125 @@
...
+
+function _earliestChangedDate (importedLicence, existingLicenceDetails) {
+  const { endDate: currentFinancialYearEndDate } = determineCurrentFinancialYear()
+  const changedDates = []
+
+  let date
+
+  // NOTE: Because checking if dates are equal does not give the expected result in JavaScript (it sees 2 different
+  // objects even if they represent the same date and time) we convert the date object to strings allowing us to compare
+  // the two
```
Cruikshanks, line 53: I'd also update this to explain using `String()` over `myDate.getTime()` also means we can handle where a date value is `null`.

app/services/licences/supplementary/process-imported-licence.service.js
```diff
@@ -0,0 +1,125 @@
...
+}
+
+function _flagForSrocSupplementary (existingLicenceDetails) {
+  const { sroc_charge_versions: chargeVersions } = existingLicenceDetails
+
+  // If the licence has no sroc charge versions return false. We check this before the existing flag, because this
+  // is an opportunity to remove the SROC supplementary billing flag from a licence that won't ever be picked up by
+  // the SROC billing engine (so the flag will never get removed)
+
+  return !!chargeVersions
```
Cruikshanks, line 104: Good call on removing my unnecessary check of the existing flag! But it does mean the comment needs to be updated accordingly.

test/services/licences/supplementary/fetch-existing-licence-details.service.test.js
```diff
@@ -0,0 +1,160 @@
...
+describe('Fetch Existing Licence Details Service', () => {
+  describe('when given a valid licenceId', () => {
+    let licence
+
+    describe('that has charge versions', () => {
+      beforeEach(async () => {
+        licence = await LicenceHelper.add()
+      })
+
+      describe('that are pre-sroc', () => {
```
Cruikshanks, line 28: Could the licence setup be bumped a level? It would just read more clearly if it could.

```suggestion
describe('Fetch Existing Licence Details Service', () => {
  let licence

  describe('when given a licence ID for a licence that exists', () => {
    beforeEach(async () => {
      licence = await LicenceHelper.add()
    })
    
    describe('and that has charge versions', () => {
      describe('that are pre-sroc', () => {
```

test/services/licences/supplementary/fetch-existing-licence-details.service.test.js
Cruikshanks, line 1: I think you could also merge some of the licences being used by these tests. For example, you could create a licence that has all the flags and all the dates, and it should support all your tests for confirming `x is true`. 🤷 

The whole Database.clean() debacle has shown we need to start trying to think about how we can interact less with the database in our tests. Certainly, the less records we add the less chance there is of clashes/breaking other test.

test/services/licences/supplementary/persist-supplementary-billing-flags.service.test.js
Cruikshanks, line 1: I think the creation of the licence needs to move down into the describe blocks, and the `beforeEach()` in them needs to change to a `before()`.

For example

```javascript
    describe('and a pre sroc and sroc flag set to true', () => {
      describe('but no two-part tariff financial years', () => {
        it('does x', => {
        })

        it('does y', => {
        })
      })
    })
```

Because that describe block has two tests (`it()`) with a `beforeEach()` we are creating 2 licences when we could use the same licence in both tests.

test/services/licences/supplementary/process-imported-licence.service.test.js
```diff
@@ -0,0 +1,274 @@
...
+    Sinon.restore()
+    clock.restore()
+  })
+
+  describe('when given a valid licenceId', () => {
+    let licence
+    let importedLicence
+
+    beforeEach(async () => {
+      licence = await LicenceHelper.add()
```
Cruikshanks, line 42: If we are stubbing `FetchExistingLicenceDetailsService`, why do we need to create a real licence in the DB?

Cruikshanks (changes requested):

Cruikshanks (dismissed):

Cruikshanks (approved):

PR #1353, opened 2024-09-25
Create return cycle

robertparkinson:

As part of the work to replace the legacy system NALD we need to generate return logs. This PR creates a return cycle if one does not already exist and adds it to the return log


db/migrations/public/20240918125758_create-return-cycles-view.js
```diff
@@ -0,0 +1,26 @@
...
+  return knex
+    .schema
+    .createView(viewName, (view) => {
+      view.as(knex('return_cycles').withSchema('returns').select([
+        'return_cycle_id AS id',
+        'start_date',
+        'end_date',
+        'due_date',
+        'is_summer',
+        'is_submitted_in_wrls',
```
Cruikshanks, line 15: In order to match our conventions we would drop the `is_` prefix from these fields.

```suggestion
        'is_summer AS summer',
        'is_submitted_in_wrls AS submitted_in_wrls',
```

test/support/helpers/return-cycle.helper.js
```diff
@@ -0,0 +1,70 @@
...
+const { generateUUID } = require('../../../app/lib/general.lib.js')
+const { timestampForPostgres } = require('../../../app/lib/general.lib.js')
+const ReturnCycleModel = require('../../../app/models/return-cycle.model.js')
+
+/**
+ * Add a new return cycle
+ *
+ * If no `data` is provided, default values will be used. These are
+ *
+ * - `id` - UUID
```
Cruikshanks, line 16:
```suggestion
 * - `id` - [random UUID]
```

Cruikshanks (changes requested):
I have some suggested changes from a first pass. Give me a ping if you have any questions/challeneges.

test/services/jobs/return-logs/fetch-return-cycle.service.test.js
```diff
@@ -0,0 +1,76 @@
...
+  })
+
+  describe('when summer is true and the date is after the current cycle', () => {
+    it('should return an empty array', async () => {
+      const result = await FetchReturnCycleService.go('3021-09-01', true)
+
+      expect(result).to.be.undefined()
+    })
+  })
+})
```
Cruikshanks, line 76: Perhaps

```suggestion
describe('Fetch return cycle service', () => {
  let currentAllYearReturnCycle
  let currentSummerReturnCycle
  let previousAllYearReturnCycle
  let previousSummerReturnCycle
  let summer

  before(async () => {
    currentAllYearReturnCycle = await ReturnCycleHelper.select(3)
    previousAllYearReturnCycle = await ReturnCycleHelper.select(5)
    currentSummerReturnCycle = await ReturnCycleHelper.select(2)
    previousSummerReturnCycle = await ReturnCycleHelper.select(4)
  })

  describe('when "summer" is false', () => {
    before(() => {
      summer = false
    })

    describe('and the date is after the end of april', () => {
      it('should return the correct all year log cycle UUID', async () => {
        const result = await FetchReturnCycleService.go('2021-05-01', summer)
  
        expect(result).to.equal(currentAllYearReturnCycle.id)
      })
    })
  
    describe('and the date is before the end of april', () => {
      it('should return the correct all year log cycle UUID', async () => {
        const result = await FetchReturnCycleService.go('2021-01-01', summer)
  
        expect(result).to.equal(previousAllYearReturnCycle.id)
      })
    })
  
    describe('and the date is after the current cycle', () => {
      it('should return an empty array', async () => {
        const result = await FetchReturnCycleService.go('3031-01-01', summer)
  
        expect(result).to.be.undefined()
      })
    })
  })

  describe('when "summer" is true', () => {
    before(() => {
      summer = true
    })

    describe('and the date is after the end of october', () => {
      it('should return the correct summer log cycle UUID', async () => {
        const result = await FetchReturnCycleService.go('2021-12-01', summer)
  
        expect(result).to.equal(currentSummerReturnCycle.id)
      })
    })
  
    describe('and the date is before the end of october', () => {
      it('should return the correct summer log cycle UUID', async () => {
        const result = await FetchReturnCycleService.go('2021-09-01', summer)
  
        expect(result).to.equal(previousSummerReturnCycle.id)
      })
    })
  
    describe('and the date is after the current cycle', () => {
      it('should return an empty array', async () => {
        const result = await FetchReturnCycleService.go('3021-09-01', summer)
  
        expect(result).to.be.undefined()
      })
    })
  })
})
```

Cruikshanks (changes requested):
Just some minor suggested changes

PR #1355, opened 2024-09-25
Import addresses data for a licence

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4666

This change will use the nald licence id and region to update the addresses data. This will insert all the addresses found for a licence ref.


test/support/helpers/address.helper.js
```diff
@@ -61,12 +61,32 @@ function defaults (data = {}) {
   }
 }
 
+/**
+ * Generate an Upr for an address
+ *
+ * @returns {string} - A random upr
```
Cruikshanks, line 67: So close!

```suggestion
 * Generate an UPRN for an address
 *
 * @returns {string} - A random UPRN
```

test/support/helpers/address.helper.js
```diff
@@ -61,12 +61,32 @@ function defaults (data = {}) {
...
+ */
 function generateUprn () {
   return randomInteger(100, 999999)
 }
 
+/**
+ * Generate an address external id
+ *
+ * This is build from NALD import data using the region code and address id
```
Cruikshanks, line 76:
```suggestion
 * This is built from NALD import data using the region code and address id
```

Cruikshanks (changes requested):
Just 2 really small NITS!

Cruikshanks (approved):

PR #1359, opened 2024-09-25
Alter 2PT agreement logic in rtn req setup journey

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4672

> Part of the work to migrate return versions from NALD to WRLS

When creating a new return version, if you select the 'Start with abstraction data' option, the service will generate the return version based on data held against the licence. If the licence has a two-part tariff agreement in place, we mark each return requirement generated as having a two-part tariff agreement.

However, UAT has found this approach too 'blunt'. We're marking return requirements as two-part tariffs even when the purpose linked to them is not.

Now, we only mark the requirement as having a two-part tariff agreement if the purpose it is linked to is also flagged as two-part tariff. This only applies when generating the return version from abstraction data against the licence.


test/services/return-requirements/setup/generate-from-abstraction-data.service.test.js
```diff
@@ -124,6 +116,25 @@ describe('Return Requirements - Generate From Abstraction Data service', () => {
       })
     })
 
+    describe('and the licence has a "current" purpose that is two-part tariff', () => {
+      beforeEach(() => {
+        fetchResult = _fetchResult(licenceId)
+        // The 3rd licence version purpose is already linked to a two-part tariff purpose. We set the second to be true
+        // was well just to emphasise this is what is driving the logic in the service.
```
Jozzey, line 123:
```suggestion
        // as well just to emphasise this is what is driving the logic in the service.
```

Jozzey (changes requested):
Looks good, just a typo

Jozzey (approved):
👍🏼

PR #1371, opened 2024-09-30
Import company address data for a licence

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4669

This change will use the nald licence id and region to update the addresses' data. This will insert or update the company addresses record.

This is not a concept in nald and is created for WRLS to link an address to company.


app/services/import/legacy/transform-company-addresses.service.js
```diff
@@ -0,0 +1,39 @@
...
+ */
+async function go (regionCode, licenceId, transformedCompanies) {
+  const naldAddresses = await FetchCompanyAddressesService.go(regionCode, licenceId)
+
+  naldAddresses.forEach((naldAddress) => {
+    const matchingCompany = _matchingCompany(transformedCompanies, naldAddress)
+
+    const address = CompanyAddressPresenter.go(naldAddress)
+
+    matchingCompany.companyAddresses = [...(matchingCompany?.companyAddresses || []), address]
```
Cruikshanks, line 27: I'm not sure we should be playing it safe here. If we somehow got an address back from the FetchCompanyAddressesService, that didn't match to a company we'd found in an earlier step, something has gone wrong (I believe).

I'd rather it blow up and we go looking, than the issue be quietly swallowed.

app/services/import/legacy/transform-company-addresses.service.js
```diff
@@ -0,0 +1,39 @@
+'use strict'
+
+/**
+ * Transforms NALD  data into a valid object that matches the WRLS structure for a company address
```
Cruikshanks, line 4:
```suggestion
 * Transforms NALD data into a valid object that matches the WRLS structure for a company address
```

app/services/import/legacy/transform-company-addresses.service.js
```diff
@@ -0,0 +1,39 @@
...
+ * @module ImportLegacyTransformCompanyAddressesService
+ */
+
+const CompanyAddressPresenter = require('../../../presenters/import/legacy/company-address.presenter.js')
+const FetchCompanyAddressesService = require('./fetch-company-address.service.js')
+
+/**
+ * Transforms NALD  data into a valid object that matches the WRLS structure for a company address
```
Cruikshanks, line 12:
```suggestion
 * Transforms NALD data into a valid object that matches the WRLS structure for a company address
```

Cruikshanks (changes requested):
If you put aside the NITs, it's more some double-checking questions. Give me a ping if it will be easier to chat through them rather than type out responses.

app/services/import/legacy/transform-company-addresses.service.js
jonathangoulding: change addresses

app/presenters/import/legacy/company-address.presenter.js
```diff
@@ -0,0 +1,54 @@
...
+ * @private
+ */
+function _endDate (address) {
+  const oldestDate = [address.end_date, address.lapsed_date, address.expired_date, address.revoked_date]
+    // Removes any null values
+    .filter((date) => { return date })
+    // Sorts in asc order
+    .sort((date1, date2) => { return date1 - date2 })
+    // We only want the oldest date
+    .slice(-1)[0]
```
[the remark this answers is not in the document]
jonathangoulding: Updated to handle if there are no dates 

```
function _endDate (address) {
  const endDates = [
    address.end_date,
    address.lapsed_date,
    address.expired_date,
    address.revoked_date
  ].filter((endDate) => {
    return endDate
  })

  const oldestDate = new Date(Math.max(...endDates))

  return !isNaN(oldestDate) ? oldestDate : null
}
```

Cruikshanks (approved):

PR #1379, opened 2024-10-03
Migrate abstraction details view page

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4705

As the legacy licence view page has now been rebuilt, the next process is to update some of the links pointing to legacy pages. One of these legacy pages is the abstraction details view. If we are able to move this over to our system, it will help towards removing the use of old legacy code.

Screenshot 2024-10-28 at 11 42 59

This PR is focused on rebuilding the abstraction details view page from the legacy code base.


app/presenters/licences/view-licence-purposes.presenter.js
```diff
@@ -0,0 +1,99 @@
...
+      return null
+    } else {
+      uniquePurposeDetails.add(licenceString)
+    }
+
+    return {
+      abstractionAmounts: _formatAbstractionAmounts(licenceVersionPurpose),
+      abstractionPeriod: _abstractionPeriod(licenceVersionPurpose),
+      abstractionPoints: _formatAbstractionPoints(licenceVersionPurpose.points),
+      purposeDescription: licenceVersionPurpose.purpose.description ? licenceVersionPurpose.purpose.description : ''
```
rvsiyad, line 47: Can probably remove the check if exists since it is a set value in the purpose table so is never empty!

app/presenters/licences/view-licence-purposes.presenter.js
```diff
@@ -0,0 +1,99 @@
...
+  if (dailyQuantity) {
+    details.push(`${parseFloat(dailyQuantity).toFixed(2)} cubic metres per day`)
+  }
+
+  if (hourlyQuantity) {
+    details.push(`${parseFloat(hourlyQuantity).toFixed(2)} cubic metres per hour`)
+  }
+
+  if (instantQuantity) {
+    details.push(`${parseFloat(instantQuantity).toFixed(2)} cubic metres per second (Instantaneous Quantity)`)
```
rvsiyad, line 91: The legacy code used the statement "cubic litres per second (Instantaneous Quantity)". Since i borrowed this code from somewhere in system, I assume this has been changed now?

Cruikshanks (changes requested):
Think the feedback would be easier to go through in a pairing session. So contact me when you see this 😉

Cruikshanks (approved):

PR #1387, opened 2024-10-07
Refactor the import legacy persist logic

jonathangoulding:
We have recently introduced functionality to import data from NALD using for a licence and companies for that particular licence.

This change will break out the persist functionality into logical groups of functionality.


jonathangoulding:
I think it's worth having the big test to test the complete transaction. 

Do we want to add duplicate tests per persist service ? 

Worth a discussion.

app/services/import/persist-import.service.js
```diff
@@ -0,0 +1,38 @@
+'use strict'
+
+/**
+ * Creates or updates an imported licence and its child entities that have been transformed and validated
+ * @module PersistImportService
+ */
+
+const PersistImportService = require('./persist/persist-licence.service.js')
```
Cruikshanks, line 8: If that is what we are naming the file we should name the value to match (and think this is the right way round!)

```suggestion
const PersistLicenceService = require('./persist/persist-licence.service.js')
```

app/services/import/legacy/process-licence.service.js
```diff
@@ -6,7 +6,7 @@
  */
 
 const LicenceStructureValidator = require('../../../validators/import/licence-structure.validator.js')
-const PersistLicenceService = require('../persist-licence.service.js')
+const PersistLicenceService = require('../persist-import.service.js')
```
Cruikshanks, line 9: If that is what we are naming the file we should name the value to match (and think this is the right way round!)

```suggestion
const PersistImportService = require('../persist-import.service.js')
```

app/services/import/persist/persist-company.service.js
```diff
@@ -0,0 +1,140 @@
...
+ * Creates or updates an imported company and its child entities that have been transformed and validated
+ *
+ * @param trx
+ * @param updatedAt
+ * @param {object} transformedCompanies - An object representing a valid WRLS licence
+ *
+ * @returns {Promise<string>} - the licence id from WRLS
+ */
+async function go (trx, updatedAt, transformedCompanies) {
+  await _persistCompanies(trx, updatedAt, transformedCompanies)
```
Cruikshanks, line 23: We can return this directly.

```suggestion
  return _persistCompanies(trx, updatedAt, transformedCompanies)
```

app/services/import/persist/persist-company.service.js
```diff
@@ -0,0 +1,140 @@
...
+const AddressModel = require('../../../models/address.model.js')
+const CompanyModel = require('../../../models/company.model.js')
+const ContactModel = require('../../../models/contact.model.js')
+const { db } = require('../../../../db/db.js')
+
+/**
+ * Creates or updates an imported company and its child entities that have been transformed and validated
+ *
+ * @param trx
+ * @param updatedAt
```
Cruikshanks, line 17: These JSDoc params need to be completed.

app/services/import/persist/persist-company.service.js
```diff
@@ -0,0 +1,140 @@
...
+ * @param {object} transformedCompanies - An object representing a valid WRLS licence
```
Cruikshanks, line 18: Think this comment needs to be corrected.

app/services/import/persist/persist-licence-versions.service.js
```diff
@@ -0,0 +1,113 @@
...
+
+const LicenceVersionPurposeConditionModel = require('../../../models/licence-version-purpose-condition.model.js')
+const LicenceVersionPurposeModel = require('../../../models/licence-version-purpose.model.js')
+const LicenceVersionModel = require('../../../models/licence-version.model.js')
+
+/**
+ * Creates or updates an imported licence's versions and its child entities that have been transformed and validated
+ *
+ * @param trx
+ * @param updatedAt
```
Cruikshanks, line 16: These need to be completed

app/services/import/persist/persist-licence-versions.service.js
```diff
@@ -0,0 +1,113 @@
...
+ * @param {object} transformedLicence - An object representing a valid WRLS licence
+ *
+ * @param id
```
Cruikshanks, line 19: Same here, and new line is in the wrong place.

app/services/import/persist/persist-company.service.js
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * @returns {Promise<string>} - the licence id from WRLS
```
Cruikshanks, line 20: This isn't correct. But also aware it's returning something we don't expect any calling module to use.

```suggestion
 * @returns {Promise} A promise is returned but it does not resolve to anything we expect the caller to use
```

app/services/import/persist/persist-licence.service.js
```diff
@@ -0,0 +1,45 @@
...
+ * @module PersistLicenceService
+ */
+
+const LicenceModel = require('../../../models/licence.model.js')
+
+/**
+ * Creates or updates an imported licence that have been transformed and validated
+ *
+ * @param trx
+ * @param updatedAt
```
Cruikshanks, line 14: These need to be completed

test/services/import/persist-licence.service.test.js
Cruikshanks, line 1: File and main description need to match the module being tested.

test/services/import/persist-licence.service.test.js
```diff
@@ -67,80 +72,97 @@ describe('Persist licence service', () => {
...
+      beforeEach(async () => {
+        // Call the thing under test
+        result = await PersistImportService.go(transformedLicence, transformedCompanies)
+
+        // Get the persisted data
+        newLicence = await _fetchPersistedLicence(transformedLicence.licenceRef)
+        company = await _fetchPersistedCompany(transformedCompany.externalId)
+        contact = await _fetchPersistedContact(transformedCompany.externalId)
+        address = await _fetchPersistedAddress(transformedCompany.addresses[0].externalId)
+      })
```
Cruikshanks, line 84: Now this is placed here, and the test has been broken out into subtests, does this not mean we are first inserting, then repeatedly updating and fetching before each test?

Would this be better set as a `before()` block, so it only gets called once?

test/services/import/legacy/process-licence.service.test.js
```diff
@@ -13,7 +13,7 @@ const { generateUUID } = require('../../../../app/lib/general.lib.js')
 const { generateLicenceRef } = require('../../../support/helpers/licence.helper.js')
 
 // Things to stub
-const PersistLicenceService = require('../../../../app/services/import/persist-licence.service.js')
+const PersistLicenceService = require('../../../../app/services/import/persist-import.service.js')
```
Cruikshanks, line 16: Thing being stubbed needs to match name of module being imported.

```suggestion
const PersistImportService = require('../../../../app/services/import/persist-import.service.js')
```

Cruikshanks:
> I think it's worth having the big test to test the complete transaction.
> 
> Do we want to add duplicate tests per persist service ?
> 
> Worth a discussion.

Looks like we need a discussion! For me, the main driver for this was just how big and complex the tests were becoming, which is a 'smell' that the service under test is doing too much.

So, I would expect to see tests specifically for the things we have broken out, and `persist-import.service.test.js` to feature a lot of stubbing of those child services.

The child services can then focus on ensuring they behave as expected when inserting or updating. In `PersistImportService` we could, for example, start looking at how it behaves when one of those child services errors, i.e. confirm that nothing gets persisted. 🤷

Cruikshanks (changes requested):

app/services/import/persist/persist-licence-versions.service.js
```diff
@@ -0,0 +1,112 @@
...
+ * Creates or updates an imported licence's versions and its child entities that have been transformed and validated.
+ *
+ * @param {object} trx - An Objection.js transaction object for PostgreSQL.
+ * @param {string} updatedAt - The timestamp indicating when the entity was last updated.
+ * @param {object} transformedLicence - An object representing a valid WRLS licence.
+ * @param {string} licenceId - A licence id from WRLS
+ *
+ */
+async function go (trx, updatedAt, transformedLicence, licenceId) {
+  await _persistLicenceVersions(trx, updatedAt, transformedLicence.licenceVersions, licenceId)
```
Cruikshanks, line 22: We can just return this.

```suggestion
  return _persistLicenceVersions(trx, updatedAt, transformedLicence.licenceVersions, licenceId)
```

Cruikshanks (changes requested):
Just gonna ping these back and then will check in again later, as they'll also help make the changes a bit less confusing for me!

app/services/import/persist/persist-company.service.js
jonathangoulding: It does not return anything ?

Cruikshanks (approved):

PR #1389, opened 2024-10-07
Refactor logic for creating return cycles when importing new licence

robertparkinson:

As part of the work to replace the legacy system NALD we need to generate return logs. This PR creates a return cycle if one does not already exist when creating return logs for a new licence when it is imported. This will create the return cycles even if the date for them to be ran has not yet happened.

### DIFF
--- a/app/services/jobs/return-logs/generate-return-logs.service.js
+++ b/app/services/jobs/return-logs/generate-return-logs.service.js
@@ -24,18 +24,7 @@ const GenerateReturnCycleService = require('./generate-return-cycle.service.js')
  * @returns {Promise<Array>} the array of return log payloads to be created in the database
  */
 async function go (returnRequirements) {
-  const today = formatDateObjectToISO(new Date())
-
-  let allYearReturnCycleId = await FetchReturnCycleService.go(today, false)
-  let summerReturnCycleId = await FetchReturnCycleService.go(today, true)
-
-  if (!allYearReturnCycleId) {
-    allYearReturnCycleId = await GenerateReturnCycleService.go(false)
-  }
-
-  if (!summerReturnCycleId) {
-    summerReturnCycleId = await GenerateReturnCycleService.go(true)
-  }
+  const { allYearReturnCycleId, summerReturnCycleId } = await _fetchReturnCycleId()
 
   const returnLogs = returnRequirements.map(async (requirements) => {
     const startDate = _startDate(requirements.summer, requirements.returnVersion)
@@ -100,6 +89,26 @@ function _earliestDate (summer, returnVersion) {
   })
 
   return new Date(Math.min(...dates))
+}
+
+async function _fetchReturnCycleId () {
+  const today = formatDateObjectToISO(new Date())
+
+  let allYearReturnCycleId = await FetchReturnCycleService.go(today, false)
+  let summerReturnCycleId = await FetchReturnCycleService.go(today, true)
+
+  if (!allYearReturnCycleId) {
+    allYearReturnCycleId = await GenerateReturnCycleService.go(false)
+  }
+
+  if (!summerReturnCycleId) {
+    summerReturnCycleId = await GenerateReturnCycleService.go(true)
+  }
+
+  return {
+    allYearReturnCycleId,
+    summerReturnCycleId
+  }
 }
 
 function _id (requirements, startDate, endDate) {
--- a/app/services/jobs/return-logs/generate-return-logs.service.js
+++ b/app/services/jobs/return-logs/generate-return-logs.service.js
@@ -13,16 +13,30 @@ const {
   cycleStartDate,
   formatDateObjectToISO
 } = require('../../../lib/dates.lib.js')
+const FetchReturnCycleService = require('./fetch-return-cycle.service.js')
+const GenerateReturnCycleService = require('./generate-return-cycle.service.js')
 
 /**
  * Generates the payload for submission to the returns table.
  *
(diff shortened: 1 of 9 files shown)


app/services/jobs/return-logs/generate-return-logs.service.js
```diff
@@ -13,16 +13,30 @@ const {
...
+  let allYearReturnCycleId = await FetchReturnCycleService.go(today, false)
+  let summerReturnCycleId = await FetchReturnCycleService.go(today, true)
+
+  if (!allYearReturnCycleId) {
+    allYearReturnCycleId = await GenerateReturnCycleService.go(false)
+  }
+
+  if (!summerReturnCycleId) {
+    summerReturnCycleId = await GenerateReturnCycleService.go(true)
+  }
```
Cruikshanks, line 38: This has added a lot to the existing function and is distinctly different to what it was doing before. Therefore, I'd suggest moving this to a private function that is responsible for returning the return cycle IDs needed (whether they be fetched or generated). This then becomes a single line call in the `go()`.

test/models/return-cycle.model.test.js
```diff
@@ -50,8 +50,12 @@ describe('Return Cycle model', () => {
...
         expect(result.id).to.equal(testRecord.id)
 
-        expect(result.returnLogs[0]).to.be.an.instanceOf(ReturnLogModel)
-        expect(result.returnLogs[0]).to.equal(testReturnLog)
+        const expectedReturnLog = result.returnLogs.find((returnLog) => {
+          return returnLog.id === testReturnLog.id
+        })
+
+        expect(expectedReturnLog).to.be.an.instanceOf(ReturnLogModel)
+        expect(expectedReturnLog).to.equal(testReturnLog)
```
Cruikshanks, line 58: I can see why this was done. But when viewing the file as a whole to confirm it, I spotted we've chosen the wrong style of test.

The pattern we've followed is for a 'BelongsTo' or 'HasOne' relationship, when we should have followed the pattern for a 'HasMany'.

So, this would be a suitable time for a bit of housekeeping to correct the test pattern. I then also wonder if you could get the expect to work without needing the `find()` call by switching it to be `expect(result.returnlogs).to.include(testReturnLogs[0])`. We've done this in other model tests and it has worked fine.

test/services/jobs/return-logs/fetch-return-cycle.service.test.js
```diff
@@ -48,11 +50,30 @@ describe('Fetch return cycle service', () => {
...
+
+        stub.returns({
+          select: selectStub,
+          where: whereStub,
+          limit: limitStub,
+          first: firstStub
+        })
+      })
+
+      it('should create the return cycle and return its id', async () => {
```
Cruikshanks, line 69: I don't think it is doing that!

test/services/jobs/return-logs/fetch-return-cycle.service.test.js
```diff
@@ -78,11 +99,30 @@ describe('Fetch return cycle service', () => {
...
+
+        stub.returns({
+          select: selectStub,
+          where: whereStub,
+          limit: limitStub,
+          first: firstStub
+        })
+      })
+
+      it('should create the return cycle and return it', async () => {
```
Cruikshanks, line 118: Still not doing that!

Cruikshanks (changes requested):
First pass of the changes with my suggestions.

app/services/jobs/return-logs/generate-return-logs.service.js
```diff
@@ -102,6 +91,26 @@ function _earliestDate (summer, returnVersion) {
   return new Date(Math.min(...dates))
 }
 
+async function _fetchReturnCycleId () {
```
Cruikshanks, line 94: NIT!

```suggestion
async function _fetchReturnCycleIds () {
```

test/models/return-cycle.model.test.js
Cruikshanks, line 1: Close! This definitely means we can remove the `find()` in the middle of the test.

But the pattern still needs correcting. So, the top of the test becomes this.

```javascript
describe('Return Cycle model', () => {
  let testRecord
  let testReturnLogs

  before(async () => {
    testRecord = await ReturnCycleHelper.select()

    testReturnLogs = []
    for (let i = 0; i < 2; i++) {
      const returnLog = await ReturnLogHelper.add({ returnCycleId: testRecord.id })

      testReturnLogs.push(returnLog)
    }
  })
```

Then the test itself becomes this.

```javascript
  describe('Relationships', () => {
    describe('when linking to return logs', () => {
      it('can successfully run a related query', async () => {
        const query = await ReturnCycleModel.query()
          .innerJoinRelated('returnLogs')

        expect(query).to.exist()
      })

      it('can eager load the return logs', async () => {
        const result = await ReturnCycleModel.query()
          .findById(testRecord.id)
          .withGraphFetched('returnLogs')

        expect(result).to.be.instanceOf(ReturnCycleModel)
        expect(result.id).to.equal(testRecord.id)

        expect(result.returnLogs).to.be.an.array()
        expect(result.returnLogs[0]).to.be.an.instanceOf(ReturnLogModel)
        expect(result.returnLogs).to.include(testReturnLogs[0])
        expect(result.returnLogs).to.include(testReturnLogs[1])
      })
    })
  })
```

Cruikshanks (changes requested):
Tweaks to the changes!

Cruikshanks (changes requested):

Cruikshanks (approved):

PR #1390, opened 2024-10-08
Al's take on Licence import flagging supp. billing

Cruikshanks:

~There are so many dates to contend with, and possible scenarios to cover that providing guidance over a flaky video call is proving difficult. Also, I've not been focused or directly hands on with this (till this branch!) so I haven't been providing the clearest guidance.~

~So, this is me putting my money where my mouth is, and building on what has been done to try and provide a working 'take' on how this might be done a little simpler and with less processing.~

Now this covers suggested changes to the tests to reduce the number of times we create records in the DB, and remove duplication in the tests themselves.


Cruikshanks:
Closing this. The original has now been approved and merged so we no longer need this PR or branch.

PR #1392, opened 2024-10-08
Import licence document for a licence

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4669

WRLS has the concept of a licence document. NALD does not.

As part of the import process this licence document is created from the NALD licence. The start and end date use the licence and licence versions to calculate the earliest date for both.

This change will replicate this logic.


test/services/import/legacy/transform-licence-document.service.test.js
```diff
@@ -0,0 +1,76 @@
...
+      expect(transformedLicence.licenceDocument).to.equal({
+        deletedAt: null,
+        licenceRef,
+        endDate: null,
+        startDate: new Date('1999-01-01')
+      })
+    })
+  })
+
+  describe('when no matching legacy licence version is found', () => {
```
Cruikshanks, line 59:
```suggestion
  describe('when no matching legacy licence is found', () => {
```

test/services/import/persist/persist-licence-document.service.test.js
```diff
@@ -0,0 +1,120 @@
...
+
+  describe('when given a valid transformed licence document', () => {
+    describe('and that licence does not already exist', () => {
+      beforeEach(() => {
+        licenceDocument = _transformedLicenceDocument(licenceRef)
+
+        transformedLicence.licenceDocument = licenceDocument
+      })
+
+      it('creates a new licence record plus child records in WRLS and returns the licence ID', async () => {
```
Cruikshanks, line 50:
```suggestion
      it('creates a new licence document record', async () => {
```

test/services/import/persist/persist-licence-document.service.test.js
```diff
@@ -0,0 +1,120 @@
...
+
+      beforeEach(async () => {
+        const existing = await _createExistingRecords(licenceRef)
+
+        existingLicence.licenceDocument = existing.licenceDocument
+
+        transformedLicence.licenceDocument = _transformedLicenceDocument(licenceRef)
+      })
+
+      it('should return the updated licence', async () => {
```
Cruikshanks, line 75:
```suggestion
      it('should update the existing licence document record', async () => {
```

app/services/import/persist/persist-licence-document.service.js
```diff
@@ -0,0 +1,37 @@
...
+
+async function _persistLicenceDocument (trx, updatedAt, licenceDocument) {
+  return LicenceDocumentModel.query(trx)
+    .insert({ ...licenceDocument, updatedAt })
+    .onConflict('licenceRef')
+    .merge([
+      'endDate',
+      'startDate',
+      'updatedAt',
+      'deletedAt'
```
Cruikshanks, line 31: I'd have to dig into how `deletedAt` is determined (because I think that is set during the import as well).

But we shouldn't be overwriting this if it is already set.

Cruikshanks (changes requested):
A couple of copy and paste typos.

The one change I'd make is not to overwrite the `deletedAt` date when we upsert. But that also raises the question of how does the import currently determine `deletedAt`?

Once we know that we can make a decision as to whether it should be included in this change or added separately.

app/services/import/persist/persist-licence-document.service.js
jonathangoulding: I see what you mean now.

jonathangoulding:
```
-- auto-generated definition
create table documents
(
    document_id   uuid      default gen_random_uuid()                        not null
        primary key,
    regime        varchar   default 'water'::character varying               not null,
    document_type varchar   default 'abstraction_licence'::character varying not null,
    document_ref  varchar                                                    not null,
    start_date    date                                                       not null,
    end_date      date,
    external_id   varchar,
    date_created  timestamp default now()                                    not null,
    date_updated  timestamp default now()                                    not null,
    is_test       boolean   default false                                    not null,
    date_deleted  timestamp
);

```

I can remove `'deletedAt'` from the upsert.

PR #1424, opened 2024-10-18
Import licence document roles for a licence - licence holder

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4717

WRLS has the concept of a licence document. NALD does not.

As part of the import process a licence document is created from the NALD licence.

There are two roles we want to import returnsTo and licence holder.

This change will focus on the licence holder licence role.

Previous work to set handle import was done for the returnTo role here - https://github.com/DEFRA/water-abstraction-system/pull/1414


Cruikshanks (dismissed):

jonathangoulding:
This is not ready. 

The issue i have left to solve is: 

```
Each licence version with a an issue should be imported.

A licence version with multiple increments and the same issue number should be imported. 

However if that increment has the same start date as another increment on the same issue. Then the end date of the oldest should be used.
```

jonathangoulding:
> This is not ready.
> 
> The issue i have left to solve is:
> 
> ```
> Each licence version with a an issue should be imported.
> 
> A licence version with multiple increments and the same issue number should be imported. 
> 
> However if that increment has the same start date as another increment on the same issue. Then the end date of the oldest should be used.
> ```

This has been fixed.

Cruikshanks (approved):

PR #1425, opened 2024-10-18
Return log for ended licence

robertparkinson:

As part of the work to replace the legacy system NALD we need to generate return logs. This PR sets up the logic for voiding and reissuing of return logs when a new return version has been entered. It will use the new dates of the return version to figure out which return logs need to be voided, voids them and the creates new return logs to cover the return cycles from that period forward.


app/lib/dates.lib.js
```diff
@@ -9,6 +9,27 @@ const february = 2
...
+    .filter((date) => {
+      return date
+    })
+    .map((date) => {
+      return new Date(date)
+    })
+
+  const _earliestDate = new Date(Math.min(..._dates))
+
+  return _earliestDate
```
Cruikshanks, line 30: If we expect `dates` to be an array, we can work with it directly.

I think we can also split up the steps rather than do them in one pass. IMHO, I think it makes it clearer what is happening.

We can also directly return the result rather than temporarily instantiating a value.

Finally, having then looked at the tests, we hadn't covered what happens if the array we pass in is empty or contains all null values. The result was unexpected, which led me to add in specific handling for that scenario.

```suggestion
  const allEmptyValuesRemoved = dates.filter((date) => {
    return date
  })
  
  if (allEmptyValuesRemoved.length === 0) {
    throw Error('No dates provided to determine earliest')
  }

  const valuesAsDates = allEmptyValuesRemoved.map((date) => {
    return new Date(date)
  })

  const earliestDateTimestamp = Math.min(...valuesAsDates)

  return new Date(earliestDateTimestamp)
```

P.S. An `_` is a convention for marking a _function_ as private, i.e., it won't be exported from the module and is only intended for internal use. You don't typically prefix variables with `_`.

app/lib/dates.lib.js
```diff
@@ -9,6 +9,27 @@ const february = 2
 const lastDayOfFebruary = 28
 const lastDayOfFebruaryLeapYear = 29
 
+/**
+ * Given a return version and return cycle end date, return the earliest date as a string in the ISO format.
+ *
+ * @param {Array} returnVersion - The list of dates to check.
+ * @returns {String} - The earliest date as a string in the ISO format.
+ * @private
+ */
```
Cruikshanks, line 18: This doesn't match what the function is actually doing or what the tests describe it as doing. Plus the JSDoc schema is incorrect.

```suggestion
/**
 * From an array of dates as strings, filter out empty values and return the earliest as a `Date`
 *
 * This was created as part of our work on generating return logs for licences, and needing to work out the earliest
 * end date between the return version end date, and the return cycle end date for the given year.
 *
 * @param {string[]} dates - The dates from which to select the earliest
 *
 * @returns {Date} The earliest date string as a `Date`
 */
```

test/lib/dates.lib.test.js
```diff
@@ -138,4 +138,18 @@ describe('Dates lib', () => {
       expect(result).to.be.false()
     })
   })
+
+  describe('earliestDate', () => {
```
Cruikshanks, line 142: This needs to be the first section in the file, to match the alphabetical ordering, and the functions position in the source file.

test/lib/dates.lib.test.js
```diff
@@ -138,4 +138,18 @@ describe('Dates lib', () => {
...
+      const result = DateLib.earliestDate(['2025-04-01', '2025-03-31', '2025-03-30'])
+
+      expect(result).to.equal(new Date('2025-03-30'))
+    })
+
+    it('should be able to handle null valuse and still return the earliest date', () => {
+      const result = DateLib.earliestDate(['2025-04-01', null, '2025-03-31', '2025-03-30'])
+
+      expect(result).to.equal(new Date('2025-03-30'))
+    })
```
Cruikshanks, line 153: I think the function as it is, suggests some additional tests we could cover in order to fully describe the behaviour we expect.

```suggestion
    let dates

    describe('given an array of date strings', () => {
      beforeEach(() => {
        dates = ['2025-04-01', '2025-03-30', '2025-03-31']
      })

      it('returns the earliest as a Date value', () => {
        const result = DateLib.earliestDate(dates)

        expect(result).to.equal(new Date('2025-03-30'))
      })
    })

    describe('given an array of date strings that contains null and undefined values', () => {
      beforeEach(() => {
        dates = ['2025-04-01', null, '2025-03-30', undefined]
      })

      it('still returns the earliest as a Date value', () => {
        const result = DateLib.earliestDate(dates)

        expect(result).to.equal(new Date('2025-03-30'))
      })
    })

    describe('given an array of date strings that only contains null and undefined values', () => {
      beforeEach(() => {
        dates = [null, undefined]
      })

      it('throws an error', () => {
        expect(() => {
          return DateLib.earliestDate(dates)
        }).to.throw('No dates provided to determine earliest')
      })
    })

    describe('given an empty array', () => {
      beforeEach(() => {
        dates = []
      })

      it('throws an error', () => {
        expect(() => {
          return DateLib.earliestDate(dates)
        }).to.throw('No dates provided to determine earliest')
      })
    })
```

app/lib/dates.lib.js
```diff
@@ -9,6 +9,27 @@ const february = 2
...
+function earliestDate(dates) {
```
Cruikshanks, line 19: Can I suggest a tweak to the name, to make it a bit more descriptive?

```suggestion
function determineEarliestDate(dates) {
```

Looking at `app/lib/general.lib.js` probably gives a better feel/example of the naming of helper functions like this.

Cruikshanks (changes requested):
This is just the feedback for `earliestDate()`.

PR #1426, opened 2024-10-18
Update some unit tests to use Model stubbing

Jozzey:
We have found a pattern we can use to stub Objection Models in our unit tests.

In this PR, we are going to use the model stubbing in some unit tests where it makes sense so that we have some examples of the pattern being used in our code for future reference.


Jozzey:
I didn't find any tests in which to stub models that would clearly demonstrate how the stubbing works. The ones I found are too simple, and the complex ones are generally in their own service and need to be tested properly.

I'm therefore going to document model stubbing in our [Team repo](https://github.com/DEFRA/water-abstraction-team) for future reference.

Beckyrose200 (approved):

robertparkinson (approved):

PR #1435, opened 2024-10-21
Licence Import Supplementary billing test endpoint

Beckyrose200:

Due to the fact the above ticket is happening during the licence import (which is automated), we need to add a test endpoint to allow the QA team to test the new supplementary flagging functionality. The process of flagging a licence for supplementary billing only happens if there is a new end date added to the licence (lapsed, expired or revoked). The imported licence is looking at NALD data, so built into our testing endpoint is functionality that means you can pass a new licence end date and it will act as though that is the imported NALD data.


Beckyrose200:
Ignore Sonar Cloud complaining, there are now tests for this endpoint as it will only get deleted once the ticket has been tested by QA

Cruikshanks (changes requested):
This needs to go on the `/check` route and controller. That is the one we have designated for testing.

app/controllers/check.controller.js
```diff
@@ -5,10 +5,37 @@
...
+async function flagForBilling (request, h) {
+  const { licenceId, expiredDate, lapsedDate, revokedDate } = request.payload
+
+  const transformedLicence = {
+    expiredDate,
+    lapsedDate,
+    revokedDate
+  }
+
+  DetermineSupplementaryBillingFlagsService.go(transformedLicence, licenceId)
```
Cruikshanks, line 34: Does this need to be awaited? As this is for testing (we don't worry about hitting our 1 sec or less response times), plus we're just doing a single licence so should be fine anyway.

Probably best so the person testing doesn't expect the change has happened as soon as they get a response.

app/controllers/check.controller.js
```diff
@@ -5,10 +5,37 @@
...
+
+  return h.response().code(redirectStatusCode)
```
Cruikshanks, line 36: Not your fault as was already named this, but `204` is not a redirect status code. That should be `NO_CONTENT_STATUS_CODE` (because it is a constant as well as being incorrectly named!) https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204

Cruikshanks (changes requested):

Cruikshanks (approved):

PR #1436, opened 2024-10-21
create return cycle dates library

robertparkinson:

As part of the work to replace the legacy system NALD we need to generate return logs and return cycles. This PR creates a library for calculating dates related to return cycles and the unit tests to cover them.


app/lib/return-cycle-dates.lib.js
```diff
@@ -0,0 +1,207 @@
...
+
+    return formatDateObjectToISO(new Date(`${year}-${returnCycleDates.summer.dueDate.month + 1}-${returnCycleDates.summer.dueDate.day}`))
+  }
+
+  if (month > returnCycleDates.allYear.endDate.month) {
+    return formatDateObjectToISO(new Date(`${year + 1}-${returnCycleDates.allYear.dueDate.month + 1}-${returnCycleDates.allYear.dueDate.day}`))
+  }
+
+  return formatDateObjectToISO(new Date(`${year}-${returnCycleDates.allYear.dueDate.month + 1}-${returnCycleDates.allYear.dueDate.day}`))
+}
```
Cruikshanks, line 71: This applies to a few of the methods in this file. Our linter is not picking this up but some of the lines have gone beyond 120 chars. I think we can resolve this and make things a bit clearer at the same time. My suggestion would be the following.

```suggestion
function cycleDueDateByDate (date, summer) {
  const year = date.getFullYear()
  const month = date.getMonth()

  let cycleDueYear
  let cycleDueMonth
  let cycleDueDay

  if (summer) {
    cycleDueDay = returnCycleDates.summer.dueDate.day
    cycleDueMonth = returnCycleDates.summer.dueDate.month + 1

    cycleDueYear = month > returnCycleDates.summer.endDate.month ? year + 1 : year
  } else {
    cycleDueDay = returnCycleDates.allYear.dueDate.day
    cycleDueMonth = returnCycleDates.allYear.dueDate.month + 1

    cycleDueYear = month > returnCycleDates.allYear.endDate.month ? year + 1 : year
  }

  const cycleDueDate = new Date(`${cycleDueYear}-${cycleDueMonth}-${cycleDueDay}`)

  return formatDateObjectToISO(cycleDueDate)
}
```

But open to other ideas. I think the current pattern is based on my harping on about the 'guard clause' pattern. But I think this is a good example of an exception to that pattern.

test/lib/return-cycle-dates.lib.test.js
```diff
@@ -0,0 +1,562 @@
...
+
+const { returnCycleDates } = require('../../app/lib/static-lookups.lib.js')
+
+// Thing under test
+const ReturnCycleDatesLib = require('../../app/lib/return-cycle-dates.lib.js')
+
+describe('Return Cycle Dates lib', () => {
+  const today = new Date()
+  const year = today.getFullYear()
+  const month = today.getMonth()
```
Cruikshanks, line 18: It looks like we have to amend our assertions based on what the current date is.

I'd rather make this explicit in the tests that we have to do this by

- controlling what 'today' is in the tests
- having an explicit `describe()`, for example, `"when 'todays' month is greater than the return cycle end date's month"`

Take a look at `let clock = Sinon.useFakeTimers(testDate)` (we use it in some of our tests, for example, `test/lib/general.lib.test.js`) to understand how we can control the date and time within a test.

Cruikshanks (changes requested):
Thank you for splitting this out. It's made reviewing the changes much easier.

And this means I have some feedback and suggested changes. Do let me know if you have any issues with them, or challenges.

Cruikshanks (approved):

PR #1440, opened 2024-10-23
Fix testing endpoint for supplementary flags

Beckyrose200:

Recently, a PR was raised to create a testing endpoint for flagging a licence for supplementary billing during the import service. This process looks at the licence end dates and determines what flags should be added to the licence. During testing, the feature wasn't working as expected. 

After some investigation, it turns out that the testing endpoint wasn't configured correctly. When the arguments are passed in the body of the request, it converts everything to a String. This means the data we are mocking is in a `String` data type and not a `Date` data type. We need the object to be a `Date` type as during the import that is what would be fetched from the db. 

This PR fixed the object before being passed to the supplementary flagging process.

### DIFF
--- a/app/controllers/check.controller.js
+++ b/app/controllers/check.controller.js
@@ -26,9 +26,9 @@ async function flagForBilling (request, h) {
   const { licenceId, expiredDate, lapsedDate, revokedDate } = request.payload
 
   const transformedLicence = {
-    expiredDate,
-    lapsedDate,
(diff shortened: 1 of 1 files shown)


robertparkinson (approved):

Cruikshanks (approved):
Can you update the PR description with a link to the relevant Jira ticket, and ensure it is included as part of the squashed merge?

All good though!

PR #1444, opened 2024-10-24
Refactor billing account contact details to model

Cruikshanks:

We currently display a billing account's contact details when viewing a bill in a bill run. This is achieved by the `FetchBillingAccountService` and the `ViewBillPresenter`.

We have to show the _exact_ same details in the _exact_ same format when reviewing the return match and allocation results in a two-part tariff bill run.

We strongly suspect that more pages will need to display this in the coming months.

We've already started a [tidy up of our two-part tariff review code](https://github.com/DEFRA/water-abstraction-system/pull/1443). Moving the fetch _and_ the presentation logic to somewhere reusable will help and put us in a better position for future changes.

Previously, when we needed to fetch various pieces of data to make a determination about something on a model, we've done it on the model directly using modifiers and instance methods.

This change refactors our existing billing account detail logic to follow this pattern.


Cruikshanks:
I used my executive powers 💪 to merge even though CI was failing! We have a [known issue](https://github.com/DEFRA/water-abstraction-system/pull/1445) with 4 of the return cycle tests that is being resolved.

PR #1451, opened 2024-10-28
Migrate licence points view page

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4727

Having migrated the view licence page to our new code base, we are starting to look at the various pages linked from it.

Migrating the legacy pages not only means we can better support the service, but it is also a necessary step to updating to the latest GDS design system components and then carrying out an accessibility audit. Next on our hit list is the Licence points view page which is linked from the summary tab.

This PR is focused on rebuilding the licence points view page from the legacy code base.


app/presenters/licences/view-licence-points.presenter.js
```diff
@@ -0,0 +1,56 @@
...
+    licenceRef: licence.licenceRef,
+    pageTitle: 'Licence abstraction points'
+  }
+}
+
+function _formatLicencePoints (points) {
+  return points.map((point) => {
+    const pointInstance = PointModel.fromJson(point)
+
+    return {
```
rvsiyad, line 34: SonarCloud is complaining about the functions complexity with the added ternary operations going on. I couldn't think of a better way to do this, but I'm open to suggestions.

StuAA78: We could try using the nullish coalescing operator `??` here:

```js
    return {
      bgsReference: pointInstance.bgsReference ?? '',
      category: pointInstance.category ?? '',
      // etc...
    }
```

If the first thing isn't `null` or `undefined` then it gives us the first thing, otherwise it gives us the second thing, so effectively shorthand for what we've got here. It may keep SonarCloud happier, or it may not!

StuAA78 (changes requested):
Looks good, just got a small suggestion on something that will streamline the logic a little and may keep SonarCloud happy 😁

StuAA78 (approved):

PR #1453, opened 2024-10-29
New 'import' job to trigger return log and supplementary flag checks

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4728

This change adds a new 'import' job into the repo to trigger the work previously done to bring the legacy import logic.

This change adds a new endpoint to be triggered from jenkins and will use the existing process licence service.

There are a lot of licence refs to iterate over so a batching mechanism is used to reduce the load on the system.


app/services/jobs/import/import-licence.service.js
```diff
@@ -0,0 +1,35 @@
...
+
+const FetchLicences = require('./fetch-licences.service.js')
+const ProcessImportLicence = require('./process-import-licence.service.js')
+const { calculateAndLogTimeTaken, currentTimeInNanoseconds } = require('../../../lib/general.lib.js')
+
+/**
+ * Extracts and imports licence from NALD
+ *
+ * If a licence in NALD does not have a status of DRAFT, and at least one non-draft licence version
+ * then it will be extracted
```
Cruikshanks, line 16: I think this could be expanded. So, disregarding that triggering the endpoint from `water-abstraction-import` led to a DDOS attack (!) we can focus on what the job is actually for.

```suggestion
 * Processes NALD licences due for import on a nightly basis 
 *
 * Overnight the {@link https://github.com/DEFRA/water-abstraction-import | water-abstraction-import} app imports each
 * abstraction licence found in NALD. New licences are added, existing ones are updated.
 * 
 * When an existing licence is imported, there are some additional processes we need to trigger. The first is
 * `DetermineSupplementaryBillingFlagsService`, which checks whether a licence needs to be flagged for supplementary
 * billing if a change has been made.
 * 
 * For the same reason, `ProcessLicenceReturnLogsService` needs to check if any changes need to be made to a licence's
 * return logs.
 * 
 * Because we are migrating from the legacy apps we couldn't trigger these processes in **water-abstraction-import**.
 * Instead, we have this job. The one caveat is that it needs to be scheduled to run _after_ the
 * {@link https://github.com/DEFRA/water-abstraction-team/blob/main/jobs/import.md#nald-import | NALD import job},
 * (specifically once the NALD data has been extracted and imported into the `import` schema), but _before_ the
 * {@link https://github.com/DEFRA/water-abstraction-team/blob/main/jobs/import.md#nald-import | Licence import job}
 * runs. This is so these processes can see the differences between the NALD licence record and ours, to determine
 * if and what they need to do.
 *
 * > If a licence in NALD does not have a status of DRAFT, and at least one non-draft licence version then it will be
 * excluded
```

app/services/jobs/import/process-import-licence.service.js
Cruikshanks, line 1: If what this service is doing were to remain, then I'd suggest that the file be renamed to `process-import-licences.service.js`.

However, I'd suggest keeping it as it is but look to move the batching bit to `import-licence.service.js` and leaving this here to handle the processing of each individual licence.

```javascript
async function go (licence) {
  const endDates = {
    expiredDate: licence.expired_date,
    lapsedDate: licence.lapsed_date,
    revokedDate: licence.revoked_date
  }

  await DetermineSupplementaryBillingFlagsService.go(endDates, licence.id)

  await ProcessLicenceReturnLogsService.go(licence.id)
}
```

We could then even wrap the two calls in a try/catch, so should an error happen

- we can zero in on which licence caused the error
- we can try and prevent one licence failing from stopping others being imported

app/controllers/jobs.controller.js
```diff
@@ -62,13 +63,20 @@ async function returnLogs (request, h) {
...
   ProcessReturnLogsService.go(cycle, licenceReference)
 
-  return h.response().code(redirectStatusCode)
+  return h.response().code(NO_CONTENT_STATUS_CODE)
+}
+
+async function importLicence (_request, h) {
+  ImportLicence.go()
+
+  return h.response().code(NO_CONTENT_STATUS_CODE)
```
Cruikshanks, line 72: Alphabety-spaghetti wise, I think this needs to go higher in the controllers.

app/controllers/jobs.controller.js
```diff
@@ -62,13 +63,20 @@ async function returnLogs (request, h) {
...
 }
 
 module.exports = {
   exportDb,
   licenceUpdates,
   returnLogs,
   sessionCleanup,
-  timeLimited
+  timeLimited,
+  importLicence
```
Cruikshanks, line 81: Same comment about ordering

Cruikshanks (changes requested):
There is some general cleanup around comments and ordering here.

The key suggestion I have though is about changing what work the 2 services do. Any questions or challenges on this do give me a ping.

jonathangoulding:
Refactor added to move the process import licences into the import licences service.

app/services/jobs/import/import-licences.service.js
```diff
@@ -0,0 +1,111 @@
...
+ * When we process the licences we want to batch them into smaller chunks in an attempt to make the
+ * import process as efficient as possible
+ *
+ *
+ * @param {object[]} licences - An array of licences
+ *
+ * @private
+ */
+async function _processLicences (licences) {
+  // for (let i = 0; i < licences.length; i += batchSize) {
```
jonathangoulding, line 73: Remove if p-map is going to be used.

jonathangoulding:
Maybe merge https://github.com/DEFRA/water-abstraction-system/pull/1538 in first. 

I can not remember the reason we have not merged this.

app/controllers/jobs.controller.js
```diff
@@ -54,13 +61,22 @@ async function returnLogs(request, h) {
     return h.response().code(NOT_FOUND_STATUS_CODE)
   }
 
+  let licenceReference
+
   ProcessReturnLogsService.go(cycle)
 
+  if (h.request.payload !== null && h.request.payload.licenceReference) {
```
jonathangoulding, line 68: I think this might be a merge issue ? 

i am not in the git blame @robertparkinson

robertparkinson: yeah that should't be there

robertparkinson (approved):

PR #1458, opened 2024-10-31
Spike into using the node test runner

jonathangoulding:
This is just a quick spike to see the effort to replace the '@hapi/lab' test runner with 'node:test'.

We can initially (or forver) keep sinon.js as the mocking tool and '@hapi/code' for the assertion library.

### DIFF
--- a/test/errors/expanded.error.test.js
+++ b/test/errors/expanded.error.test.js
@@ -27,10 +27,6 @@ describe('ExpandedError', () => {
     it('will assign those to the error instance', () => {
       const result = new ExpandedError('My test error', additionalData)
 
-      // Node test runner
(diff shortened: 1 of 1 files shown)


jonathangoulding:
Basically replaced all occurrences of `exports.lab = Lab.script()` with ` require('node:test')`

jonathangoulding:
Can look into the not yet stable (in v20) coverage reporting

https://nodejs.org/docs/latest-v20.x/api/test.html#collecting-code-coverage

jonathangoulding:
Some test are stalling. It is consistent but not obvious whats wrong. Needs some investigating.

Some in the CI and some locally.

PR #1464, opened 2024-11-05
Add quarterly returns submissions to return versions

jonathangoulding:

As part of the quarterly returns submissions work, we need to add a new checkbox to the additional-submission-options page for the Quarterly returns submissions.

This has introduced some logic to determine if the return version is for a water company and if the return version start date is beyond the start of quarterly returns.


jonathangoulding:
This pr tackles part of https://eaflood.atlassian.net/browse/WATER-4707 to add the quarterly returns option to the check page and handle the changing default logic.

PR #1465, opened 2024-11-05
Add default for multiple upload on return version journey

jonathangoulding:

As part of the quarterly returns submissions work, we need to default the multiple uploads additional option to true when the licence is for a water company.


test/presenters/return-versions/setup/additional-submission-options.presenter.test.js
```diff
@@ -39,7 +39,8 @@ describe('Return Versions Setup - Additional Submission Options presenter', () =
       expect(result).to.be.equal({
         backLink: `/system/return-versions/setup/${session.id}/check`,
         licenceId: '8b7f78ba-f3ad-4cb6-a058-78abc4d1383d',
-        additionalSubmissionOptions: [],
+        multipleUpload: undefined,
```
Cruikshanks, line 42: I don't think this is correct anymore. `multipleUpload` is being set in `InitiateService`, so it will not be `undefined`.

We should probably update the 'session' object in this test (and other test files) to reflect this.

test/presenters/return-versions/setup/additional-submission-options.presenter.test.js
```diff
@@ -54,24 +55,70 @@ describe('Return Versions Setup - Additional Submission Options presenter', () =
     })
   })
 
-  describe('the "additionalSubmissionOptions" property', () => {
-    describe('when the user has previously submitted additional submission options', () => {
+  describe('the "multipleUpload" property', () => {
+    describe('when a user has selected "multipleUpload" for additional options', () => {
```
Cruikshanks, line 59: We seem to have dropped the word 'previously' from these tests. It was there to try and highlight where a user has previously visited this page and submitted an option, then we expect the presenter to replay it.

Happy to change it to something else if it's not making that clear. But I think just dropping it makes it unclear we are talking about selections a user previously made, instead of ones they have just made.

Cruikshanks (changes requested):
Apologies, running out of time. But I have this feedback so far so stopping here in order it can be digested, queried and challenged.

test/presenters/return-versions/setup/additional-submission-options.presenter.test.js
jonathangoulding: Good spot i have done this else where (i think). 

Add multiple upload to default to false

Cruikshanks (dismissed):

Cruikshanks (approved):

PR #1467, opened 2024-11-07
Add quarterly returns to return versions table

jonathangoulding:

As part of the work to introduce quarterly returns in the system repo we need to add a column to mark a return version for quarterly returns.

The work to handle this logic is done in the system repo.

This change is just to introduce the quarterly returns in the previous water tables.


Cruikshanks (approved):

Cruikshanks:
@jonathangoulding I'd hold off merging until the water-abstraction-service change is merged though.

PR #1469, opened 2024-11-08
Fix broken return version start date page

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4687

In [Decouple start date logic](https://github.com/DEFRA/water-abstraction-system/pull/1461) we made a change to support adding quarterly returns.

It was determined the logic for working out the start date from the values entered by the user was being duplicated in other places. The change ensured the logic stays in just the `SubmitStartDateService`, and added a new property `returnVersionStartDate` that other services could use instead.

Unfortunately, an error was introduced.

```javascript
  // app/services/return-versions/setup/submit-start-date.service.js

  // Reminder! Because of the unique qualities of Javascript, Year and Day are literal values, month is an index! So,
  // January is actually 0, February is 1 etc. This is why we deduct 1 from the month.
  session.returnVersionStartDate = new Date(`${payload['start-date-year']}-${payload['start-date-month'] - 1}-${payload['start-date-day']}`).toISOString().split('T')[0]
```

This way of creating the date was added. It uses a [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) to combine the values provided by the user and pass it as an argument o JavaScript's `new Date()`.

But it also applies a common fix for the fact JavaScript expects month to be provided as an index, not literally! So, January = 0, February = 1 etc.

The problem is `new Date('2024-04-01')` is not the same as `new Date(2024, 4, 1)`. The first _will_ return you 1 April 2024, the second will return 1 March 2024.

JavaScript understands that 'month' is literal when you provide a string argument to `new Date()`. It is only when passing month as a separate value that you have to treat it as an index.

The result for the return versions setup journey is that a user is entering, for example, `13 May 2019` on the start page, but we're storing (and would eventually persist!) `13 April 2019`.

This change fixes the issue.

Screenshot 2024-11-08 at 08 49 58

Screenshot 2024-11-08 at 08 50 10

### DIFF
--- a/app/services/return-versions/setup/submit-start-date.service.js
+++ b/app/services/return-versions/setup/submit-start-date.service.js
@@ -69,7 +69,7 @@ async function _save (session, payload) {
     // January is actually 0, February is 1 etc. This is why we deduct 1 from the month.
     session.returnVersionStartDate = new Date(`${payload['start-date-year']}-${payload['start-date-month'] - 1}-${payload['start-date-day']}`).toISOString().split('T')[0]
   } else {
-    session.returnVersionStartDate = new Date(session.licence.currentVersionStartDate).toISOString().split('T')[0]
+    session.returnVersionStartDate = new Date(session.licence.currentVersionStartDate)
   }
 
   return session.$update()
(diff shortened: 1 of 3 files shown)


jonathangoulding (approved):
I did notice this and fixed on the PR branch for the other implementation of the code. 

Should have fixed separately. 

apologize.

PR #1470, opened 2024-11-09
Fix issues on view monitoring station page

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4736

In our last release, we shipped a new version of the view monitoring station page, replacing the legacy one. However, we quickly received feedback that the dates we display for tagged licences don't look right.

We still have access to the legacy view, and when we compared what it showed, it confirmed that we'd got it wrong.

Having done some digging the problems are

- in the legacy view, if there is no status update date, it shows a blank. We instead have opted to display the `createdAt` date
- When you tag a licence, if it has a licence condition, you appear to be able to select it. The `licence_version_purpose_condition` is then linked to the `licence_monitoring_station` record. When this happens, the legacy page uses the abstraction dates from the linked `licence_version_purpose`. We don't, and at the time of writing, it is not clear what we do fall back to
- we are displaying deleted licence monitoring station records

When we dug deeper, we found more issues with the legacy and our first iteration. The legacy page has no ordering, and we opted for `created_at`. But that isn't clear on the page, so both look unordered.

The legacy and our new page first grouped the licence monitoring station records by licence and then added a row to the table for each licence. If a licence was linked to multiple licence monitoring station records, the details, such as the abstraction period, were added in a loop into each table cell. The problem is if the page is resized, those details no longer become readable across the row. It also causes issues for accessibility and our acceptance tests.

This change fixes the issues with the data being fetched and how it gets displayed on the page. It also fixes the problems with creating table rows (we use `rowspan` and add each licence monitoring station as a separate row to the table).

Finally, because of these changes, we could also do some refactoring. The fetch query handles all ordering (licence ref then threshold), so we can drop what the presenter was doing. We use more existing lib functions to reduce some of the presenter's complexity.


Cruikshanks:
I'm so sorry this comes across as a big change. A couple of files were renamed and refactored, so the diff is seeing them as a delete and an add, which is bumping the numbers.

I have to start with updating the view to correct the confusion with the column names. This is something I could split out. But those changes would then break stuff in the code, and I then go on to replace it anyway!

I have kept the changes to 'one thing per commit'. So, when you do come to review it, I'd recommend working through the commits first and then then taking a look at things as a whole.

PR #1482, opened 2024-11-14
Add quarterly returns submissions to return versions

jonathangoulding:

As part of the quarterly returns submissions work, we need to add a new checkbox to the additional-submission-options page for the Quarterly returns submissions.

When the return version start date is >= April 01 2025 then this is considered a quarterly return.

If the return version is for a water company and the return version start date is for a quarterly return. Then the quarterly return should be defaulted to true. This logic has been determined by the business.

### DIFF
--- a/app/lib/dates.lib.js
+++ b/app/lib/dates.lib.js
@@ -114,23 +114,18 @@ function _isLeapYear (year) {
 }
 
 /**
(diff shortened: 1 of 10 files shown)


Cruikshanks (approved):
I know I failed utterly to get back to you today. But I am so glad I didn't because you have...

PR #1484, opened 2024-11-14
Create Completed Return Logs page

StuAA78:
https://eaflood.atlassian.net/browse/WATER-4745

This PR introduces the View Return Logs page. The following changes are made to accommodate this:

- `unitNames` and `returnUnits` are added to `static-lookups.lib.js` for consistency;
- `ReturnSubmissionModel` has methods added `$applyReadings()`, `$meter()`, `$method()` and `$units()`;
- `BasePresenter` has `formatNumber()` which formats a number for display;
- `BasePresenter.titleCase()` is amended to first convert the string to lower-case, which was necessary for us to correctly convert text from `ALL CAPS` to `All Caps`.

Display of the page is gated behind the `enableSystemReturnsView` feature flag.


Cruikshanks:
**Prettified!**

Cruikshanks (approved):

PR #1489, opened 2024-11-15
Add quarterly returns submissions to check return versions

jonathangoulding:

As part of the quarterly returns submissions work, we need to add a new summary list item to the return version check page.

This change adds a new summary list item - Quarterly returns submissions. The logic involved in calculating whether to show hide the list item was done as part of the previous work to add this to the additional options page - https://github.com/DEFRA/water-abstraction-system/pull/1482


jonathangoulding:
I will add the persist in a seperate PR

PR #1490, opened 2024-11-15
Save quarterly returns submissions option

jonathangoulding:

As part of the quarterly returns submissions work, we need to persist the quarterly returns option that has been either defaulted or set by the user.

This change the quarterlyReturns key to the generated return version before it is persisted.

We will only persist the 'quarterlyReturns' if the return version start date is for quarterly return submissions period (>= 2025-04-01).

### DIFF
--- a/app/services/return-versions/setup/check/generate-return-version.service.js
+++ b/app/services/return-versions/setup/check/generate-return-version.service.js
@@ -8,6 +8,7 @@
 const GenerateReturnVersionRequirementsService = require('./generate-return-version-requirements.service.js')
 const ProcessExistingReturnVersionsService = require('./process-existing-return-versions.service.js')
 const ReturnVersionModel = require('../../../../models/return-version.model.js')
+const { isQuarterlyReturnSubmissions } = require('../../../../lib/dates.lib')
 
 /**
  * Uses the session data to generate the data sets required to create a new return version for a licence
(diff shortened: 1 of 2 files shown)


app/services/return-versions/setup/check/generate-return-version.service.js
```diff
@@ -49,17 +50,23 @@ async function _generateReturnRequirements (sessionData) {
...
+    quarterlyReturns = sessionData.quarterlyReturns
+  }
+
   return {
     createdBy: userId,
     endDate,
     licenceId: sessionData.licence.id,
     multipleUpload: sessionData.multipleUpload,
     notes: sessionData?.note?.content,
+    quarterlyReturns,
```
jonathangoulding, line 69: I do think this would be better as a ternary but i am not sure we do this elsewhere. 

Reason for the false is that quarterlyReturns can be undefined when it has not been set. 

Unlike multiple-upload which is set when the session is instantiated.

```js
  quarterlyReturns: isQuarterlyReturnSubmissions(sessionData.returnVersionStartDate) ? sessionData.quarterlyReturns : false
```

Cruikshanks: I think in this case a ternary is fine, and I have no preference either way so happy if you want to switch to use it.

Cruikshanks (approved):
I'm approving as I am happy with the change in its current form. But if you would rather use the ternary, make the change and give me a ping and I'll happily re-approve.

PR #1491, opened 2024-11-15
View quarterly returns submissions option

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4730

As part of the quarterly returns submissions work, we need to show the return version is flagged for quarterly returns.

We will only show the 'quarterly return submissions' if the return version start date is for quarterly return submissions period (>= 2025-04-01).

When this is the case we show Yes or no depending on the saved state of the quarterly return.

Previous work to implement the save / persist can be found here https://github.com/DEFRA/water-abstraction-system/pull/1490


jonathangoulding:
Best waiting for - https://github.com/DEFRA/water-abstraction-system/pull/1490 to be merged before this one.

Cruikshanks (approved):

PR #1492, opened 2024-11-15
Fix outstanding JSDoc issues

Cruikshanks:
In [Lint JSDOC's](https://github.com/DEFRA/water-abstraction-system/pull/1269), we added linting of our JSDoc comments. If we are going to do it, we should do it correctly and consistently!

There were quite a few, so since then, we have cleaned up certain areas of the codebase while configuring the linter to only warn us of issues.

With this final change, we deal with the remaining issues so we can switch the linter to error when it encounters an issue.


app/validators/bill-runs/setup/region.validator.js
```diff
@@ -10,7 +10,8 @@ const Joi = require('joi')
 /**
  * Validates data submitted for the `/bill-runs/setup/{sessionId}/region` page
  *
- * @param {object} payload - The payload from the request to be validated
+ * @param {object} data - The payload from the request to be validated
```
jonathangoulding, line 13: Should these be updated to say 'The data' etc. 

i see all others are the same so resolve if a none issue.

jonathangoulding (dismissed):

app/validators/bill-runs/setup/region.validator.js
Cruikshanks: Fair comment. TL;DR; meh 🤷 

Regarding all the validators, it's just data that needs to be validated. Occasionally, a validator checks something that _didn't_ come via a payload.

So we were going for something 'generic'. But, in most cases, it is coming from a payload, and there are other JSDoc comments where we call out where we expect the param to be coming from, for example, the result of a fetch service. Hence, the mention in the comment.

I'm happy for others to wade in before merging.

jonathangoulding: Cool. 

I am not precious. 

I will resolve can always adjust in the future.

robertparkinson (dismissed):

Jozzey (approved):

PR #1495, opened 2024-11-15
Add abstraction method to licence purpose page

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4705

During the process of migrating the [licence points page](https://github.com/DEFRA/water-abstraction-system/pull/1451), the original plan was to include the abstraction method for each licence point. However, after further consideration, it was decided to refactor and include the abstraction method within the licence purpose page instead.

This PR adds the abstraction method to each purpose on the licence purpose page.


app/models/licence-version-purpose.model.js
```diff
@@ -32,6 +32,14 @@ class LicenceVersionPurposeModel extends BaseModel {
           to: 'licenceVersionPurposeConditions.licenceVersionPurposeId'
         }
       },
+      licenceVersionPurposePoints: {
```
Jozzey, line 35: Need to add a unit test for this new relationship

Jozzey (changes requested):

Jozzey (approved):

PR #1496, opened 2024-11-18
Fix mark for supp billing page attributes

Beckyrose200:
https://github.com/DEFRA/water-abstraction-acceptance-tests/pull/144

As part of the work for the acceptance tests for supplementary billing flags, this PR introduces the necessary data-test attributes on the view page for the "Mark for Supplementary Billing" feature.

### DIFF
--- a/app/controllers/licences.controller.js
+++ b/app/controllers/licences.controller.js
@@ -28,7 +28,7 @@ async function markedForSupplementaryBilling (request, h) {
   const pageData = await MarkedForSupplementaryBillingService.go(licenceId)
 
   return h.view('licences/marked-for-supplementary-billing.njk', {
-    pageTitle: 'You’ve marked this licence for the next supplementary bill run',
+    pageTitle: "You've marked this licence for the next supplementary bill run",
     activeNavBar: 'search',
     ...pageData
   })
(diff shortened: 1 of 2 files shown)


test/controllers/licences.controller.test.js
```diff
@@ -538,7 +538,7 @@ describe('Licences controller', () => {
           const response = await server.inject(options)
 
           expect(response.statusCode).to.equal(200)
-          expect(response.payload).to.contain('You’ve marked this licence for the next supplementary bill run')
+          expect(response.payload).to.contain('You&#39;ve marked this licence for the next supplementary bill run')
```
Jozzey, line 541:
```suggestion
          expect(response.payload).to.contain("You've marked this licence for the next supplementary bill run")
```

Jozzey (changes requested):
Just the one suggestion

test/controllers/licences.controller.test.js
Beckyrose200: image
Doing it that way doesn't work because it reads it from the payload and in the payload it shows as 'You&#39;ve'. 
There are other controllers that have had the same issue where they have used 'You&#39;ve' instead of 'You've'

image

Jozzey (approved):

PR #1498, opened 2024-11-19
Create `.git-blame-ignore-revs` file

StuAA78:
[Our move away from Standard](https://github.com/DEFRA/water-abstraction-system/pull/1476) has involved a huge number of changes. In order to preserve the `git blame` record, we create a `.git-blame-ignore-revs` file, which we can use to ignore specific commits. We previously introduced it to our other repos when implementing Standard:

https://github.com/DEFRA/water-abstraction-team/issues/37
https://github.com/DEFRA/water-abstraction-service/pull/1746

This PR creates the file for this repo with the required commit to ignore the large-scale linting changes.

Note that while GitHub supports this file, it's up to us to configure our local repo to make use of it. This can be done with:
```bash
git config blame.ignorerevsfile .git-blame-ignore-revs
```


Cruikshanks:
**Prettified!**

Cruikshanks (approved):

PR #1499, opened 2024-11-19
Stop returning command-only async calls

Cruikshanks:
There are two scenarios where folks will see @Cruikshanks sweep in and ask for a change.

```javascript
// Scenario 1
async function go (someId) {
  const result = await _fetch(someId)

  return result
}

// Scenario 2
async function go (someId) {
  await _updateStatus(someId)
}
```

In both cases, he'll ask folks to return the call immediately.

```javascript
// Scenario 1
async function go (someId) {
  return _fetch(someId)
}

// Scenario 2
async function go (someId) {
  return _updateStatus(someId)
}
```

In a recent change where we added a job, we made a command-only async call as a batch to `Promise.all()`. The original change had `await ProcessLicence.go(licenceId)`, so in swooped @Cruikshanks !

This was challenged, so we carried out a mini spike, did some Googling, and came to the conclusion.

- We have it bang on for scenario 1; we should return immediately
- @Cruikshanks has it _so wrong_ for scenario 2 🤦

The spike concluded that, technically, there is no difference. Using `await` rather than `return` doesn't block anything, and the resolution is the same.

Googling backed up the challenge made: declaring `return _doThing()` makes it look like we intend to return something when we don't. When doing this, we have to add the following to the JSDocs, which should have clued us in earlier.

> `@returns {Promise<object>} the promise returned is not intended to resolve to any particular value`

We found a blog post that put into 'bigger' words (!!) why this is bad.

> `await` should be used for commands to ensure void is returned. The reason why void is returned for command is to enforce [command-query separation](https://en.wikipedia.org/wiki/Command%E2%80%93query_separation). Do not mix updates with reads as that violates [single-responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle).
>
> [When to use return vs await vs return await vs nothing](https://blog.battlefy.com/when-to-use-return-vs-await-vs-return-await-vs-nothing)

So, this change is @Cruikshanks putting all the instances where he got it wrong right!


app/services/bill-runs/submit-cancel-bill-run.service.js
```diff
@@ -48,15 +46,7 @@ async function go (billRunId) {
...
-
-  // NOTE: We originally believed that should anything error in _cancelBillRun() the try/catch in the controller would
-  // catch it. However, when testing this theory we found it crashed the test framework. So, we tried it for real and
-  // again confirmed we'd crash the service. The controller would never see the error. It is because we are not awaiting
-  // this call that any errors thrown are considered uncaught. However, if we instead use the ES6 variant the error _is_
-  // caught. All we can do at this point is log it.
-  _cancelBillRun(billRun).catch((error) => {
-    global.GlobalNotifier.omfg(error.message, { billRunId }, error)
-  })
+  await _cancelBillRun(billRun)
```
jonathangoulding, line 49: Just want to check why the error handling has been removed. 

As i read it we no longer require this specific error message to be used. There for the error will just bubble up.

jonathangoulding (dismissed):

app/services/bill-runs/submit-cancel-bill-run.service.js
Cruikshanks: Because it is already handled in the controller that uses this service, and the intent was always to get the error to bubble up to there (it wasn't, hence the old-style promise handling).

However, I've spotted something I've broken (related to this 😁 ), so taking it back to update. Thanks for looking through it though.

jonathangoulding: Resolve when you are happy. I can reapprove if needed.

PR #1501, opened 2024-11-20
Refactor cancel bill run handling

Cruikshanks:
Recently, we realised that sometimes we were returning async calls when in fact we should have been `await` them.

When we applied [Stop returning command-only async calls](https://github.com/DEFRA/water-abstraction-system/pull/1499), we intended to change the 'cancel bill run' and 'send bill run' processes to replace the following

```javascript
  _cancelBillRun(billRun).catch((error) => {
    global.GlobalNotifier.omfg(error.message, { billRunId }, error)
  })
```

with this

```javascript
  await _cancelBillRun(billRun)
```

Much cleaner! Plus, we could eliminate the fake delays in the unit tests. Bonus!

But when double-checking things manually, we spotted we'd broken these processes. The intent was that

- The user is made to wait until the bill run status is updated (cancelling/sending)
- Then the request is returned so we don't block the UI (user is redirected to `/bill-runs`)
- The app continues cancelling the bill run in the background

We needed the old logic to support the controller awaiting the call to  `SubmitCancelBillRunService` for the status change but then letting it return control whilst cancelling the bill run in the background.

So, this was never related to incorrectly using `return myAsyncCall()`.

The most straightforward fix is to put back the changes. But the previous code felt like a hack, and the delays in the unit tests were 'icky'. In hindsight, these fudges tell us we're asking the `SubmitCancelBillRunService` to do too much. If we split its responsibilities, we could get it working again without these fudges.

This change refactors the cancel bill run process ahead of PR #1499, so we can pretend we never broke it!


Jozzey (approved):

app/services/bill-runs/cancel/delete-bill-run.service.js
```diff
@@ -0,0 +1,229 @@
...
+const BillRunVolumeModel = require('../../../models/bill-run-volume.model.js')
+const { db } = require('../../../../db/db.js')
+const ChargingModuleDeleteBillRunRequest = require('../../../requests/charging-module/delete-bill-run.request.js')
+const { calculateAndLogTimeTaken } = require('../../../lib/general.lib.js')
+const ReviewChargeVersionModel = require('../../../models/review-charge-version.model.js')
+const ReviewLicenceModel = require('../../../models/review-licence.model.js')
+const ReviewReturnModel = require('../../../models/review-return.model.js')
+
+/**
+ * Deletes a bill run, all its associated records and the its match in the Charging Module API
```
robertparkinson, line 21: you have an extra 'the' in here by the looks of it?

app/services/bill-runs/cancel/delete-bill-run.service.js
```diff
@@ -0,0 +1,229 @@
...
+    ])
+
+    _logResult(startTime, billRun, results)
+  } catch (error) {
+    global.GlobalNotifier.omfg('Delete bill run failed', billRun, error)
+  }
+}
+
+/**
+ * These deals with all the tables that may be populated for a bill run. Because we have to deal with bills created
```
robertparkinson, line 65: this should start with 'this' I think?

robertparkinson (approved):

app/services/bill-runs/cancel/delete-bill-run.service.js
Cruikshanks: 💯 

But I'm about to do exactly the same refactor for sending a bill run. So, I promise 😇 I'll fix it in that PR rather than wasting you lovely approval!

app/services/bill-runs/cancel/delete-bill-run.service.js
Cruikshanks: 💯 

But I'm about to do exactly the same refactor for sending a bill run. So, I promise 😇 I'll fix it in that PR rather than wasting you lovely approval!

PR #1506, opened 2024-11-21
Add determine licence flags service

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4746

This PR is contributing to the cleaning up of the supplementary billing flags process. This PR specifically focuses on the licence flags service which is legacy code we have migrated to use the new communal supplementary billing flags service. It will flag a licence based on the provided scheme. The flagging mechanisms that will use this service are:
* A new licence agreement
* An ended licence agreement
* A deleted licence agreement
* Removing a licence from a pre-sroc two-part tariff bill run.

As this is handling legacy flagging this service only cares about the pre-sroc and sroc flags. Two-part tariff years flagging is not included for any of these so we default it to false.


Cruikshanks:
**Prettified**!

app/services/licences/supplementary/determine-licence-flags.service.js
```diff
@@ -0,0 +1,63 @@
...
+  if (scheme === 'alcs') {
+    flagForPreSrocSupplementary = true
+  } else {
+    flagForSrocSupplementary = true
+  }
+
+  const result = {
+    licenceId,
+    regionId: licence.regionId,
+    startDate: null,
```
Jozzey, line 45: I'd remove the `startDate` & `endDate` rather than assign `null` to them. This is because when `flagForTwoPartTariffSupplementary` is set to false, these 2 values are not used in `app/services/licences/supplementary/process-billing-flag.service.js`

test/services/licences/supplementary/determine-licence-flags.service.test.js
```diff
@@ -0,0 +1,117 @@
...
+          const result = await DetermineLicenceFlagsService.go(licence.id, scheme)
+
+          expect(result.flagForPreSrocSupplementary).to.equal(false)
+          expect(result.flagForSrocSupplementary).to.equal(true)
+          expect(result.flagForTwoPartTariffSupplementary).to.equal(false)
+        })
+      })
+
+      describe("and a charging scheme of 'alcs'", () => {
+        it('returns the licence details', async () => {
```
Jozzey, line 98: I don't think this test is right as the scheme doesn't appear to have been set to "alcs" in a`before` and the flagging in the `expect` looks wrong.

test/services/licences/supplementary/process-billing-flag.service.test.js
```diff
@@ -275,6 +276,86 @@ describe('Process Billing Flag Service', () => {
...
     })
+
+    describe('with a licence Id', () => {
+      before(() => {
+        payload = {
+          licenceId: 'b5f81330-bec5-4c3e-95dd-267c10836fea'
+        }
+      })
+
+      describe('that should be flagged for two-part tariff supplementary billing', () => {
```
Jozzey, line 287: I'm not really getting this test. The `DetermineLicenceFlagsService` always sets `flagForTwoPartTariffSupplementary` to false and never returns a start date. So no years could be calculated for 2PT?

Jozzey (changes requested):

test/services/licences/supplementary/determine-licence-flags.service.test.js
Beckyrose200: Good spot

test/services/licences/supplementary/process-billing-flag.service.test.js
Beckyrose200: Another good spot, my bad! Thank you

Jozzey (approved):

PR #1522, opened 2024-11-27
Add a new page for notifications returns period

jonathangoulding:

As part of ongoing work to migrate the legacy code we are rebuilding the notifications journeys.

This change adds a new journey for notifications. This accessed from the legacy /manage page and is behind a feature toggle.

The initial notification routes and controller scaffolding has been put in place to build the journey.

The returns period page has been scaffolded to facilitate further development of the page and to get an idea of language used, established patterns and code structure.

This change will give us a good foundation for the notifications journey.

The service logic and submit route will be added in a later commit.


app/routes/notifications-setup.routes.js
```diff
@@ -0,0 +1,22 @@
...
+
+const routes = [
+  {
+    method: 'GET',
+    path: basePath + 'returns-period',
+    options: {
+      handler: NotificationsSetupController.viewReturnsPeriod,
+      auth: {
+        access: {
+          scope: ['billing']
```
jonathangoulding, line 15: Need to work on the scope / notifications scope of the UI.

app/routes/notifications-setup.routes.js
```diff
@@ -0,0 +1,22 @@
...
+
+const routes = [
+  {
+    method: 'GET',
+    path: basePath + 'returns-period',
+    options: {
+      handler: NotificationsSetupController.viewReturnsPeriod,
+      auth: {
+        access: {
+          scope: ['returns']
```
jonathangoulding, line 15: I think this is the only auth access needed for this page. There are some notification scopes that could be added ?

jonathangoulding:
Links to - https://github.com/DEFRA/water-abstraction-ui/pull/2644/files

app/routes/notifications-setup.routes.js
Cruikshanks: I'd have to check the scopes used on the legacy pages. But this is fine whilst we scaffold things out.

test/controllers/notifications.controller.test.js
```diff
@@ -0,0 +1,78 @@
...
+    // Create server before each test
+    server = await init()
+
+    // We silence any calls to server.logger.error made in the plugin to try and keep the test output as clean as
+    // possible
+    Sinon.stub(server.logger, 'error')
+
+    // We silence sending a notification to our Errbit instance using Airbrake
+    Sinon.stub(server.app.airbrake, 'notify').resolvesThis()
+  })
```
Cruikshanks, line 33: Small NIT. Moving the `await init()` call into a `before()` block to avoid `MaxListenersExceededWarning` as I've done in [Fix MaxListenersExceededWarning in tests](https://github.com/DEFRA/water-abstraction-system/pull/1519).

```suggestion
  // Create server before running the tests
  before(async () => {
    server = await init()
  })

  beforeEach(async () => {
    // We silence any calls to server.logger.error made in the plugin to try and keep the test output as clean as
    // possible
    Sinon.stub(server.logger, 'error')

    // We silence sending a notification to our Errbit instance using Airbrake
    Sinon.stub(server.app.airbrake, 'notify').resolvesThis()
  })
```

Note - You'll need to add `before` into the require() at the top after accepting/doing this suggestion.

PR #1525, opened 2024-11-28
Al's take on Return log for ended licence

Cruikshanks:

This will be an integral and complex part of the service, but until now, I have only a tertiary understanding of how it is being assembled.

In [Return log for ended licence](https://github.com/DEFRA/water-abstraction-system/pull/1425), we're making some fundamental changes, and my feedback is being sought, so it is time to get my hands dirty.

There is likely to be a lot of feedback to provide, and when this is the case, I find it easier to do so in a separate PR using my commits to highlight and explain my changes. This allows me to check that what I'm suggesting works!


Cruikshanks:
@robertparkinson I'm done!

I've tested the job and the `/check` endpoint locally, and everything works as expected.

So, I'm happy that we go with this if you are. If so, approve this PR, and I'll merge it into yours, then I'll approve yours, and we can merge it into `main`.

Even more exciting, I made this one-line change to `app/services/return-versions/setup/check/persist-return-version.service.js` and believe I have it working for all scenarios where someone adds a new return version via licence setup 😁🤞

Screenshot 2024-12-06 at 18 38 33

robertparkinson (approved):

PR #1527, opened 2024-11-28
Notifications returns period options - January

jonathangoulding:

This change adds the notifications returns periods option for between january.

The other options will be added in following commits.

The submit logic will be developed in a separate commit so the 'values' used to capture the selected options may change.

### DIFF
--- a/app/presenters/notifications/setup/returns-period.presenter.js
+++ b/app/presenters/notifications/setup/returns-period.presenter.js
@@ -9,6 +9,7 @@ const { monthsAsIntegers } = require('../../../lib/static-lookups.lib')
 
 const currentPeriod = 'currentPeriod'
 const nextPeriod = 'nextPeriod'
+const twentyEighth = 28
 
 /**
  * Formats data for the `/notifications/setup/returns-period` page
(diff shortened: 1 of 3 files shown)


app/presenters/notifications/setup/returns-period.presenter.js
```diff
@@ -18,9 +18,44 @@ function go() {
...
+
   return []
 }
 
+/*
+ *  Checks if a date is in January
+ *  A date is in January if it is between 1st January - 28th January
+ */
+function _dayIsInJanuary(date) {
+  return date.getMonth() === 0 && date.getDate() >= 1 && date.getDate() <= 28
```
Cruikshanks, line 37: Really, _really_ minor NIT. But do we need to check `date.getDate() >= 1` _and_ `date.getDate() <= 28`. Wouldn't that last check suffice?

Cruikshanks (changes requested):
I just might not be seeing the possible scenario 🤷

app/presenters/notifications/setup/returns-period.presenter.js
jonathangoulding: Good point. can remove.

Cruikshanks (dismissed):

Cruikshanks (approved):
I'm not approving it again. Take it or leave it! 😁

PR #1532, opened 2024-12-02
Migrate abstraction conditions view page

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4825

As the legacy licence view page has now been rebuilt, the next process is to update some of the links pointing to legacy pages. One of these legacy pages is the abstraction conditions view. If we are able to move this over to our system, it will help towards removing the use of old legacy code.

This PR is focused on rebuilding the abstraction conditions view page from the legacy code base.


test/presenters/licences/view-licence-conditions.presenter.test.js
```diff
@@ -0,0 +1,305 @@
...
+                }
+              ]
+            }
+          }
+        ]
+      }
+    ],
+    licence
+  }
+}
```
Cruikshanks, line 305: I believe we use this test code block in `test/services/licences/view-licence-conditions.service.test.js` as well.

This pattern came after you started recreating these legacy pages. But could I suggest we start a `licences.fixture.js` in `test/fixtures` and move this block there? We can then just reference the fixture from both these tests.

Cruikshanks (changes requested):
Just the one suggested change! 💪🙇

Cruikshanks (approved):

PR #1545, opened 2024-12-10
Update model tests to use before block

Beckyrose200:
DEFRA/water-abstraction-team#128

This PR is part of our ongoing work to make unit tests faster, more efficient and less flakey, focusing on the model unit tests.

We are simplifying the database seeding process by moving it to a top-level before block instead of using beforeEach. Since the model unit tests check the relationships between table models, putting all the test data in one place means we only need to seed the database once instead of before each test. By seeding less data into the test database, our hope is that the tests run faster, the code is cleaner, and they are less flakey due to conflicting test data.

### DIFF
--- a/test/models/billing-account-address.model.test.js
+++ b/test/models/billing-account-address.model.test.js
@@ -4,7 +4,7 @@
 const Lab = require('@hapi/lab')
 const Code = require('@hapi/code')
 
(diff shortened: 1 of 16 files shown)


Cruikshanks:
Superseded by https://github.com/DEFRA/water-abstraction-system/pull/2478

PR #1549, opened 2024-12-10
Display multiple purposes in 2PT review pages

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4712

During 2PT UAT, the business has requested that we display all the purposes attached to a return or charge element rather than just one.

This PR will update the two-part tariff review pages to display the purposes as a comma-separated list if there are more than 1.


Jozzey:
The SonarCloud failure is down to this issue which is being worked on. The new code should be 100% unit tested:
`This action is deprecated and will be removed in a future release. Please use the sonarqube-scan-action action instead. The sonarqube-scan-action is a drop-in replacement for this action.`

app/presenters/bill-runs/review/base-review.presenter.js
```diff
@@ -196,6 +196,32 @@ function formatIssues(issues) {
...
+ *
+ * It is possible that a return log has more than one purpose associated it. If this is the case we need the description
+ * of each purpose from the array of purposes to be concatenated into a single comma separated string ready to present.
+ *
+ * @param {Array} purposes - the purposes taken from the return logs metadata
+ *
+ * @returns {string} the purpose descriptions presented as a comma separated string
+ */
+function formatPurposes(purposes) {
+  const firstPurposeDescription = purposes[0].tertiary.description
```
Beckyrose200, line 210: Could this not be done by using array.join(', ')
image

app/presenters/bill-runs/review/review-licence.presenter.js
```diff
@@ -180,7 +181,7 @@ function _formatReviewReturns(reviewReturns) {
       abstractionPeriod: formatAbstractionPeriod(periodStartDay, periodStartMonth, periodEndDay, periodEndMonth),
       description,
       issues: formatIssues(issues),
-      purpose: purposes[0].tertiary.description,
+      purpose: formatPurposes(purposes),
```
Beckyrose200, line 184: If you do array.join() you wouldn't need that in the base presenter so it could just be put into there

Beckyrose200 (commented):
Small comment :)

app/presenters/bill-runs/review/base-review.presenter.js
Beckyrose200: `function purposes(purposes) {
  return purposes.map((purpose) => {
    return purpose.tertiary.description.join(', ')
  })
}`

Jozzey: I cannot perform a `join` on `purpose.tertiary.description` as it's not an array

Jozzey: I've modified your idea a bit to get it to work with my array, cheers.

app/presenters/bill-runs/review/review-licence.presenter.js
Jozzey: It wouldn't work on this as `purposes` is an array of objects, and the string I want is buried in the object

Beckyrose200 (approved):

PR #1555, opened 2024-12-11
Add ad-hoc returns enter licence page

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4816

As part of the work for sending ad-hoc returns, this PR focuses on adding the enter a licence page to the journey.


app/services/notifications/ad-hoc-returns/submit-licence.service.js
```diff
@@ -0,0 +1,86 @@
...
+  const dueReturns = await ReturnLogModel.query().where('licenceRef', licenceRef).where('status', 'due').first()
+
+  return !!dueReturns
+}
+
+async function _fetchLicence(licenceRef) {
+  const licence = await LicenceModel.query().where('licenceRef', licenceRef).first()
+
+  return !!licence
+}
```
StuAA78, line 62: Just a small thing but I'd be expecting a function called "fetch..." to fetch _and return_ something. These are more like `_dueReturnsExist` and `_licenceExists` or something like that.

StuAA78 (changes requested):
Looks good, just a couple of tiny tweaks for clarity requested 🙂

app/services/notifications/ad-hoc-returns/initiate-session.service.js
```diff
@@ -0,0 +1,31 @@
+'use strict'
```
Jozzey, line 1: Was there a reason why we couldn't use `app/services/bill-runs/setup/initiate-session.service.js` to do this if we moved it to a more generic folder?

StuAA78 (approved):

PR #1559, opened 2024-12-12
Feature returns period lib determine upcoming periods

jonathangoulding:

This change introduces the logic determine the current and next return period. This is based on the due date.


app/lib/return-periods.lib.js
```diff
@@ -270,6 +270,82 @@ function _newYearElapsedQuarterFourDueDate(determinationDate, period) {
...
+
+/**
+ * Determines the next return period based on a given determination date. It calculates the return periods,
+ * maps them, sorts them by due date, and returns the next return period.
+ *
+ * @param {Date} [determinationDate=new Date()] - The date to determine the next return period from. Defaults to the current date if not provided.
+ *
+ * @returns {object} The next return period after the given determination date, which includes properties like `dueDate`, `startDate`, and `endDate`.
+ */
+function determineNextReturnPeriod(determinationDate = new Date()) {
```
robertparkinson, line 296: why do you need two functions? can't you just have one function that returns both the current and the next and then do what you need with it?

robertparkinson (changes requested):

app/lib/return-periods.lib.js
jonathangoulding: I was trying to make it more explicit. 

If we wanted to get the previous return period then it would follow this pattern. 

A single function, i think, should return all the sorted periods. But we do not need them all.

jonathangoulding: Maybe there is a case for a single function that returns them all and it's up to the caller to use the data for it's own use case.

jonathangoulding: `determineUpcomingReturnPeriods` ?

robertparkinson (approved):

PR #1564, opened 2024-12-13
reissue return logs when new return version saved

robertparkinson:

As part of our work to replace NALD in generation of return logs we need to be able to reissue return logs when a new return version is created. 

This PR updates the logic for return return requirements to return them from both return versions when there is a changed made. This then enables the voiding and creation of return logs both before and after the new dates that are generated as part of the new return version.


app/services/return-versions/setup/check/submit-check.service.js
```diff
@@ -29,6 +30,12 @@ async function go(sessionId, userId) {
 
   await PersistReturnVersionService.go(returnVersionData)
 
+  const dateOffset = 24 * 60 * 60 * 1000
```
jonathangoulding, line 33: If this is whole day as an int it could be worded better or a comment added ? 

`oneDayInMilliseconds` - if i have understood it right.

jonathangoulding (approved):

PR #1575, opened 2024-12-17
Feature notifications setup returns period session

jonathangoulding:

This change introduces session save and retrieval for the returns period page. When he user has selected a returns period then this will be saved to the session for the notifications setup journey.


StuAA78 (approved):
LGTM 👍

PR #1579, opened 2024-12-20
Highlight code with no tests but reported in coverage

jonathangoulding:
As part of ongoing improvements we are trying to make sure all code has coverage (where deemed reasonable).

Some of our code may be tested indirectly and this helps cover parts of the codebase.

In this change we are highlighting that the:

- app server does not have/ or need a test
- check endpoints are for testing and is no production code so this is deemed to not need a test
- app plugins are inherently tested when testing with the server running with hapi lab

We can look to test these specifically at a later date.

### DIFF
--- a/.labrc.js
+++ b/.labrc.js
@@ -3,7 +3,17 @@
 module.exports = {
   verbose: true,
   coverage: true,
-  'coverage-exclude': ['db/seeds'],
+  'coverage-exclude': [
+    'db/seeds',
+    'app/controllers/check.controller.js', // Test helper endpoints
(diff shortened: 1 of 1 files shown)


Cruikshanks (approved):
Works for me!

Also, great use of _"Waiting for something else"_ time. 💪⭐

Cruikshanks (changes requested):
Works for me!

Also, great use of _"Waiting for something else"_ time. 💪⭐  

But just realised we need to make this change in the `.labrc.js` file as well to keep the two outputs in sync.

Cruikshanks (approved):
Oh just ignore me. First day back from holiday!

PR #1589, opened 2024-12-30
Replace Hapi Lab with node-test for test runner

Cruikshanks:
https://github.com/DEFRA/water-abstraction-team/issues/130

We as a team would love to move the [water-abstraction-system](https://github.com/DEFRA/water-abstraction-system) project to use [ECMAScript Modules (ESM)](https://nodejs.org/api/esm.html). The one thing that has held us back is our test framework [Hapi Lab](https://hapi.dev/module/lab/).

We've worked hard over the last year restructuring how we build our tests and [moving away from our dependence on the database](https://github.com/DEFRA/water-abstraction-team/issues/124).

This means we are now ready to make the switch to a test framework that natively supports ESM. We've opted for [node-test](https://nodejs.org/api/test.html#test-runner).

In this change, we

- remove the **Hapi Lab** dependency
- update all tests to require **node-test**
- update our scripts to use **node-test**


Cruikshanks:
Superseded by #3489

PR #1591, opened 2024-12-31
Exclude rtn vers. not started in reissue rtn logs

robertparkinson:

As part of the testing of the historic changes, it was noticed that additional return logs were being created that had start dates after their end dates.

After some digging, we figured out that when dealing with a historic change it was creating return logs for the return versions that had not started yet.

This change excludes return requirements from a return version, which have a version start date after the return cycle end date.

### DIFF
--- a/test/fixtures/return-logs.fixture.js
+++ b/test/fixtures/return-logs.fixture.js
@@ -41,6 +41,28 @@ function returnCycles() {
   ]
 }
 
+function returnCyclesAcrossReturnVersions() {
+  return [
+    ...returnCycles(),
+    {
+      id: '4c5ff4dc-dfe0-4693-9cb5-acdebf6f76b5',
+      startDate: new Date('2023-11-01'),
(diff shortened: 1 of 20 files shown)


test/fixtures/return-logs.fixture.js
```diff
@@ -18,10 +18,11 @@ function returnCycle(summer = false) {
 /**
  * Represents a list of return cycles from the database, as fetched in `ProcessLicenceReturnLogsService`
  *
+ * @param {number} [numberOfCycles=2] - the number of return cycles to return - defautls to the first two
```
Cruikshanks, line 21: NIT

```suggestion
 * @param {number} [numberOfCycles=2] - the number of return cycles to return - defaults to the first two

```

Cruikshanks (changes requested):
Just a minor NIT

Cruikshanks (approved):

PR #1602, opened 2025-01-07
Return Logs setup received date

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4844

As part of the work to migrate the return logs setup journey, this PR is focused on adding the 'When was this return received' page.


StuAA78 (changes requested):
Looks good -- just a couple of small tweaks requested 🙂

StuAA78 (approved):

PR #1603, opened 2025-01-08
Create quarterly return logs

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4841

As part of our work to replace NALD for the generation of return logs we're looking at introducing the concept of a quarterly return. This is an all year cycle that has been split into 4 3 month periods that water companies will return against through out the year.

This PR checks if the return version has been marked for quarterly returns, if it starts after the go live date and then creates the smaller return logs. 

There is also some refactoring of the quarter period naming.


app/services/jobs/return-logs/create-quarterly-return-cycles.service.js
Cruikshanks, line 1: Just a NIT.

I try and reserve `create` for when we are persisting something to the DB. So, how about `Determine` instead?

Then, as we are dealing with periods and not cycles, could we amend that as well?

So, my suggestion would be to rename the service `DetermineQuarterlyReturnPeriods`.

app/services/jobs/return-logs/create-quarterly-return-cycles.service.js
```diff
@@ -0,0 +1,47 @@
...
+ * Given a return cycle, generate each quarterly return period
+ *
+ * @param {object} returnCycle - the return cycle to break into quarterly period
+ *
+ * @returns {Array<object>} the created quarterly return periods
+ */
+function go(returnCycle) {
+  if (!returnCycle) {
+    return []
+  }
```
Cruikshanks, line 20: Is there ever a scenario we would not pass it a return cycle, and then happily deal with an empty result?

If we never pass it an empty cycle, we shouldn't have the guard clause to handle the scenario.

app/services/jobs/return-logs/create-quarterly-return-cycles.service.js
```diff
@@ -0,0 +1,47 @@
...
+
+  // To reuse as much code as possible we use two dates to determine the quarterly return periods
+  // The first one uses the start date of the return cycle, and the second uses the end date
+  // This enables us to get the correct year for quarter four while reusing exisitng code
+  const quarterlyReturnPeriods = determineReturnsPeriods(returnCycle.startDate)
+  const quarterFourReturnPeriod = determineReturnsPeriods(returnCycle.endDate)
```
Cruikshanks, line 26: I wonder, if we moved this function to `app/lib/return-periods.lib.js` would that actually make it even easier/clearer?

Also, it means you can ignore my file change suggestion!

Cruikshanks (changes requested):
Before you make any changes, I think I might need a chat to get me back up to speed again about the periods/cycles stuff in `app/lib/return-periods.lib.js`.

Because I'm not 💯 on that bit, so I'd just be hand-waving it through.

Cruikshanks (approved):

PR #1604, opened 2025-01-09
Add notification review initial page

jonathangoulding:

Add the initial page, controller and router for the notification review page.

The presenter and service will be added in a later change as the service will have a complex database query and we will need to format the data for the view.


jonathangoulding:
This is added to easily discuss the route `review` (if we are happy with that path name) and to give us the basic structure before adding complexity.

app/views/notifications/setup/view-review.njk
Cruikshanks, line 1: Could we rename this to just `review.njk`?

test/controllers/notifications-setup.controller.test.js
```diff
@@ -138,6 +139,37 @@ describe('Notifications Setup controller', () => {
...
+        getOptions = {
+          method: 'GET',
+          url: basePath + `/${session.id}/review`,
+          auth: {
+            strategy: 'session',
+            credentials: { scope: ['returns'] }
+          }
+        }
+      })
+      describe('when a request is valid', () => {
```
Cruikshanks, line 155: NIT
```suggestion
      })

      describe('when a request is valid', () => {
```

Cruikshanks (changes requested):
Just a couple of NITS.

Other than that, makes perfect sense!

Cruikshanks (approved):

PR #1606, opened 2025-01-10
Add guidance view on entering multiple readings

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4812

For some returns, users are required to submit readings for each day, which can be very tedious to enter manually. To simplify this, a previous page for entering multiple readings is being developed - [Enter multiple abstraction values](https://eaflood.atlassian.net/browse/WATER-4811). Since the process requires the readings to be submitted in a specific format (either comma-separated or line-separated), some users who are unfamiliar with the process may need guidance. This will be provided through the new guidance page, which explains how to submit multiple returns on the `Enter multiple daily volumes` page.

This PR focuses on creating the static guidance page for entering multiple abstraction values.


app/views/return-logs/setup/guidance.njk
Cruikshanks, line 1: I'd be curious if we could get the same result using our standard implementation of `{% extends 'layout.njk' %}`. It seems to work for `app/views/info.njk`.

We then also wouldn't need to worry about `{% block pageTitle %}`, `{% block head %}` and `{% block header %}`.

Screenshot 2025-01-21 at 21 38 14

Cruikshanks (commented):
More one to try rather than a request for changes (unless it works 😉 !)

app/views/return-logs/setup/guidance.njk
Cruikshanks: Give me a ping 😁

Cruikshanks (dismissed):

Cruikshanks (approved):

PR #1615, opened 2025-01-14
Return Logs setup - Meter provided

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4846

As part of the work to migrate the return logs setup journey, this PR is focused on adding the 'Have meter details been provided' page.


app/controllers/return-logs-setup.controller.js
```diff
@@ -43,6 +52,25 @@ async function start(request, h) {
   return h.view('return-logs/setup/start.njk', pageData)
 }
 
+async function submitMeterProvided(request, h) {
+  const {
+    params: { sessionId },
+    payload
+  } = request
```
Jozzey, line 59: This is more of my preference, I'm not really a fan of the way this is written as it's a bit harder to follow: ` const {
    params: { sessionId },
    payload
  } = request`
  
I prefer `const { sessionId } = request.params` and then passing the `request.payload` to the service. I just find that easier to read. I see that it has been done that way in some other controllers so I guess it's fine to do it like that.

app/presenters/return-logs/setup/meter-provided.presenter.js
```diff
@@ -0,0 +1,33 @@
...
+ * Format data for the `/return-log/setup/{sessionId}/meter-provided` page
+ *
+ * @param {module:SessionModel} session - The return log setup session instance
+ *
+ * @returns {object} page data needed by the view template
+ */
+function go(session) {
+  const {
+    id: sessionId,
+    data: { returnReference },
```
Jozzey, line 18: You should be able to do this thanks to the way the model has been built:
```suggestion
    returnReference,
```

app/presenters/return-logs/setup/meter-provided.presenter.js
```diff
@@ -0,0 +1,33 @@
...
+    meterProvided
+  } = session
+
+  return {
+    backLink: `/system/return-logs/setup/${session.id}/units`,
```
Jozzey, line 23: As you have the variable `sessionId` you could use that instead?
```suggestion
    backLink: `/system/return-logs/setup/${sessionId}/units`,
```

app/views/return-logs/setup/meter-provided.njk
```diff
@@ -0,0 +1,66 @@
...
+        fieldset: {
+          legend: {
+            classes: "govuk-fieldset__legend--l"
+          }
+        },
+        items: [
+          {
+            value: "yes",
+            text: "Yes",
+            checked: "yes" == meterProvided
```
Jozzey, line 53: Is there a reason why `==` has been used, shouldn't it be:
```suggestion
            checked: "yes" === meterProvided
```

app/views/return-logs/setup/meter-provided.njk
```diff
@@ -0,0 +1,66 @@
...
+          },
+          {
+            value: "no",
+            text: "No",
+            checked: "no" == meterProvided
```
Jozzey, line 58: Same for this one:
```suggestion
            checked: "no" === meterProvided
```

app/views/return-logs/setup/units.njk
```diff
@@ -50,22 +50,22 @@
           {
             value: "cubic-meters",
             text: "Cubic Meters",
-            checked: "cubic-meters" == reported
+            checked: "cubic-meters" == units
```
Jozzey, line 53: Is there a reason why `==` has been used, shouldn't it be:
```suggestion
            checked: "cubic-meters" === units
```

app/views/return-logs/setup/units.njk
```diff
@@ -50,22 +50,22 @@
...
           },
           {
             value: "litres",
             text: "Litres",
-            checked: "litres" == reported
+            checked: "litres" == units
```
Jozzey, line 58:
```suggestion
            checked: "litres" === units
```

app/views/return-logs/setup/units.njk
```diff
@@ -50,22 +50,22 @@
...
           },
           {
             value: "mega-litres",
             text: "Mega litres",
-            checked: "mega-litres" == reported
+            checked: "mega-litres" == units
```
Jozzey, line 63:
```suggestion
            checked: "mega-litres" === units
```

app/views/return-logs/setup/units.njk
```diff
@@ -50,22 +50,22 @@
...
           },
           {
             value: "gallons",
             text: "Gallons",
-            checked: "gallons" == reported
+            checked: "gallons" == units
```
Jozzey, line 68:
```suggestion
            checked: "gallons" === units
```

Jozzey (changes requested):
Just a few minor issues

app/views/return-logs/setup/meter-provided.njk
Beckyrose200: Looking at other examples in the njk files where we are comparing the checked values, we use a mix of both, there doesn't seem to be a convention on which to use

app/presenters/return-logs/setup/meter-provided.presenter.js
Beckyrose200: This is going to be done in a tidy up PR:
https://github.com/DEFRA/water-abstraction-system/pull/1626

app/views/return-logs/setup/meter-provided.njk
Jozzey: As discussed. Since these appear to be in quite a few of the views & this isn't going live for a while, we'll have a catch with the tech lead and sort them all out at once if we need to.

Jozzey (approved):

PR #1623, opened 2025-01-16
Return Logs setup - Meter details

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4809

As part of the work to migrate the return logs setup journey, this PR is focused on adding the 'Meter details' page.


Cruikshanks (changes requested):
I don't think we need to break up the validation of the payload. Joi has the ability to validate it as a single entity, something we have done on previous pages.

One worth pairing on using an existing page as an example perhaps?

Cruikshanks (approved):

PR #1636, opened 2025-01-20
Replace `==` in the views with `===`

Jozzey:
It has been spotted in a few views that we have been using `==` rather than `===`.

Generally, we use the strict comparison `===` to compare content & type unless it is necessary to perform a loose equality comparison.

It looks like it is not necessary to use `==` in the views. So where applicable they will be updated to `===`.

### DIFF
--- a/app/views/bill-runs/index.njk
+++ b/app/views/bill-runs/index.njk
@@ -10,17 +10,17 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
       {# Bill runs busy banner #}
-      {% if busy == 'both' %}
+      {% if busy === 'both' %}
         {{ govukNotificationBanner({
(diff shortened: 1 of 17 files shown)


Cruikshanks (commented):
I'm happy to approve. But only if the acceptance tests report zero failures when run.

Could you add a screenshot of the test summary to the PR to confirm this?

Jozzey:
The acceptance tests picked up an issue with `app/views/bill-runs/review/review-licence.njk` which has been fixed. There is just 1 failing acceptance test, but the same test also fails when run on the main branch so is not related to the changes in this PR.

Screenshot 2025-01-21 at 14 17 59
Screenshot 2025-01-21 at 14 18 08

Cruikshanks (approved):

PR #1637, opened 2025-01-20
Remove duplicate recipients

jonathangoulding:

As part of the work to replace the existing notification service we have changed the recipients business rules to remove duplicates.

When a duplicate is found we determine a preferred role and use that role as the only contact detail (this will result in a licence holder only receiving one letter rather than two, as an example)


app/services/notifications/setup/dedupe-recipients.service.js
```diff
@@ -0,0 +1,56 @@
...
+}
+
+function _removeDuplicateContactHash(acc, obj) {
+  if (!acc[obj.contact_hash_id]) {
+    acc[obj.contact_hash_id] = obj
+  } else {
+    const currentRole = obj.contact.role
+    const existingRole = acc[obj.contact_hash_id].contact.role
+
+    if (_licenceHolderAndReturnToRoles(currentRole, existingRole)) {
```
jonathangoulding, line 30: Need to add - 

`For email, if both ‘Primary user’ and ‘Returns agent’ emails are the same, the recipient is listed once`

app/services/notifications/setup/dedupe-recipients.service.js
```diff
@@ -0,0 +1,53 @@
...
+ * De Duplicates recipients for sending notifications
+ *
+ * When presenting or sending notification to recipients we need to remove duplicate contacts
+ *
+ * Deduplicates a list of recipients based on their contact hash ID, preferring certain roles if duplicates are found.
+ *
+ *
+ * @param {object[]} recipients - The array of recipient objects to be deduplicated.
+ * @returns {object[]} - A new array containing the deduplicated recipient objects, with the preferred roles chosen in case of duplicates.
+ */
```
Cruikshanks, line 18:
```suggestion
 * Dedupes the recipients fetched when sending notifications
 *
 * The recipients we fetch using `FetchRecipientsService` will contain _all_ entries. For example, if a licence has both
 * a 'Licence holder' and 'Returns to' contact the service will fetch both.
 *
 * But in a number of cases, the contact entries, once you cater for 'case' (lower and upper), are duplicated. Left as
 * is it means certain users will receive two letters, or two emails, on the same day regarding the same thing.
 *
 * Not very professional!
 *
 * The query `FetchRecipientsService` uses generates a hash ID for each contact entry. This service dedupes the entries
 * based on this hash ID to avoid us spamming our users.
 *
 * Where we have a duplicate, we favour either the licence holder or primary user entry (depending on method type) just
 * to be on the safe side, as it is the licence holder we are obligated to notify when sending comms.
 *
 * @param {object[]} recipients - The array of recipient objects to be deduplicated.
 * 
 * @returns {object[]} - A new array containing the deduplicated recipient objects, with the preferred roles chosen in
 * case of duplicates.
 */
```

Cruikshanks (changes requested):
Had to dash so didn't check every file. But here's my initial feedback!

test/services/notifications/setup/dedupe-recipients.service.test.js
```diff
@@ -0,0 +1,150 @@
...
+        expect(result).to.equal([
+          {
+            all_licences: testDuplicateRecipients.duplicatePrimaryUser.all_licences,
+            contact: null,
+            contact_hash_id: 14567627,
+            message_type: 'Email - both',
+            recipient: 'primary.user@important.com'
+          }
+        ])
+      })
```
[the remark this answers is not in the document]
jonathangoulding: I think this test is redundant now the code has changed from multiple ifs to a better solution.

PR #1644, opened 2025-01-23
Fix megalitres content error return logs setup

Beckyrose200:
https://github.com/DEFRA/water-abstraction-system/pull/1612

The units page has a content error displaying 'megalitres'. The page displays it as two words 'mega litres' when it should be one. This PR fixes this.

### DIFF
--- a/test/presenters/return-logs/setup/units.presenter.test.js
+++ b/test/presenters/return-logs/setup/units.presenter.test.js
@@ -59,7 +59,7 @@ describe('Return Logs Setup - Units presenter', () => {
       })
     })
 
-    describe('when the user has previously selected "Mega litres" as the reported type', () => {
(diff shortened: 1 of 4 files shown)


test/presenters/return-logs/setup/units.presenter.test.js
```diff
@@ -61,13 +61,13 @@ describe('Return Logs Setup - Units presenter', () => {
 
     describe('when the user has previously selected "Mega litres" as the reported type', () => {
```
Jozzey, line 62:
```suggestion
    describe('when the user has previously selected "Megalitres" as the reported type', () => {
```

Jozzey (changes requested):
Just the 1 missed typo

Jozzey (approved):

PR #1648, opened 2025-01-23
Create new quarterly return versions for water companies

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4881

As part of our work in setting up quarterly returns for water companies we need to run a one off migration for each of them to create a new return version. This PR creates an end point that can be hit as part of a job. It will loop through each of the water companies that do not have a return version starting next year and create one.


app/services/jobs/return-logs/fetch-water-undertakers.service.js
```diff
@@ -0,0 +1,58 @@
...
+      builder
+        .select(['id', 'startDate', 'reason'])
+        .where('status', 'current')
+        // A return version must include return requirements in order for us to be able to copy from it
+        .whereExists(
+          ReturnRequirementModel.query()
+            .select(1)
+            .whereColumn('returnVersions.id', 'returnRequirements.returnVersionId')
+        )
+        .orderBy('startDate', 'desc')
```
Cruikshanks, line 42: Do we need them all? Or is it just the latest? In which case, if we add a `.limit(1)`, we can reduce the amount of data coming back.

app/services/jobs/return-logs/return-version-migration.service.js
```diff
@@ -0,0 +1,65 @@
+'use strict'
+
+/**
+ * Determines what licences need to have new return versions created
```
Cruikshanks, line 4: This should match the first line of the JSDoc for the `go()` function. Not fussed which is picked!

app/services/jobs/return-logs/fetch-water-undertakers.service.js
Cruikshanks, line 1: We are not excluding licences that already have a quarterly return version, or where there is a return version that starts after 1 April 2025.

I might be mistaken, but if this is the case we cannot run this more than once.

The idea is that either because of a previous run, or because users have directly dealt with problem ones manually, there may be water company licences that have return versions for 1 April 2025 onwards, so these should be excluded.

Cruikshanks (changes requested):
I stopped when I got to `FetchWaterUndertakersService` for the reasons stated. If I've misunderstood how the code works though do feel free to put me right!

app/services/jobs/return-logs/fetch-water-undertakers.service.js
```diff
@@ -0,0 +1,75 @@
...
+    .where((builder) => {
+      builder.whereNull('lapsedDate').orWhere('lapsedDate', '>=', quarterlyStartDate)
+    })
+    .where((builder) => {
+      builder.whereNull('revokedDate').orWhere('revokedDate', '>=', quarterlyStartDate)
+    })
+    .whereNotExists(
+      ReturnVersionModel.query()
+        .where('returnVersions.status', 'current')
+        .where('quarterlyReturns', true)
```
Cruikshanks, line 59: I am so sorry because I told you to put this there, and you'll have done a lot of testing. 😳 🤦 

The behaviour I intended was if the licence has a return version that starts on or after `2025-04-01`, then its because

- we've already run the migration
- a user has manually added the return version and determined quarterly is not needed

By leaving this clause in, we'll not handle the second scenario. If we take it out, we'll handle both.

So, again, very sorry!

```suggestion
```

Cruikshanks (changes requested):
Just one more small change, and its all my fault!

Cruikshanks (approved):

PR #1659, opened 2025-01-27
Remove where clause on fetch return cycles

robertparkinson:

As part of the testing of this story it was found that we were not voiding all the return logs after the licence end dates return cycle. This PR fixes that by moving the logic into the process return cycle function. When it is now looping through all the return cycles it will not generate returns for cycles after the licence end date but will continue to void them.


app/services/return-logs/process-licence-return-logs.service.js
```diff
@@ -94,19 +90,24 @@ async function _processReturnCycle(returnCycle, returnRequirements, changeDate,
...
-  // Because we've processed _all_ return requirements for the cycle, we no any return logs whose ID is not in
-  // `generatedReturnLogIds` have been made redundant by whatever the 'change' was
-  for (const returnRequirement of requirementsToProcess) {
-    const returnLogIds = await CreateReturnLogsService.go(returnRequirement, returnCycle)
-
-    generatedReturnLogIds.push(...returnLogIds)
+
+  // If there is no licenceEndDate or if there is a licenceEndDate and the return cycle starts before the licenceEndDate
+  // then create the return logs, otherwise just void the return logs for that cycle
+  if (!licenceEndDate || (licenceEndDate && returnCycle.startDate < licenceEndDate)) {
```
Cruikshanks, line 96: Just a NIT!

I believe you can simplify the clause slightly with this.

```suggestion
  if (!licenceEndDate || returnCycle.startDate < licenceEndDate) {
```

app/services/return-logs/process-licence-return-logs.service.js
```diff
@@ -94,19 +90,24 @@ async function _processReturnCycle(returnCycle, returnRequirements, changeDate,
...
+    // Iterate through the requirements and call CreateReturnLogsService. It will generate a return log from the data
+    // provided and attempt to insert it. If it generates a return log that already exists (denoted by the return ID
+    // matching an existing one), the insert will be ignored.
+    //
+    // All generated return ID's are returned by the service and used by VoidLicenceReturnLogsService to identify which
+    // return logs for the given cycle _not_ to mark as 'void'.
+    //
+    // Because we've processed _all_ return requirements for the cycle, we no any return logs whose ID is not in
```
Cruikshanks, line 104: Not your typo, but whilst you are in there! 😬 

```suggestion
    // Because we've processed _all_ return requirements for the cycle, we know any return logs whose ID is not in
```

Cruikshanks (changes requested):
Love it! Just a couple of NITs 😁

Cruikshanks (approved):

PR #1660, opened 2025-01-27
Return Logs setup - Period used

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4853

As part of the work to migrate the return logs setup journey, this PR is focused on adding the 'What period was used for this volume?' page.


Beckyrose200:
Please ignore SonarClouds complaint about code duplication. Its because the Submit Service looks the same to every other Submit Service and it doesn't like it :)

Cruikshanks (approved):

PR #1662, opened 2025-01-27
Add a note to return submissions

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4807

Create functionality to allow the user to add/edit/remove a note from a return submission.

The link to allow the user to add/edit/remove a note will be on the "Check details and enter new volumes or readings" page.

The migrations for the `notes` and `created_by` columns can be found in these PRs:

- https://github.com/DEFRA/water-abstraction-returns/pull/406
- https://github.com/DEFRA/water-abstraction-system/pull/1661


Jozzey:
SonarCloud is complaining about duplicated code. This is because the note service for return logs is very similar to the one for return versions. Whilst similar, they are not exactly the same.

I'm making the assumption that we wouldn't want to try and create some generic notes functionality in this PR that could be used for both journeys to satisfy SonarCloud, so I haven't.

If it is desirable to have the same notes functionality used for both journeys, let me know and I can have a go at sorting that out in another PR

Cruikshanks:
> SonarCloud is complaining about duplicated code. This is because the note service for return logs is very similar to the one for return versions. Whilst similar, they are not exactly the same.
> 
> I'm making the assumption that we wouldn't want to try and create some generic notes functionality in this PR that could be used for both journeys to satisfy SonarCloud, so I haven't.
> 
> If it is desirable to have the same notes functionality used for both journeys, let me know and I can have a go at sorting that out in another PR

💯 agree @Jozzey . Not the time to worry about this, and a common trap the previous team fell into.

Cruikshanks (approved):

PR #1665, opened 2025-01-29
Sort Notifications recipients

jonathangoulding:

This change introduces a sort on the recipients array in the presenter. It sorts by the contacts name.

We need to work out the name before we can sort. So we first map the recipients and then sort.

This will result in all the recipients being mapped and sorted, with only 25 (the current default pagination per page) of the results being used. We have deemed this acceptable for the sake of simplicity and have accounted for performance concerns.

This change also introduces the duplicate recipient fixtures data into the presenter test to ensure recipients are grouped by the name (and subsequently the licence ref)

### DIFF
--- a/test/presenters/notifications/setup/review.presenter.test.js
+++ b/test/presenters/notifications/setup/review.presenter.test.js
@@ -27,6 +27,9 @@ describe('Notifications Setup - Review presenter', () => {
     }
 
     testRecipients = RecipientsFixture.recipients()
+    // This data is used to ensure the recipients are grouped when they have the same licence ref / name.
+    // Ignore the fact that these would be considered duplicates elsewhere in the code
+    // (e.g. contact hash / address being identical)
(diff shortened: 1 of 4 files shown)


test/presenters/notifications/setup/review.presenter.test.js
```diff
@@ -26,8 +27,12 @@ describe('Notifications Setup - Review presenter', () => {
     }
 
     testRecipients = RecipientsFixture.recipients()
+    // This data is used to ensure the recipients are grouped when they have the same licence ref / name.
```
jonathangoulding, line 30: Just want to highlight if this is acceptable. 

We could manually add them in rather than use the fixture (or use the fixtures as a base and create them in this test to have exact message_types etc)

happy to change as i think this is becoming a bit unwieldy to use now we introduce the sort / duplicates.

Cruikshanks (approved):

PR #1667, opened 2025-01-29
Add more contacts to licence contact details

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4879

We need to add the primary user and returns agent to the licence contact details page.

This page shows the contacts stored in crm.

Ths change will update the fetch query to return the 'primary_user' and 'user_agent' if present in the database.

These detail will be displayed in the view as their own contact block.


test/presenters/licences/view-licence-contact-details.presenter.test.js
```diff
@@ -174,7 +174,8 @@ function _testFetchLicenceContactDetailsData() {
           }
         ],
         IsCurrent: false
-      }
+      },
+      licenceEntityRoles: []
```
jonathangoulding, line 178: Add test scenario

test/services/licences/view-licence-contact-details.service.test.js
```diff
@@ -82,7 +82,8 @@ function _testFetchLicenceContactDetailsData() {
         AddressLine2: 'HORIZON HOUSE',
         AddressLine3: 'DEANERY ROAD',
         AddressLine4: ''
-      }
+      },
+      licenceEntityRoles: []
```
jonathangoulding, line 86: Add test scenario

app/services/licences/fetch-licence-contact-details.service.js
```diff
@@ -29,6 +29,14 @@ async function _fetch(licenceId) {
     .modifyGraph('licenceDocumentHeader', (builder) => {
       builder.select(['id', 'metadata'])
     })
+    .withGraphFetched('licenceDocumentHeader.licenceEntityRoles')
```
jonathangoulding, line 32: Add test scenario

app/views/licences/licence-contact-details.njk
```diff
@@ -46,7 +46,16 @@
           value: {
             html: displayAddress(licenceContactDetail.address)
           }
+        } if licenceContactDetail.address,
```
jonathangoulding, line 49: Alternative implementation: 

```
        {
          key: {
          text: "Address" if licenceContactDetail.address else "Email",
          classes:"govuk-!-font-weight-regular"
        },
          value: {
          html:  displayAddress(licenceContactDetail.address)  if licenceContactDetail.address else licenceContactDetail.email
        }
        }
```

app/presenters/licences/view-licence-contact-details.presenter.js
```diff
@@ -20,11 +25,39 @@ function go(licence) {
...
+
+function _sortLicenceEntities(a, b) {
+  if (a.role < b.role) {
+    return -1
+  }
+  if (a.role > b.role) {
+    return 1
+  }
+  return 0
+}
```
Cruikshanks, line 59: Clearly, my Ruby background showing through, but I look to use method names as a means of describing what is going on, whilst avoiding having to document it.

And, much as JavaScript has cool features like treating functions as [First class objects](https://developer.mozilla.org/en-US/docs/Glossary/First-class_Function), I _try_ to only use them when necessary.

This comes from my experience working on projects in Gov; those in charge of assignments see us devs as interchangeable and don't consider existing language skills (i.e. how a .Net dev suddenly became a Ruby, then a Node dev! 😁)

So, I'm always going to prefer to prefer to write code like this, in this style.

```suggestion
function _formatLicenceEntityRoles(licenceEntityRoles) {
  return licenceEntityRoles.map((licenceEntityRole) => {
    return {
      role: ROLES[licenceEntityRole.role],
      email: licenceEntityRole.licenceEntity.name
    }
  })
}

function _licenceEntityRoles(licenceEntityRoles) {
  const formattedLicenceEntityRoles = _formatLicenceEntityRoles(licenceEntityRoles)

  return _sortLicenceEntityRoles(formattedLicenceEntityRoles)
}

function _sortLicenceEntityRoles(licenceEntityRoles) {
  return licenceEntityRoles.sort((a, b) => {
    if (a.role < b.role) {
      return -1
    }

    if (a.role > b.role) {
      return 1
    }

    return 0
  })
}
```

app/presenters/licences/view-licence-contact-details.presenter.js
```diff
@@ -7,6 +7,11 @@ const { filteredContactDetailsByRole } = require('../crm.presenter.js')
  * @module ViewLicenceContactDetailsPresenter
  */
 
+const ROLES = {
```
Cruikshanks, line 10: Just so they don't get confused with the contact roles in the metadata, I'd suggest going a bit more verbose.

```suggestion
const ENTITY_ROLES = {
```

Cruikshanks (changes requested):
Admittedly, the suggested changes are a personal preference. But they are consistent with stuff were have done before.

Cruikshanks (approved):

PR #1671, opened 2025-01-30
Add 'Mark as under query' button to View Return Logs page

StuAA78:
https://eaflood.atlassian.net/browse/WATER-4893


app/views/return-logs/view.njk
```diff
@@ -104,20 +112,46 @@
...
+                classes: "govuk-button--secondary",
+                name: "mark-query",
+                value: queryButtonValue,
+                preventDoubleClick: true
+              })
+            }}
+          {% if actionButton %}
+          </div>
+        </form>
+        {% endif %}
```
Cruikshanks, line 152: 💯 get and agree about the need to put the button group div inside the `if`. But it looks like things are a little out of alignment. Plus I think the closing `</form>` should be outside the second `if`.

```suggestion
        <form method="post">
          <input type="hidden" name="wrlsCrumb" value="{{wrlsCrumb}}"/>
          {% if actionButton %}
            {# The button group div goes inside the if because we only need a group when there's two buttons #}
            <div class="govuk-button-group">
              {# Submit/Edit return button #}
              {{
                govukButton({
                  text: actionButton.text,
                  preventDoubleClick: true,
                  href: actionButton.href
                })
              }}
          {% endif %}

          {# Mark as under query / Resolve query button #}
          {% if underQuery %}
            {% set queryButtonText = 'Resolve query' %}
            {% set queryButtonValue = 'resolve' %}
          {% else %}
            {% set queryButtonText = 'Mark as under query' %}
            {% set queryButtonValue = 'mark' %}
          {% endif %}
          {{
            govukButton({
              text: queryButtonText,
              classes: "govuk-button--secondary",
              name: "mark-query",
              value: queryButtonValue,
              preventDoubleClick: true
            })
          }}

          {% if actionButton %}
            </div>
          {% endif %}
        </form>
```

test/presenters/return-logs/view-return-log.presenter.test.js
```diff
@@ -608,6 +608,36 @@ describe('View Return Log presenter', () => {
...
+        expect(result.underQuery).to.equal(true)
+      })
+    })
+
+    describe('when the return log is not under query', () => {
+      beforeEach(() => {
+        testReturnLog.underQuery = false
+      })
+
+      it('returns true', () => {
```
Cruikshanks, line 634:
```suggestion
      it('returns false', () => {
```

Cruikshanks (changes requested):
Just a couple of NITs! 😁

Cruikshanks (approved):

PR #1677, opened 2025-02-03
Refactor existing CSV converter

jonathangoulding:

As part of the work to implement notifications in the system repo we have found some logic that will need to be duplicated in upcoming work.

This change lifts the convert to csv service into the lib.

It has been renamed to transform to csv,

The existing logic has been updated to be more precise in the transformation it is doing (transforms an array into a CSV row).


app/lib/transform-to-csv.lib.js
```diff
@@ -1,32 +1,33 @@
...
+ * (string, number, object, boolean, date, etc.)
+ *
+ * @returns {string} A CSV formatted string with each value separated by commas,
+ * with a separated by newline characters (`\n`) to signify the end of the row.
+ *
+ * @private
  */
-function go(data) {
-  if (!data) {
+function TransformArrayToCSVRow(arrayToTransform) {
```
Cruikshanks, line 25: NIT - All our other library functions, the name starts with a lowercase character.

```suggestion
function transformArrayToCSVRow(arrayToTransform) {
```

Cruikshanks (changes requested):
Just one NIT

app/lib/transform-to-csv.lib.js
jonathangoulding: Updated in all places

Cruikshanks (approved):

PR #1682, opened 2025-02-04
Add notifications download route

jonathangoulding:

As part of the work to implement notifications in the system repo we have a requirement to download the recipients data as a CSV.

This change adds the controller, service and route for the notifications download link.

This is the plumbing before adding the presenter and joining the fetch service in.

### DIFF
--- a/app/services/notifications/setup/download-recipients.service.js
+++ b/app/services/notifications/setup/download-recipients.service.js
@@ -5,6 +5,8 @@
  * @module DownloadRecipientsService
  */
 
+const SessionModel = require('../../../models/session.model.js')
+
 /**
  * Orchestrates fetching and formatting the data needed for the notifications setup download link
  *
@@ -12,15 +14,20 @@
  *
  * It does not seem necessary to use a data stream to create the csv as the data is relatively small.
  *
+ * @param {string} sessionId - The UUID for setup ad-hoc returns notification session record
+ *
  * @returns {Promise<object>} The data for the download link (csv string, filename and type)
  */
-async function go() {
+async function go(sessionId) {
+  const session = await SessionModel.query().findById(sessionId)
+  const { notificationType, referenceCode } = session
+
   const csv = 'Licences\n12234\n'
 
   return {
     data: csv,
     type: 'text/csv',
-    filename: 'recipients.csv'
+    filename: `${notificationType} - ${referenceCode}.csv`
   }
 }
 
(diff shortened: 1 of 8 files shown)


app/services/notifications/setup/download-recipients.service.js
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @returns {Promise<object>} The data for the download link
+ */
+async function go() {
+  const csv = 'Licences\n12234\n'
+
+  return {
+    data: csv,
+    type: 'text/csv',
+    filename: 'recipients.csv'
```
jonathangoulding, line 22: Are we happy with this name ? 

Do we need to add the due date / sleected returns period ? 

`summer-2022-03-01-recipients.csv` ?

app/services/notifications/setup/initiate-session.service.js
```diff
@@ -1,29 +1,68 @@
...
+}
+
+/**
+ * A function to generate a pseudo-unique reference code for recipients notifications
+ *
+ * @param {string} prefix
+ *
+ * @returns {string} A reference code with a prefix and random string (RINV-A14GB8)
+ */
+function _generateReferenceCode(prefix) {
```
jonathangoulding, line 57: This function replicates the existing naming convention for the csv file. 

https://github.com/DEFRA/water-abstraction-service/blob/main/src/lib/reference-generator.js

app/services/notifications/setup/download-recipients.service.js
jonathangoulding: Updated

test/services/notifications/setup/download-recipients.service.test.js
```diff
@@ -0,0 +1,35 @@
...
+
+// Test helpers
+const SessionHelper = require('../../../support/helpers/session.helper.js')
+
+// Thing under test
+const DownloadRecipientsService = require('../../../../app/services/notifications/setup/download-recipients.service.js')
+
+describe('Notifications Setup - Download recipients service', () => {
+  let session
+  const referenceCode = 'RINV-00R1MQ'
```
Cruikshanks, line 18: NIT - We try to place `const`s before `let`s separated by a new line.

```suggestion
  const referenceCode = 'RINV-00R1MQ'

  let session  
```

Cruikshanks (changes requested):
Just one NIT!

Cruikshanks (approved):

PR #1684, opened 2025-02-05
Add notifications download presenter

jonathangoulding:

As part of the work to implement notifications in the system repo we have a requirement to download the recipients data as a CSV.

This change updates the notification download service to use the fetch download service and download recipients presenter.

The download recipients presenter will map the data into a csv format. The headers are defined and set in the order expected for the csv and the rows mapped based on the recipient data.

The notion of a Registered licence and unregistered licence is used in the CSV to signify the notification method (currently email / letter).

### DIFF
--- a/app/presenters/notifications/setup/download-recipients.presenter.js
+++ b/app/presenters/notifications/setup/download-recipients.presenter.js
@@ -37,39 +37,27 @@ function go(recipients) {
  */
 function _transformForCsv(recipients) {
   return recipients.map((recipient) => {
-    const common = {
+    const { contact } = recipient
+
+    return {
       Licences: recipient.licence_ref,
       'Return references': recipient.return_reference,
       'Returns period start date': formatDateObjectToISO(recipient.start_date),
       'Returns period end date': formatDateObjectToISO(recipient.end_date),
       'Returns due date': formatDateObjectToISO(recipient.due_date),
-      'Message type': 'email',
+      'Message type': contact ? 'letter' : 'email',
       'Message reference': 'invitations',
-      'Licence holder': '',
+      'Licence holder': contact ? contactName(recipient.contact) : '',
       'Recipient name': '',
       Email: recipient.email || '',
-      'Address line 1': '',
-      'Address line 2': '',
-      'Address line 3': '',
-      'Address line 4': '',
-      'Address line 5': '',
-      'Address line 6': '',
-      Postcode: ''
+      'Address line 1': contact ? contact.addressLine1 : '',
+      'Address line 2': contact ? contact.addressLine2 : '',
+      'Address line 3': contact ? contact.addressLine3 : '',
+      'Address line 4': contact ? contact.addressLine4 : '',
+      'Address line 5': contact?.town || contact?.county,
+      'Address line 6': contact ? contact.country : '',
+      Postcode: contact ? contact.postcode : ''
     }
-
-    if (recipient.contact) {
-      common['Message type'] = 'letter'
-      common['Licence holder'] = contactName(recipient.contact)
-      common['Address line 1'] = recipient.contact.addressLine1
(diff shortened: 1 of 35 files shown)


app/presenters/notifications/setup/download-recipients.presenter.js
```diff
@@ -0,0 +1,89 @@
...
+    return {
+      Licences: recipient.licence_ref,
+      'Return references': recipient.return_reference,
+      'Returns period start date': formatDateObjectToISO(recipient.start_date),
+      'Returns period end date': formatDateObjectToISO(recipient.end_date),
+      'Returns due date': formatDateObjectToISO(recipient.due_date),
+      'Message type': contact ? 'letter' : 'email',
+      'Message reference': 'invitations',
+      'Licence holder': contact ? contactName(recipient.contact) : '',
+      'Recipient name': '',
```
jonathangoulding, line 51: Are we removing this based on discussions.

app/presenters/notifications/setup/download-recipients.presenter.js
```diff
@@ -0,0 +1,89 @@
...
+      Email: recipient.email || '',
+      'Address line 1': contact ? contact.addressLine1 : '',
```
jonathangoulding, line 53: I was going to lift this out into a seperate function for the addresses. But we loose the contaxt of the header order.

app/presenters/notifications/setup/download-recipients.presenter.js
Cruikshanks: The discussion was that the `Licence holder` column was to be dropped.

However, the logic you have (`contact ? contactName(recipient.contact) : '',`) is supposed to be in the `Recipient name` column.

Cruikshanks: Can we also swap the order so `Email` comes before `Recipient name`.

app/presenters/notifications/setup/download-recipients.presenter.js
Cruikshanks, line 1: I'm sure that it won't make a noticeable difference to the timings. But I think it would make things simpler, because there would be no need for `_transformJsonToCsv()`.

As we know the headers, why not have them declared as constant, for example,

```javascript
const { transformArrayToCSVRow } = require('../../../lib/transform-to-csv.lib.js')

const HEADERS = [
  'Licences',
  'Return references',
  'Returns period start date',
  'Returns period end date',
  'Returns due date',
  'Message type',
  'Message reference',
  'Licence holder',
  'Recipient name'
]
```

And then `_transformForCsv()` just needs to return an array of arrays, or even better, return the result of `transformArrayToCSVRow()`.

```javascript
function _transformForCsv(recipients) {
  return recipients.map((recipient) => {
    const { contact } = recipient

    const row = [
      recipient.licence_ref,
      recipient.return_reference,
      formatDateObjectToISO(recipient.start_date),
      formatDateObjectToISO(recipient.end_date),
      formatDateObjectToISO(recipient.due_date),
      contact ? 'letter' : 'email',
      'invitations',
      recipient.email || '',
      contact ? contactName(recipient.contact) : '',
      ..._address(contact)
    ]

    return transformArrayToCSVRow(row)
  })
}
```

At the moment, we're iterating the `recipients` twice, and putting the result of our first iteration through another transform step. I think we can skip that second iteration and transform with this change. What do you think?

Shout if I have missed something, which makes my suggestion moot.

Cruikshanks (changes requested):
We need to make a change to the output. But I think we can also simplify the presenter, unless I've overlooked something.

app/presenters/notifications/setup/download-recipients.presenter.js
jonathangoulding: Updated.

PR #1695, opened 2025-02-07
Add remove licences from notification queries

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4777

As part of the work to implement notifications in the system repo we have a requirement allow the user to remove licences from teh recipients list.

This change adds the functionality to exclude licences from the fetch recipients and fetch download recipients service queries.

### DIFF
--- a/app/services/notifications/setup/fetch-download-recipients.service.js
+++ b/app/services/notifications/setup/fetch-download-recipients.service.js
@@ -45,22 +45,26 @@ const { db } = require('../../../../db/db.js')
  *
  * @param {Date} dueDate
  * @param {string} summer
+ * @param {string[]} removeLicences - The licences to exclude from the results
  *
  * @returns {Promise<object[]>} - matching recipients
  */
-async function go(dueDate, summer) {
-  const { rows } = await _fetch(dueDate, summer)
+async function go(dueDate, summer, removeLicences = ['']) {
+  const { rows } = await _fetch(dueDate, summer, removeLicences)
 
   return rows
 }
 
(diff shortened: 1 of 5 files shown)


app/services/notifications/setup/fetch-download-recipients.service.js
Cruikshanks, line 1: There is an alternative way to do a query with a dynamic set of values for the arguments.

Rather than dynamically generating `?` placeholders and using `WHERE ldh.licence_ref NOT IN ()`, instead use `WHERE NOT (ldh.licence_ref = ANY())`.

- [Postgres Query Boost: Using ANY Instead of IN](https://www.crunchydata.com/blog/postgres-query-boost-using-any-instead-of-in)
- [IN vs ANY operator in PostgreSQL](https://stackoverflow.com/a/34627688/6117745)

Here's an example of it in use in the **water-abstraction-import** app https://github.com/DEFRA/water-abstraction-import/blob/main/src/modules/return-logs/lib/void-returns.js#L10-L19

I think it will help remove some code and make things a little easier to read/follow.

Cruikshanks (changes requested):
Stopping with my one suggestion as it will mean some changes across both files.

Technically it should work, but let me know if you have any other issues with this approach.

app/services/notifications/setup/fetch-download-recipients.service.js
jonathangoulding: Updated

Cruikshanks (approved):

PR #1702, opened 2025-02-11
Update remove licence validation to check licence refs

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4777

As part of the work to implement notifications in the system repo we have a requirement allow the user to remove licences from teh recipients list.

This change updates the validation for the remove licences page. The user can input anything into the text field. We want to check that each licence is valid before displaying the recipients list or creating the download csv.

The licences to exclude are provided and checked against the due returns logs.

### DIFF
--- a/app/validators/notifications/setup/remove-licences.validator.js
+++ b/app/validators/notifications/setup/remove-licences.validator.js
@@ -12,7 +12,7 @@ const { transformStringOfLicencesToArray } = require('../../../lib/general.lib.j
  * Validates data submitted for the `/notifications/setup/remove-licences` page
  *
  * @param {object} payload - The payload from the request to be validated
- * @param {object[]} validLicences - The licences present in the database
+ * @param {object[]} validLicences - An array of licences found in the database from `payload.removeLicences`
  *
  * @returns {object} the result from calling Joi's schema.validate(). It will be an object with a `value:` property. If
  * any errors are found the `error:` property will also exist detailing what the issues were
--- a/app/validators/notifications/setup/remove-licences.validator.js
+++ b/app/validators/notifications/setup/remove-licences.validator.js
@@ -6,38 +6,28 @@
  */
 
 const Joi = require('joi')
+
 const { transformStringOfLicencesToArray } = require('../../../lib/general.lib.js')
 
 /**
  * Validates data submitted for the `/notifications/setup/remove-licences` page
  *
  * @param {object} payload - The payload from the request to be validated
- * @param {object[]} validLicences - An array of licences found in the database from `payload.removeLicences`
+ * @param {object[]} validLicences - The licences that that have due returns for the selected notification period
  *
  * @returns {object} the result from calling Joi's schema.validate(). It will be an object with a `value:` property. If
  * any errors are found the `error:` property will also exist detailing what the issues were
  */
 function go(payload, validLicences) {
-  const schema = Joi.object(
-    {
-      removeLicences: Joi.custom((value, helpers) => {
-        const expectedLicences = validLicences.map((licence) => licence.licenceRef)
-        const result = transformStringOfLicencesToArray(value)
+  const validLicenceRefs = validLicences.map((licence) => {
+    return licence.licenceRef
+  })
 
-        const missingLicences = _findMissingElements(result, expectedLicences)
-
-        if (missingLicences.length > 0) {
-          const messageKey = missingLicences.length === 1 ? 'missingLicence' : 'missingLicences'
-          const missingValue = missingLicences.length === 1 ? missingLicences[0] : missingLicences.join(', ')
-
-          return helpers.error(messageKey, { missingLicences: missingValue })
-        }
-
-        return result
-      })
-    },
-    'Custom Licence Validation'
-  ).messages({
+  const schema = Joi.object({
+    removeLicences: Joi.custom((value, helpers) => {
+      return _removedLicencesInValidLicencesValidator(value, helpers, validLicenceRefs)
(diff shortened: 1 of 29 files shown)


app/validators/notifications/setup/remove-licences.validator.js
Cruikshanks, line 1: To be honest, there are quite a number of corrections and changes I'd make to this file. So, it's easier if I just provide a version of how I'd go about it.

```javascript
'use strict'

/**
 * Validates data submitted for the `/notifications/setup/remove-licences` page
 * @module RemoveLicencesValidator
 */

const Joi = require('joi')

const { transformStringOfLicencesToArray } = require('../../../lib/general.lib.js')

/**
 * Validates data submitted for the `/notifications/setup/remove-licences` page
 *
 * @param {object} payload - The payload from the request to be validated
 * @param {object[]} validLicences - The licences that that have due returns for the selected notification period
 *
 * @returns {object} the result from calling Joi's schema.validate(). It will be an object with a `value:` property. If
 * any errors are found the `error:` property will also exist detailing what the issues were
 */
function go(payload, validLicences) {
  const validLicenceRefs = validLicences.map((licence) => {
    return licence.licenceRef
  })

  const schema = Joi.object({
    removeLicences: Joi.custom((value, helpers) => {
      return _removedLicencesInValidLicencesValidator(value, helpers, validLicenceRefs)
    }, 'Custom Licence Validation')
  }).messages({
    missingLicence: 'There are no returns due for licence {{#missingLicences}}',
    missingLicences: 'There are no returns due for licences {{#missingLicences}}'
  })

  return schema.validate(payload, { abortEarly: false })
}

function _findRemovedLicencesNotInValidLicences(licenceRefsToRemove, validLicenceRefs) {
  const missingLicences = []

  for (const licenceRefToRemove of licenceRefsToRemove) {
    if (!validLicenceRefs.includes(licenceRefToRemove)) {
      missingLicences.push(licenceRefToRemove)
    }
  }

  return missingLicences
}

function _removedLicencesInValidLicencesValidator(value, helpers, validLicenceRefs) {
  const result = transformStringOfLicencesToArray(value)

  const missingLicences = _findRemovedLicencesNotInValidLicences(result, validLicenceRefs)

  if (missingLicences.length > 0) {
    const messageKey = missingLicences.length === 1 ? 'missingLicence' : 'missingLicences'
    const missingValue = missingLicences.join(', ')

    return helpers.error(messageKey, { missingLicences: missingValue })
  }

  return result
}

module.exports = {
  go
}
```

test/validators/notifications/setup/remove-licences.validator.test.js
Cruikshanks, line 1: Same again for this file

```javascript
'use strict'

const { describe, it, beforeEach } = (exports.lab = Lab.script())
const { expect } = Code

// Thing under test
const RemoveLicencesValidator = require('../../../../app/validators/notifications/setup/remove-licences.validator.js')

describe('Notifications Setup - Remove licences validator', () => {
  let validLicences
  let payload

  beforeEach(() => {
    validLicences = [{ licenceRef: '123/67' }]
  })

  describe('when licences are valid to be removed', () => {
    beforeEach(() => {
      payload = { removeLicences: '123/67' }
    })

    it('confirms the data is valid', () => {
      const result = RemoveLicencesValidator.go(payload, validLicences)

    })
  })

  describe('when invalid data is provided', () => {
    describe('because a licence is not a valid licence to be removed', () => {
      beforeEach(() => {
        payload = { removeLicences: '01/123' }
      })

      it('fails validation', () => {
        const result = RemoveLicencesValidator.go(payload, validLicences)

        expect(result.error.details[0].message).to.equal('There are no returns due for licence 01/123')
      })
    })

    describe('because multiple licences are not valid licences to be removed', () => {
      beforeEach(() => {
        payload = { removeLicences: '01/123,678' }
      })

      it('fails validation', () => {
        const result = RemoveLicencesValidator.go(payload, validLicences)

        expect(result.error.details[0].message).to.equal('There are no returns due for licences 01/123, 678')
      })
    })
  })
})
```

Cruikshanks (changes requested):
I have some changes for the validator 😬

app/validators/notifications/setup/remove-licences.validator.js
jonathangoulding: Much better.

test/services/notifications/setup/fetch-returns-due.service.test.js
```diff
@@ -0,0 +1,47 @@
...
+    await ReturnLogHelper.add({})
+
+    dueDate = returnLog.dueDate
+
+    licenceRefs = [returnLog.licenceRef]
+  })
+
+  describe('when there are licences', () => {
+    it('correctly returns the matching licences', async () => {
+      const result = await FetchReturnsDueService.go(licenceRefs, dueDate, isSummer)
```
Cruikshanks, line 38: We have a `dueDate` variable which we set in the `beforeEach()`. But we don't then use the value when creating the return logs.

Instead, we overwrite what we set with whatever the helper used when creating the return  log. And, though I know the helper will use the same 'due date', it's not clear from the test that the second return log will have the same values as the first.

I'd suggest passing in `dueDate` and `isSummer` to both calls to `ReturnLogHelper.add({})` so it is clearer the difference between them is the licence ref (might help to specify those as well? 🤔 )

N.B. If you don't need to override any defaults, you don't need to pass an empty object in.

Cruikshanks (changes requested):
Some changes needed to one of the tests.

Cruikshanks (approved):

PR #1706, opened 2025-02-12
Support reissue return logs 4 'no returns required'

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4552

When a user creates a no return required return version the return logs need to be updated. This PR adds in the logic to correctly void the return logs from the start date of the no returns required version up to the current date or up to the start of any future return version.


Cruikshanks (changes requested):
The new service is called from the same `Submit` service that the return version it depends on gets generated and then persisted in.

Rather than query for the return version we've just created, could we not just pass it in?

app/services/return-logs/void-no-return-required-licence-return-logs.service.js
Cruikshanks, line 1: I love a verbose name! But I think we could live with `void-return-logs.service.js`. 😁 

Yes, it is intended to be used in the scenario of voiding a licence's return logs when a no returns required return version has been added.

But there isn't anything in the code that makes it only applicable to this scenario. When we call `VoidReturnLogsService.go()` in the context we do, I reckon that is enough.

app/services/return-logs/void-no-return-required-licence-return-logs.service.js
```diff
@@ -0,0 +1,33 @@
...
+const ReturnLogModel = require('../../models/return-log.model.js')
+
+/**
+ * Handles voiding the return logs for a licence after a no returns required return version has been created
+ *
+ * @param {string} licenceRef - The licenceRef of the licence
+ * @param {Date} startDate - The start date of no returns required return version
+ * @param {Date} endDate - The optional end date of no returns required return version
+ */
+async function go(licenceRef, startDate, endDate) {
```
Cruikshanks, line 18:
```suggestion
 * @param {Date} [endDate=null] - The optional end date of no returns required return version
 */
async function go(licenceRef, startDate, endDate = null) {
```

test/services/return-logs/void-no-return-required-licence-return-logs.service.test.js
```diff
@@ -0,0 +1,86 @@
...
+        returnVersion.endDate
+      )
+
+      returnLogBeingChecked = await returnLogMatchingVersion.$query()
+      expect(returnLogBeingChecked.status).to.equal('void')
+
+      returnLogBeingChecked = await returnLogNotMatchingVersion.$query()
+      expect(returnLogBeingChecked.status).to.equal('due')
+    })
+  })
```
Cruikshanks, line 58: A bit like the file name, I think this description is trying to convey the scenario this would be used in, but is confusing when applied to the service: we don't provide it a licence ID or the licence's end date.

We're also just testing a service that just patches return logs, and doesn't care about the return versions they are linked to. So, I am not sure we need to create any return versions as part of this test. I actually think we don't even need to create a licence record.

So, perhaps ...

- when provided with a licence ref
  - for which there are return logs
    - and no "endDate" is provided
    - and an "endDate" is provided
  - for which there are no return logs

🤷

Cruikshanks (changes requested):

test/services/return-logs/void-return-logs.service.test.js
```diff
@@ -0,0 +1,86 @@
...
+        startDate: new Date('2022-04-01')
+      })
+      returnLogNotMatchingVersion = await ReturnLogHelper.add({
+        endDate: new Date('2024-03-31'),
+        licenceRef,
+        startDate: new Date('2023-04-01')
+      })
+    })
+
+    it('voids the return logs that are from the start date of the return version forward', async () => {
```
Cruikshanks, line 64:
```suggestion
    it('voids the return logs that are from the start date forward', async () => {
```

Cruikshanks (changes requested):

Cruikshanks (approved):

PR #1708, opened 2025-02-12
Update notifications setup initiate session

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4899

As part of the work to implement notifications in the system repo we have multiple journeys for notifications.

This change updates ad-hoc returns initiate session to use a shared initiate session.

This change also adds a 'journey' key to session for the notifications setup journey to enable alternative logic for multiple notification types.


jonathangoulding:
Relies on - https://github.com/DEFRA/water-abstraction-ui/pull/2665

Cruikshanks (approved):

PR #1709, opened 2025-02-12
Update notifications fetch services to use the session

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4899

As part of the work to implement notifications in the system repo we have multiple journeys for notifications.

This change updates the notification setup fetch services to use the session.

This will enable us to use the journey context within the query to handle ad-hoc returns and other notification types.

This change for ad-hoc will come at in a later change.


app/services/notifications/setup/fetch-recipients.service.js
```diff
@@ -144,14 +146,16 @@ const { db } = require('../../../../db/db.js')
...
- * @param {string} summer - Whether we are looking for outstanding summer or all year return logs
- * @param {string[]} removeLicences - The licences to exclude from the results
+ * @param {module:SessionModel} session - The session instance to format
  *
  * @returns {Promise<object[]>} The contact data for all the outstanding return logs
  */
-async function go(dueDate, summer, removeLicences) {
-  const { rows } = await _fetch(dueDate, summer, removeLicences)
+async function go(session) {
+  const { returnsPeriod, summer } = DetermineReturnsPeriodService.go(session.returnsPeriod)
```
jonathangoulding, line 154: In the next PR we can introduce the journey context and move this into something like `_fetchMultpleLicences()` and move the js docs into this private function in preparation for the `_fetchByLicences()` / 'ad-hoc' step to be introduced. 

Just felt alot was changing that it would be better to separate them out.

Cruikshanks (approved):

PR #1718, opened 2025-02-13
Add notifications setup ad hoc licence fetch

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4900

This change introduces the 'journey' context into the notifications setup fetch services.

Due to the complexity of the queries we have decided to pass the session into the fetch services. The session contains a journey key which we use to determine what query to make.

When the journey is for an ad hoc licence we only want to fetch the recipients for that licence (we are not concerned with the returns period, due date and if is a summer return).

This change replaces some of the existing fetch requests 'where' clauses. As the query is a template literal string we can 'inject' the relevant where clause into the query.

The previous returns periods query results should not be affected.


jonathangoulding:
I will update the page title and any other issues (not related to the fetch) in another pr.

Cruikshanks (approved):

PR #1733, opened 2025-02-17
Splitting single volume on return log lines

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4853

As part of the work for the final page on the return log setup journey, a user can enter a single volume and an abstraction period for which the single volume was taken in. 

With the pages built all that's left is for the backend to split this single volume amongst the lines that the abstraction period covers on the return. 

When a total volume is divided across multiple lines, some of the amounts may have long decimal places. If you add them all back together, the total can end up slightly higher than the original due to small rounding differences. To make sure the total stays correct, rounding is only applied to the last line.


app/validators/return-logs/setup/period-used.validator.js
```diff
@@ -55,10 +55,14 @@ function _fromDateBeforeToDate(value, helpers) {
...
+  if (!year && !month && !day) {
+    return null
+  }
+
   const paddedMonth = month ? leftPadZeroes(month, 2) : ''
   const paddedDay = day ? leftPadZeroes(day, 2) : ''
 
-  return `${year}-${paddedMonth}-${paddedDay}`
+  return new Date(`${year}-${paddedMonth}-${paddedDay}`)
 }
```
Jozzey, line 66: I assume that this `if` is to stop the `new Date` on line 65 from blowing up if any values are missing. As `new Date` will go pop if any of the values are missing, I'd use `OR` rather than `AND` in the if statement as all values are needed. Also on lines 62 & 63, if the month or day didn't exist and an empty string was used this would still cause the `new Date` to break. Apart from that, when creating the new date it is not required to pad the day and month with zeros.
If re-write the function like this:
```suggestion
function _fullDate(day, month, year) {
  if (!year || !month || !day) {
    return null
  }

  return new Date(`${year}-${month}-${day}`)
}
```

app/services/return-logs/setup/allocate-single-volume-to-lines.service.js
```diff
@@ -0,0 +1,76 @@
...
+  })
+
+  if (allocatedLineTotal !== Number(singleVolume)) {
+    const roundingError = singleVolume - allocatedLineTotal
+    const lastIndex = linesInsideAbstractionPeriod.length - 1
+    linesInsideAbstractionPeriod[lastIndex].quantity = linesInsideAbstractionPeriod[lastIndex].quantity + roundingError
+  }
+}
+
+function _linesInsideAbstractionPeriod(lines, fromDate, toDate) {
```
StuAA78, line 53: This function could be streamlined using `.filter()` instead of `.forEach()`; something like:

```js
const abstractionPeriodLines = lines.filter((line) => {
  delete line.quality // Don't need the if, as deleting a property that doesn't exist doesn't cause an error
  return _lineWithinAbstractionPeriod(line.startDate, line.endDate, fromDate, toDate)
})

return abstractionPeriodLines
```

Or even just:

```js
return lines.filter((line) => {
  // ...
})
```

app/validators/return-logs/setup/period-used.validator.js
Jozzey, line 1: I know this isn't part of this PR but regarding lines 32 and 33. This doesn't quite look right where we are adding `fromFullDate` and `toFullDate` to the payload. Personally, I wouldn't add items to the payload like this; I would have just kept them as separate variables.

app/validators/return-logs/setup/period-used.validator.js
```diff
@@ -55,10 +55,14 @@ function _fromDateBeforeToDate(value, helpers) {
...
 
 function _fullDate(day, month, year) {
+  if (!year && !month && !day) {
```
StuAA78, line 58: Just wanted to check that the logic here is correct -- it only returns `null` if _all_ of these are falsy?

StuAA78 (changes requested):
Looks good! Just some small suggested tweaks.

test/services/return-logs/setup/allocate-single-volume-to-lines.service.test.js
```diff
@@ -0,0 +1,149 @@
...
+
+    describe('and a single volume that cannot be divided equally', () => {
+      beforeEach(() => {
+        singleVolume = 1001
+        fromDate = new Date('2023-04-01').toISOString()
+        toDate = new Date('2023-09-30').toISOString()
+        lines = _lines()
+      })
+
+      it('handles rounding errors by adjusting the last line', () => {
```
Jozzey, line 92: Could we add a test that checks if the sum of all the quantities on the lines where the `singleVolume` can't be divided equally equals the `singleVolume`.

Jozzey (changes requested):
A few bits to have a look at

Jozzey (approved):

StuAA78 (approved):
👍

PR #1734, opened 2025-02-17
Add notifications setup cancel page

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4901

This change add the cancel page for the notifications setup journey.

The change displays a new view to the user giving them the option to cancel the notifications they are about the send.

This page works for both 'ad-hoc' and other journeys. In the case of an 'ad-hoc' journey we show the licence number the user previously entered. For the other journeys we will show the returns period selected.

This change also handle the submit action and will clear the session and redirect the user to the /manage page.


test/presenters/notifications/setup/cancel.presenter.test.js
```diff
@@ -0,0 +1,145 @@
...
+      const result = CancelPresenter.go(session)
+
+      expect(result.summaryList).to.equal({
+        text: 'Returns period',
+        value: 'Quarterly 1 April 2025 to 30 June 2025'
+      })
+    })
+  })
+
+  describe('when the journey has is for a "returnsPeriod"', () => {
```
Cruikshanks, line 96: There is an errant word in the description. I opted for this variant but either is fine 🤷 

```suggestion
  describe('when the journey has a "returnsPeriod"', () => {
```

Cruikshanks (changes requested):
Just the smallest of typos.

Cruikshanks (approved):

PR #1736, opened 2025-02-18
Download return log as CSV

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4860

This PR adds the functionality to allow internal users to download a completed return as a CSV file. The downloaded file will accurately reflect the currently viewed version of the return.


app/controllers/return-logs.controller.js
```diff
@@ -8,9 +8,24 @@
 const Boom = require('@hapi/boom')
 
 const SubmitViewReturnLogService = require('../services/return-logs/submit-view-return-log.service.js')
-
+const DownloadReturnLogService = require('../services/return-logs/download-return-log.service.js')
```
Cruikshanks, line 11: Needs to be added in alphabetical order.

app/controllers/return-logs.controller.js
```diff
@@ -8,9 +8,24 @@
...
 const ViewReturnLogService = require('../services/return-logs/view-return-log.service.js')
 
+async function downloadReturnLog(request, h) {
```
Cruikshanks, line 14: IMHO we can just called this handler `download`. It's on the return logs controller, so not only can this be assumed, it matches the convention of other handlers (referring to the action and not mentioning the root resource).

app/presenters/return-logs/download-return-log.presenter.js
```diff
@@ -0,0 +1,64 @@
...
+function _selectedReturnSubmission(returnSubmissions) {
+  if (!returnSubmissions) {
+    return null
+  }
+
+  return returnSubmissions[0]
+}
+module.exports = {
+  go
+}
```
Cruikshanks, line 64: It should be that you cannot click the download button unless the return log has a return submission with return submission lines.

As such, this scenario should never happen, which means we should not guard against it. That'll effectively make this method redundant which should also simplify things a little.

N.B. Also needs to be a space between the end of the function and `module.exports = {` 😁

app/services/return-logs/download-return-log.service.js
```diff
@@ -0,0 +1,33 @@
...
+ * @module DownloadReturnLogService
+ */
+
+const DownloadReturnLogPresenter = require('../../presenters/return-logs/download-return-log.presenter.js')
+const FetchDownloadReturnLogService = require('./fetch-download-return-log.service.js')
+
+/**
+ * Orchestrates fetching and formatting the data needed for the return log download link
+ *
+ * @param {string} returnLogId - The UUID of the return
```
Cruikshanks, line 14:
```suggestion
 * @param {string} returnLogId - The UUID of the return log to download
```

app/services/return-logs/download-return-log.service.js
```diff
@@ -0,0 +1,33 @@
...
+ * @param {string} version - The version number of the return
```
Cruikshanks, line 15:
```suggestion
 * @param {string} version - The version number of the return submission to download
```

app/services/return-logs/fetch-download-return-log.service.js
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+const ReturnLogModel = require('../../models/return-log.model.js')
+const ReturnSubmissionModel = require('../../models/return-submission.model.js')
+
+/**
+ * Fetches the matching return log data and associated submission needed for the csv download
+ *
+ * @param {string} returnLogId - The return log ID
+ * @param {number} version - Version number of the submission to download csv
```
Cruikshanks, line 15:
```suggestion
 * @param {string} returnLogId - The UUID of the return log to fetch for download
 * @param {number} version - The version number of the submission data to use
```

app/services/return-logs/fetch-download-return-log.service.js
Cruikshanks, line 1: This looks to be based on the fetch service we use for when someone has selected a return log for viewing.

Because the link does not include the return submission to view, that fetch service has to work out which return submission is the latest, and therefore the one to display by default.

We also need the various submissions so we can use that data to generate the 'versions' section at the bottom of the page.

When it comes to download however, the version _must_ be specified. So, we should be able to obtain it in a single query, and not need to use the same logic the other fetch service is using.

Happy to pair on this if it does not make sense!

test/presenters/return-logs/download-return-log.presenter.test.js
```diff
@@ -0,0 +1,102 @@
...
+describe('Download Return Log presenter', () => {
+  let returnLog
+
+  beforeEach(() => {
+    returnLog = _testReturnLog()
+  })
+
+  describe('the "data" property', () => {
+    describe('when provided with a returnLog with a returnSubmission.method of "abstractionVolumes"', () => {
+      it('correctly formats the data to a csv string', () => {
```
Cruikshanks, line 27: The `it()` blocks don't actually tell us what the presenter is doing in each of these scenarios. It would be better if it gave a brief explanation of what the presenter does in each case.

test/services/return-logs/download-return-log.service.test.js
```diff
@@ -0,0 +1,88 @@
...
+  const returnLog = ReturnLogModel.fromJson({
+    id: 'v1:6:11/42/18.6.3/295:10055412:2022-11-01:2023-10-31',
+    returnReference: '10055412',
+    startDate: new Date('2022-11-01T00:00:00.000Z'),
+    endDate: new Date('2023-10-31T00:00:00.000Z'),
+    returnSubmissions: [returnSubmission]
+  })
+
+  return returnLog
+}
```
Cruikshanks, line 88: I'd suggest moving this into a test fixture so that it can be reused across these tests.

test/services/return-logs/fetch-download-return-log.service.test.js
```diff
@@ -0,0 +1,126 @@
...
+        }),
+        ReturnSubmissionLineHelper.add({
+          returnSubmissionId: returnSubmissions[0].id,
+          startDate: '2023-01-01',
+          quantity: 5
+        })
+      ])
+    })
+
+    describe('when a specified version is not selected', () => {
```
Cruikshanks, line 58: We shouldn't have a test for this scenario as it should not be possible.

test/services/return-logs/fetch-download-return-log.service.test.js
```diff
@@ -0,0 +1,126 @@
...
+    // We stub on the model prototype so that any created instances have $applyReadings stubbed. We don't set any return
+    // value as we don't need it to actually do anything; we just want to be able to assert that it was called.
+    Sinon.stub(ReturnSubmissionModel.prototype, '$applyReadings')
+  })
+
+  afterEach(() => {
+    Sinon.restore()
+  })
+
+  describe('when the return log with submissions exists', () => {
```
Cruikshanks, line 36: Again, we should only be calling this service because we have displayed the download button, and we've only displayed that because the return log has submissions.

So, I think you can just focus the 'narrative' in these tests to be 'when a matching return log exists / when no matching return log exists'.

Cruikshanks (changes requested):
Almost there, just a few things to consider. ❤️ 

My time should not be so pressured anymore. So, if you need to catch me about any of this, do not hesitate! 😁

test/services/return-logs/fetch-download-return-log.service.test.js
```diff
@@ -0,0 +1,91 @@
...
+        returnSubmissionId: returnSubmissions[0].id,
+        startDate: '2023-01-01',
+        quantity: 5
+      })
+    ])
+  })
+
+  afterEach(() => {
+    Sinon.restore()
+  })
```
Cruikshanks, line 53: Because these calls are in a `beforeEach()`, and we have 3 unit tests, it means we are creating the following every time this module is run.

- 3 return logs
- 9 return submissions
- 18 return submission lines

Whereas, if they were moved into a `before()` I reckon you could get away with just the following and still the tests would pass.

- 1 return log
- 3 return submissions
- 6 return submission lines

You may need both a `before()` and a `beforeEach()` because of the stubbing. But I reckon it's worth a go!

app/presenters/return-logs/download-return-log.presenter.js
```diff
@@ -0,0 +1,57 @@
...
+  const csvData = _transformToCsv(selectedReturnSubmission)
+  const filename = _formatFileName(returnLog)
+
+  return {
+    data: [HEADERS + '\n', ...csvData].join(''),
+    filename
+  }
+}
+
+function _formatFileName(returnLog) {
```
Cruikshanks, line 34:
```suggestion
function _fileName(returnLog) {
```

app/presenters/return-logs/download-return-log.presenter.js
```diff
@@ -0,0 +1,57 @@
...
+  const { returnReference, startDate, endDate, returnSubmissions } = returnLog
+  const version = returnSubmissions[0].version
+
+  return `${returnReference}_${formatDateObjectToISO(startDate)}_${formatDateObjectToISO(endDate)}_v${version}.csv`
+}
+
+function _transformToCsv(selectedReturnSubmission) {
```
Cruikshanks, line 41:
```suggestion
function _csvData(selectedReturnSubmission) {
```

Cruikshanks (changes requested):
Have a crack and re-jigging the test setup, and see if I'm right and we can reduce the number of queries we're firing and records we're creating.

Cruikshanks (approved):

PR #1737, opened 2025-02-18
Add GOV.UK notify services

jonathangoulding:

As part of the ongoing work to migrate the legacy applications logic into the system repo we have the requirement to handle notifications using GOV.UK Notify in the system application.

This change adds the initial notify services we will commonly need to use [email, letter].

We want to control the response object from notify and "lightly" take what we need. This is for two reasons:
- If the Notify api changes we have our own internal representation
- We (currently) only needs the status and errors from the response

We want to capture the errors from notify and pass them back to the caller to handle the errors (We do not want the server to stop if a notification fails).

The tests give us a good understanding of the responses we can expect from notify to allow the calling functions to handle appropriately.


test/services/notify/notify-email.service.test.js
```diff
@@ -0,0 +1,146 @@
...
+const { expect } = Code
+
+const config = require('../../../config/notify.config.js')
+
+const { NotifyClient } = require('notifications-node-client')
+
+const NotifyEmailService = require('../../../app/services/notify/notify-email.service.js')
+
+describe.only('Notify - Email service', () => {
+  const stubNotify = true // Used to perform integration tests with notify
```
jonathangoulding, line 18: Worth a discussion on having this in the test. 

I did initially create two test files (basically identical) one for the 'unit' test wit the stub. And one for an 'integration' test. 

With this toggle we can assert that out implementation with notify is working correctly. If / when we update the notify package,  i think it would be useful to run these manually locally.

config/notify.config.js
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+// We require dotenv directly in each config file to support unit tests that depend on this subset of config.
+// Requiring dotenv in multiple places has no effect on the app when running for real.
+require('dotenv').config()
+
+const config = {
+  apiKey: process.env.GOV_UK_NOTIFY_API_KEY,
+  template: {
+    returnsInvitationPrimaryUserEmail: process.env.NOTIFY_TEMPLATE_RETURNS_INVITATION_PRIMARY_USER_EMAIL
```
jonathangoulding, line 15: These are hard coded in the current system - https://github.com/DEFRA/water-abstraction-service/blob/main/config.js#L159

I am happy to put them here and not in the ENV. Just wanted to put this like this to raise the question.

app/services/notify/notify-email.service.js
```diff
@@ -0,0 +1,62 @@
...
+const NotifyClient = require('notifications-node-client').NotifyClient
+
+const config = require('../../../config/notify.config.js')
+
+/**
+ * Send an email using GOV.UK Notify
+ *
+ * https://docs.notifications.service.gov.uk/node.html#send-an-email
+ *
+ * Notify has the option to provide an 'options' object. This is the object which populates templates placeholders
```
jonathangoulding, line 17: See if this JS doc can be improved

app/services/notify/notify-email.service.js
```diff
@@ -0,0 +1,60 @@
...
+ */
+async function go(templateId, emailAddress, options) {
+  const notifyClient = new NotifyClient(config.apiKey)
+
+  return _sendEmail(notifyClient, templateId, emailAddress, options)
+}
+
+async function _sendEmail(notifyClient, templateId, emailAddress, options) {
+  try {
+    const response = await notifyClient.sendEmail(templateId, emailAddress, options)
```
jonathangoulding, line 44: Do we want to set some sort of predicate check on the options / emailAddress. 

I am thinking more for the 'sendLetter' which requires require at least 3 lines of address.

jonathangoulding: Maybe a determine notify service with it's own validators. 

Email is an email, letter has 3 `address_lines`. 

We know the templates and the expected so it could be as granular as by template id. But we trust our upstream services have shaped the data correctly. Just might be worth considering instead of sending a bad payload to Notify.

app/services/notify/notify-email.service.js
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+const NotifyClient = require('notifications-node-client').NotifyClient
+
+const config = require('../../../config/notify.config.js')
+
+/**
+ * Send an email using GOV.UK Notify
+ *
+ * https://docs.notifications.service.gov.uk/node.html#send-an-email
```
jonathangoulding, line 15: Email (and sms) can use the notify callback to let us know the status of the message. 

https://docs.notifications.service.gov.uk/node.html#set-up-callbacks

Do we want to use this and have a retry mechanism of our own.

test/services/notify/notify-email.service.test.js
jonathangoulding: Add this in as config

config/notify.config.js
jonathangoulding: If hard coded move to static

app/services/notify/notify-email.service.js
jonathangoulding: No need - notify will do a better job

app/services/notify/notify-email.service.js
jonathangoulding: Will need to be handled on the legacy UI

config/notify.config.js
```diff
@@ -0,0 +1,16 @@
...
+ * Config values used for GOV.UK Notify
+ * @module NotifyConfig
+ */
+
+// We require dotenv directly in each config file to support unit tests that depend on this subset of config.
+// Requiring dotenv in multiple places has no effect on the app when running for real.
+require('dotenv').config()
+
+const config = {
+  apiKey: process.env.GOV_UK_NOTIFY_API_KEY || 'test-key'
```
Cruikshanks, line 13: IMHO, if a config value is required, we should leave off the default.

I'm guessing this will break your tests, certainly in CI. In which case, you can set a fake API key in `.github/workflows/ci.yml`.

Also, we need to add `GOV_UK_NOTIFY_API_KEY` to `.env.example`.

test/services/notify/notify-email.service.test.js
```diff
@@ -0,0 +1,159 @@
...
+
+// Test helpers
+const { NotifyClient } = require('notifications-node-client')
+const { notifyTemplates } = require('../../../app/lib/notify-templates.lib.js')
+
+// Thing under test
+const NotifyEmailService = require('../../../app/services/notify/notify-email.service.js')
+
+describe('Notify - Email service', () => {
+  const stubNotify = process.env.STUB_NOTIFY || true // Used to perform integration tests with notify
```
Cruikshanks, line 19: I'd move this to `config/notify.config.js` so the config can all be put in one place (see `config/log.config.js` for an example of where we have previously put 'test-only' config).

Making this change also means _all_ config comes from the `config/` files, and folks only have to look in there to see what can be changed. We aim not to have any 'hidden' config values elsewhere because of direct references to `process.env`.

app/lib/notify-templates.lib.js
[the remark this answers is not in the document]
jonathangoulding: I did but i felt with this being specific to an external provider  it would be better suited in a separate file.

Cruikshanks (approved):

PR #1738, opened 2025-02-19
quarterly return logs historic changes

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4862

As part of our work to create quarterly returns we need to be able to handle historic changes on licences with quarterly returns. 

This PR adds in the logic to create only the quarterly returns that are valid for the return version dates.


Cruikshanks (changes requested):
Just a suggested change to make the filter check a little easier to follow.

app/services/return-logs/create-return-logs.service.js
```diff
@@ -49,7 +49,15 @@ function _generateReturnLogs(returnRequirement, returnCycle) {
...
     const quarterlyReturnPeriods = determineReturnsPeriods(returnCycle)
 
-    for (const quarterlyReturnPeriod of quarterlyReturnPeriods) {
+    const periodsToProcess = quarterlyReturnPeriods.filter((period) => {
+      return (
+        returnRequirement.returnVersion.startDate <= period.endDate &&
+        (returnRequirement.returnVersion.endDate >= period.startDate ||
+          returnRequirement.returnVersion.endDate === null)
+      )
+    })
```
Cruikshanks, line 58:
```suggestion
    const periodsToProcess = quarterlyReturnPeriods.filter((period) => {
      const startDateInPeriod = returnRequirement.returnVersion.startDate <= period.endDate
      const endDateInPeriod = returnRequirement.returnVersion.endDate >= period.startDate
      const endDateIsNull = returnRequirement.returnVersion.endDate === null

      return (startDateInPeriod && endDateInPeriod || endDateIsNull)
    })
```

Cruikshanks (approved):

PR #1740, opened 2025-02-20
Return log setup received page content changes

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4844

During testing for the above ticket, it was noted that the page prototype was updated after the work was done on the ticket. This PR is adding new content for the page (the dates for today and yesterday).


app/presenters/return-logs/setup/received.presenter.js
```diff
@@ -22,6 +24,9 @@ function go(session) {
     receivedDateYear
   } = session
 
+  const yesterdaysDate = new Date()
```
Jozzey, line 27: I'd move the creation of yesterday's date into its own private function, e.g.:
```
function _yesterdaysDate() {
  const yesterdaysDate = new Date()
  yesterdaysDate.setDate(yesterdaysDate.getDate() - 1)
  return formatLongDate(yesterdaysDate)
}
```

Jozzey (changes requested):

PR #1749, opened 2025-02-26
Spike Batch Notify

jonathangoulding:

Get the batch notifications working and dive out the design / structure of the code.

### DIFF
--- a/app/services/notify/batch-send-notifications.service.js
+++ b/app/services/notify/batch-send-notifications.service.js
@@ -1,15 +1,15 @@
 'use strict'
 
 /**
- * Send an email using GOV.UK Notify
+ * Batch send notifications to Notify
  * @module BatchSendNotificationsService
  */
 
 const NotifyEmailService = require('./notify-email.service.js')
 const NotifyLetterService = require('./notify-letter.service.js')
 const notifyRateLimit = 3000
-const { notifyTemplates } = require('../../lib/notify-templates.lib.js')
-const { contactName, contactAddress } = require('../../presenters/crm.presenter.js')
+const NotifyEmailPresenter = require('./notify-returns-email.presenter.js')
+const NotifyLetterPresenter = require('./notify-returns-letter.presenter.js')
 
 /**
  *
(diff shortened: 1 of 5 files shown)


app/services/notify/batch-send-notifications.service.js
```diff
@@ -0,0 +1,137 @@
...
+      },
+      reference: 'developer-testing'
+    }
+  }
+}
+
+async function _sendLetter(recipient) {
+  const data = _formatLetter(recipient)
+
+  return NotifyLetterService.go(data.templateId, data.options).then((res) => {
```
jonathangoulding, line 89: This seems necessary to allow the Notify functions to be a simple notify response and not worry about returning the personalisation and contact hash to the caller. 

We need the contact hash if there is an error to retry the notify calls - will need to this in the catch. ?

app/services/notify/notify-returns-email.presenter.js
```diff
@@ -0,0 +1,34 @@
...
+
+const { notifyTemplates } = require('../../lib/notify-templates.lib.js')
+
+/**
+ *
+ * @param recipient
+ * @param returnsPeriod
+ * @param referenceCode
+ */
+function go(recipient, returnsPeriod, referenceCode) {
```
jonathangoulding, line 16: presenters make sense to format the data for the notify calls. 

The question would be should we do it before each request is made. Or format all the recipients into notifications. 

I am leaning to changing all in the recipients in the array into a something like notifications. As we have a clear delineation between a recipient and a notification. 

As we are using the existing DB table with a personalisation column this make sense.

app/services/notify/batch-send-notifications.service.js
```diff
@@ -0,0 +1,101 @@
...
+async function _sendLetter(recipient, returnsPeriod, referenceCode) {
+  const formatedData = NotifyLetterPresenter.go(recipient, returnsPeriod, referenceCode)
+
+  return NotifyLetterService.go(formatedData.templateId, formatedData.options).then((notifyData) => {
+    return _scheduledNotification(notifyData, recipient, formatedData)
+  })
+}
+
+async function _sendEmail(recipient, returnsPeriod, referenceCode) {
+  const formatedData = NotifyEmailPresenter.go(recipient, returnsPeriod, referenceCode)
```
jonathangoulding, line 57: Doing this to the array prior to this should allow us to handle the errors / success simpler in the batch. 

If we can remove the need to use the contact hash then i think this is better. 

We get all the results resolved promises and we  can handle the errors and resend the 'payload' 

```javascript
  const failed = notifications.filter((response) => response.status !== 'fulfilled')
```

PR #1751, opened 2025-02-26
Point Edit Return button to Check Details page

StuAA78:
https://eaflood.atlassian.net/browse/WATER-4908

Currently the “Edit Return” button on the new “View Returns” page points to the legacy code. This PR points it to the new “Check details” page, which  involves taking the submission data and loading it into the session so that it’s correctly displayed in “Check details”.


Cruikshanks (changes requested):
Just noting this is back to you @StuAA78 for the changes discussed.

Saves me thinking I've forgotten to review something!

test/services/return-logs/setup/initiate-session.service.test.js
Cruikshanks, line 1: The only bit of feedback I have is on these tests.

The three states the service has to deal with are

- not received or submitted
- received but not submitted
- received and submitted

At the moment, that is not clear from the tests. IMHO, if we had these as our top-level scenarios, then tests like `describe('and a unit is specified', () => {` could be seen within the context of when this would be.

So, it's an opinion/suggestion rather than a request for change! 😁

Cruikshanks (approved):

PR #1753, opened 2025-02-27
Add Event persist service

jonathangoulding:

As part of the work to handle notifications in the system code. We need to follow some patterns established by the legacy code.

We use 'wabs.public.events' for handling notifications. This table is used to link notifications to 'wabs.public.scheduled_notifications'.

The event table is used on the 'notifications/report' page to show all the notifications.

We need to match the existing structure to allow the legacy code to still function as expected.

This change adds the initial service to create a new event suitable for a returns notification.


Cruikshanks (changes requested):
This is great. Only thing I'd ask to change is the prefix of the service.

In other places where we have added a service purely to handle persisting something to the DB, we've used `create-*.service.js`.

See `app/services/bill-runs/create-bill-run-event.service.js` for an example.

jonathangoulding:
> This is great. Only thing I'd ask to change is the prefix of the service.
> 
> In other places where we have added a service purely to handle persisting something to the DB, we've used `create-*.service.js`.
> 
> See `app/services/bill-runs/create-bill-run-event.service.js` for an example.

Updated

Cruikshanks (approved):

PR #1755, opened 2025-02-27
Feature notifications setup submit check

jonathangoulding:

As part of the work to migrate the returns notifications from the legacy code. We need to have a service to send notification to notify. 

This change adds the initial services to prepare the data for processing the notifications.


app/services/notifications/setup/submit-check.service.js
```diff
@@ -0,0 +1,28 @@
...
+ * This service will transform the recipients into notification and start processing notifications.
+ *
+ * @param {string} sessionId - The UUID for the notification setup session record
+ *
+ * @returns {Promise<object>} The view data for the licence page
+ */
+async function go(sessionId) {
+  const session = await SessionModel.query().findById(sessionId)
+
+  // Temp - need event service pr to be merged
```
jonathangoulding, line 22: need to add - https://github.com/DEFRA/water-abstraction-system/pull/1753

When merged

Cruikshanks (changes requested):
I guess it feels like things are at odds. The PR and the way the service is called by the controller makes it look like `SubmitCheckService` will be doing all the work.

However, when you look at the service, it seems to be determining who the recipients will be, and returning them as a result.

I'm fine either way, but I would ask that the PR description, the call from the controller (and what it assumes the service is doing) and the service itself all line up.

jonathangoulding:
> I guess it feels like things are at odds. The PR and the way the service is called by the controller makes it look like `SubmitCheckService` will be doing all the work.
> 
> However, when you look at the service, it seems to be determining who the recipients will be, and returning them as a result.
> 
> I'm fine either way, but I would ask that the PR description, the call from the controller (and what it assumes the service is doing) and the service itself all line up.

This was suppose to be the skeleton (route/ controller) before putting in the try/catch and calls to the next to be notification service.

The return arg was to get the tests in place. It's hard to add the test when there is nothing to actually test yet.

app/services/notifications/setup/submit-check.service.js
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Orchestrates handling the data for `/notifications/setup/{sessionId}/check` page
+ *
+ * This service will transform the recipients into notifications and start processing notifications.
+ *
+ * @param {string} sessionId - The UUID for the notification setup session record
+ *
+ * @returns {Promise<object>} The view data for the licence page
+ */
+async function go(sessionId) {
```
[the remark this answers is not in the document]
jonathangoulding: updated

app/services/notifications/setup/submit-check.service.js
```diff
@@ -0,0 +1,40 @@
...
+const { currentTimeInNanoseconds, calculateAndLogTimeTaken } = require('../../../lib/general.lib.js')
+
+/**
+ * Orchestrates handling the data for `/notifications/setup/{sessionId}/check` page
+ *
+ * This service will transform the recipients into notifications and start processing notifications.
+ *
+ * @param {string} sessionId - The UUID for the notification setup session record
+ *
+ * Does not return to allow the process to run in the background
```
Cruikshanks, line 20: If you want to keep this note, it should be bumped above the params within the description.

But I think there are enough examples of this pattern now that it is also fine to drop it 🤷

test/services/notifications/setup/submit-check.service.test.js
```diff
@@ -0,0 +1,55 @@
...
+const Sinon = require('sinon')
+
+const { describe, it, afterEach, before } = (exports.lab = Lab.script())
+const { expect } = Code
+
+// Test helpers
+const DetermineRecipientsService = require('../../../../app/services/notifications/setup/determine-recipients.service.js')
+const RecipientsFixture = require('../../../fixtures/recipients.fixtures.js')
+const RecipientsService = require('../../../../app/services/notifications/setup/fetch-recipients.service.js')
+const SessionHelper = require('../../../support/helpers/session.helper.js')
```
Cruikshanks, line 15:
```suggestion
// Test helpers
const RecipientsFixture = require('../../../fixtures/recipients.fixtures.js')
const SessionHelper = require('../../../support/helpers/session.helper.js')

// Things we need to stub
const DetermineRecipientsService = require('../../../../app/services/notifications/setup/determine-recipients.service.js')
const RecipientsService = require('../../../../app/services/notifications/setup/fetch-recipients.service.js')
```

test/services/notifications/setup/submit-check.service.test.js
Cruikshanks, line 1: The unit tests for services intended to run in the background also include checks to ensure that nothing is ever thrown, and 'noddy' checks that things are logged.

These are often good starting ones to include, especially when the service is just a stub and intended to be iterated over a few PRs.

Not required for this PR, but something like them would need to be in place to classify the service as 'finished'.

Cruikshanks (changes requested):
The required changes are just NITS.

But wanted to say, it is perfectly fine to have a 'shell' service, and therefore 'meaningless' unit tests, if it allows you to keep PRs small and focused like this.

Don't feel you have to change the design of the service purely to allow 'some' testing to take place.

Cruikshanks (approved):

PR #1758, opened 2025-02-28
Add notification setup event presenter

jonathangoulding:

As part of the work to handle notifications in the system code. We need to follow some patterns established by the legacy code.

This change adds the event presenter to structure the data to insert into 'wabs.public.events'.

This presenter will need to use existing legacy concepts such as setting a 'subtype' in order for the legacy reporting pages to continue to work.


test/presenters/notifications/setup/event.presenter.test.js
```diff
@@ -0,0 +1,179 @@
...
+
+  beforeEach(async () => {
+    recipients = RecipientsFixture.recipients()
+
+    session = {
+      returnsPeriod: 'quarterFour',
+      removeLicences: [],
+      journey: 'invitations',
+      referenceCode: 'RINV-123',
+      recipients: [...Object.values(recipients)],
```
jonathangoulding, line 28: changes will be added in another pr to provide this data to the session. 

The recipients (fetched and duplicates removed / contact_types marked as both etc) will be provided. 

And when a return cycle is selected we will determine the return cycle at one point and use that for the rest of the journey. We currently duplicate this logic in a few places.

app/presenters/notifications/setup/event.presenter.js
Cruikshanks, line 1: Sorry, another rename request.

Nearly all presenters you've dealt with so far have been related to getting things ready for display in the UI.

This, however, is about 'presenting' data for insertion into the DB. Though much fewer, we have some examples of this.

- `app/presenters/bill-runs/create-bill-run-event.presenter.js`
- `app/presenters/charging-module/create-customer-change.presenter.js`
- `app/presenters/charging-module/create-transaction.presenter.js`

Cutting to the chase, can we rename this to match the service, i.e. `create-event.presenter.js`?

app/presenters/notifications/setup/event.presenter.js
```diff
@@ -0,0 +1,106 @@
+'use strict'
+
+/**
+ * Formats the session data to create an event for 'wabs.public.events'
```
Cruikshanks, line 4: For reference, `wabs` is just what we've called the DB in our local environment. It's different in all of them so `wabs` might not be understood by those that follow.

Also, `public.events` is actually a view of the underlying table `water.events`, which is where the record will get persisted.

So, you might be better off referring to the model, and each model has a comment that references the underlying view and table.

 ```suggestion
 * Formats a notification `SessionModel` instance into the data needed for a 'EventModel' record
```

File: app/presenters/notifications/setup/event.presenter.js [hunk 3]
```diff
@@ -0,0 +1,106 @@
...
+ * @module EventPresenter
+ */
+
+const { transformStringOfLicencesToArray } = require('../../../lib/general.lib.js')
+
+/**
+ * Formats the session data to create an event for 'wabs.public.events'
```
Cruikshanks, line 11:
```suggestion
 * Formats a notification `SessionModel` instance into the data needed for a 'EventModel' record
```

app/presenters/notifications/setup/event.presenter.js
```diff
@@ -0,0 +1,106 @@
...
+    },
+    referenceCode,
+    status: 'started',
+    subtype: _subType(journey)
+  }
+}
+
+/**
+ * All the licences associated with an event (licences that will receive notifications) are stored in
+ * 'wabs.water.events.licences'. It is not clear where theses are used. But to be consistent we follow the established
```
Cruikshanks, line 38:
```suggestion
 * `water.events.licences`. It is not clear where theses are used. But to be consistent we follow the established
```

Cruikshanks (changes requested):

app/presenters/notifications/setup/event.presenter.js
jonathangoulding: Done

app/presenters/notifications/setup/event.presenter.js
jonathangoulding: done

Cruikshanks (approved):

PR #1762, opened 2025-02-28
Update how we use DetermineReturnsPeriodService

jonathangoulding:

We have multiple places using the 'DetermineReturnsPeriodService'. This was becoming complex to reuse and test (needed to stub the time in multiple places).

This change updates all the existing instances of DetermineReturnsPeriodService with a session object containing all the relevant information (session.determinedReturnsPeriod).

This change has simplified working with a chosen returns cycle.


app/services/notifications/setup/submit-returns-period.service.js
```diff
@@ -42,6 +43,13 @@ async function go(sessionId, payload) {
 async function _save(session, payload) {
   session.returnsPeriod = payload.returnsPeriod
 
+  const { returnsPeriod, summer } = DetermineReturnsPeriodService.go(session.returnsPeriod)
```
jonathangoulding, line 46: I will add the recipients in another change.

Cruikshanks (approved):
Very, _very_ ...

PR #1763, opened 2025-02-28
Add recipients to session on notification setup check

jonathangoulding:

This change adds the determined recipients to the session. This will simplify the submit controller (will be added in another change). There is no need to fetch the recipient again as we want this data 'as is'.

### DIFF
--- a/test/services/notifications/setup/check.service.test.js
+++ b/test/services/notifications/setup/check.service.test.js
@@ -5,7 +5,7 @@ const Lab = require('@hapi/lab')
 const Code = require('@hapi/code')
 const Sinon = require('sinon')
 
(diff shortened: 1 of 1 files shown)


jonathangoulding:
I am not convinced this is necessary now we have simplified the handling of returns cycles here - https://github.com/DEFRA/water-abstraction-system/pull/1762.

We can do the fetch in the submit controller simpler with the change above going in. It could come down to what is faster - the query or putting/receiving the recipients into the session (still a DB call or two). 

I would maybe lean to not doing this now as we don't want to reuse the recipients in the render so there would less benefit of adding it to the session. 

Will leave the decision up to you @Cruikshanks

jonathangoulding:
Will leave in draft until discussion. I have requested a review.

PR #1770, opened 2025-03-04
Update notifications setup

jonathangoulding:

Work has been done previously to implement the service and presenters. This was done to keep changes and code reviews concise.

A result of this has been some wrong assumptions i.e an array of licences stored as a jsonb filed needed to be converted to a json string prior to saving in the database.

This change starts stitching the service together to create an event and start the notification process (just the presenter, the other services will be developed later).

This change also rectifies some initial assumptions and makes sure our saved data matches the legacy schema.

### DIFF
--- a/test/services/notifications/setup/submit-check.service.test.js
+++ b/test/services/notifications/setup/submit-check.service.test.js
@@ -13,6 +13,7 @@ const RecipientsFixture = require('../../../fixtures/recipients.fixtures.js')
 const SessionHelper = require('../../../support/helpers/session.helper.js')
 
 // Things we need to stub
+const CreateEventService = require('../../../../app/services/notifications/setup/create-event.service.js')
 const DetermineRecipientsService = require('../../../../app/services/notifications/setup/determine-recipients.service.js')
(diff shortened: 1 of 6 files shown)


test/services/notifications/setup/submit-check.service.test.js
```diff
@@ -59,7 +64,33 @@ describe('Notifications Setup - Submit Check service', () => {
   it('correctly triggers the "DetermineRecipientsService"', async () => {
     await SubmitCheckService.go(session.id)
 
-    expect(DetermineRecipientsService.go.calledWith([recipients.primaryUser])).to.be.true()
+    expect(DetermineRecipientsService.go.calledWith(testRecipients)).to.be.true()
+  })
+
+  it('correctly triggers the "CreateEventService"', async () => {
```
jonathangoulding, line 70: I think this will be a temporary test whilst the rest of the code is developed. 

I think when the times is right there is a benefit to having this do the real calls to the DB and go through all the code (bar notify call).

Cruikshanks (approved):

PR #1787, opened 2025-03-10
Return log setup multiple entries page

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4811

As part of the return log setup journey work, a user can enter multiple daily/weekly/monthly readings or volumes. This PR is building that page.


app/validators/return-logs/setup/multiple-entries.validator.js
```diff
@@ -0,0 +1,81 @@
...
+  const filteredValues = value.filter((item) => {
+    return item !== null
+  })
+
+  for (let i = 0; i < filteredValues.length; i++) {
+    if (filteredValues[i] < startReading) {
+      return helpers.message(`The meter readings must be greater than or equal to the start reading of ${startReading}`)
+    }
+
+    if (filteredValues[i] < filteredValues[i - 1]) {
```
Cruikshanks, line 71: As you've filtered out all null values, would I be right in thinking we really need to check that the first reading is greater than or equal to the start reading?

It's after that we then need to compare each line to the previous one.

```suggestion
  if (filteredValues[0] < startReading) {
    return helpers.message(`The meter readings must be greater than or equal to the start reading of ${startReading}`)
  }

  for (let i = 1; i < filteredValues.length; i++) {
    if (filteredValues[i] < filteredValues[i - 1]) {
```

Cruikshanks (changes requested):
Ping me if anything doesn't make sense, I've got something wrong or you disagree.

app/validators/return-logs/setup/multiple-entries.validator.js
```diff
@@ -0,0 +1,81 @@
...
+  // all valid
+  const filteredValues = value.filter((item) => {
+    return item !== null
+  })
+
+  if (filteredValues[0] < startReading) {
+    return helpers.message(`The meter readings must be greater than or equal to the start reading of ${startReading}`)
+  }
+
+  for (let i = 0; i < filteredValues.length; i++) {
```
Cruikshanks, line 70: I know this doesn't break, but it also doesn't read write. On the first iteration we're testing that `filteredValues[0] < filteredValues[-1]`, i.e. we're checking the first value against undefined.

To avoid someone getting confused by this in the future, we should start the loop at 1.

```suggestion
  for (let i = 1; i < filteredValues.length; i++) {
```

Cruikshanks (changes requested):
Just one minor thing.

Cruikshanks (approved):

PR #1794, opened 2025-03-11
Fix charge versions 2pt flagging mechanism

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4951

Recent work has been completed to consolidate code where licences should be flagged for pre-sroc, sroc and two-part tariff supplementary billing. In the new flagging process a bug has been raised. The bug specifically deals with changes on a charge version. If a charge version was 2pt and is being changed to be a non-2pt charge version then the flagging logic is incorrect and the licence isn't being flagged when it should be. Equally, the flagging logic doesn't consider non-chargeable charge versions as well. This PR is to fix these issues.


Beckyrose200:
Just a note to say the changes on the PR for the check controller can be reverted back before this PR is up for review. The check endpoint was only updated to accommodate testing the change

PR #1795, opened 2025-03-11
Update notifications setup scheduled notifications

jonathangoulding:

 The legacy ui uses a column 'send_after' in 'water.scheduled_notification.send_after' to track when a notification is set.

 This change adds this timestamp so we know then a notification is created and sent.

 The legacy code has a mechanism that uses this column to determine when to check the status of the notifications. We will turn this off in another change in the water-abstraction-service repo.

### DIFF
--- a/app/presenters/notifications/setup/scheduled-notifications.presenter.js
+++ b/app/presenters/notifications/setup/scheduled-notifications.presenter.js
@@ -94,7 +94,8 @@ function _email(recipient, returnsPeriod, referenceCode, journey, eventId) {
     },
     recipient: recipient.email,
     reference: referenceCode,
-    sendAfter: _sendAfter(),
+    // NOTE: The legacy code uses the 'sendAfter' to know when a notification is sent.
+    sendAfter: timestampForPostgres(),
     templateId
   }
 }
--- a/app/presenters/notifications/setup/scheduled-notifications.presenter.js
+++ b/app/presenters/notifications/setup/scheduled-notifications.presenter.js
@@ -8,7 +8,7 @@
 const { contactName, contactAddress } = require('../../crm.presenter.js')
 const { formatLongDate } = require('../../base.presenter.js')
 const { notifyTemplates } = require('../../../lib/notify-templates.lib.js')
-const { transformStringOfLicencesToArray } = require('../../../lib/general.lib.js')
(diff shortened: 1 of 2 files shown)


app/presenters/notifications/setup/scheduled-notifications.presenter.js
```diff
@@ -94,6 +94,7 @@ function _email(recipient, returnsPeriod, referenceCode, journey, eventId) {
     },
     recipient: recipient.email,
     reference: referenceCode,
+    sendAfter: _sendAfter(),
```
Cruikshanks, line 97:
```suggestion
    // NOTE: The legacy code uses the 'sendAfter' to know when a notification is sent.
    sendAfter: timestampForPostgres(),
```

app/presenters/notifications/setup/scheduled-notifications.presenter.js
```diff
@@ -152,6 +153,7 @@ function _letter(recipient, returnsPeriod, referenceCode, journey, eventId) {
       ..._returnsPeriod(returnsPeriod)
     },
     reference: referenceCode,
+    sendAfter: _sendAfter(),
```
Cruikshanks, line 156:
```suggestion
    // NOTE: The legacy code uses the 'sendAfter' to know when a notification is sent.
    sendAfter: timestampForPostgres(),
```

app/presenters/notifications/setup/scheduled-notifications.presenter.js
```diff
@@ -231,6 +233,15 @@ function _messageRef(journey, messageType, contactType) {
...
 
+/**
+ * The legacy code uses the 'sendAfter' to know when a notification is sent.
+ *
+ * @private
+ */
+function _sendAfter() {
+  return timestampForPostgres()
+}
+
```
Cruikshanks, line 244:
```suggestion
```

Cruikshanks (changes requested):
IMHO, I reckon you can just call the function directly. I', also fine if you want to drop the comments, but I've left them in.

But if we do I'm not sure they are correct.

My understanding of `send_after` was that it was intended to give them the flexibility to 'queue' the sending of notifications. For example, send these 4000 reminders after 9pm, when no one is using the service.

But I think for most notifications it just sets the value to the current time, which the logic in water-abstraction-service interprets as 'send immediately'.

Cruikshanks (approved):

PR #1796, opened 2025-03-11
Add notifications setup scheduled notifications update notify

jonathangoulding:

When a scheduled notification has been sent to Notify we need to capture the Notify response.

The legacy code stores parts of the notify response in 'water.scheduled_notification'.

This change replicates Notify data used in the legacy codes .

When successful we will receive a notify id to check the future status of the notification with notify.

When there is an error we capture the 'raw' error and store in the scheduled notification.


app/presenters/notifications/setup/scheduled-notification-update-notify.presenter.js
```diff
@@ -0,0 +1,49 @@
+'use strict'
+
+/**
+ * Updates a scheduled notification to use the notify response data
+ * @module ScheduledNotificationsUpdateNotifyPresenter
```
jonathangoulding, line 5: I am stuck on how to name this. 

This will be called after we have the notify response.

app/presenters/notifications/setup/scheduled-notification-update-notify.presenter.js
```diff
@@ -0,0 +1,49 @@
...
+ * capture the response. When the request is successful we need to store the notify id to make future check on its
+ * status. When a request fails we will not receive a notify id and will need to store the error in the same pattern as
+ * the legacy code.
+ *
+ * @param {object} scheduledNotification
+ * @param {object } notify - the response from notify
+ *
+ * @returns {object} - a scheduled notification updated with the notify response
+ */
+function go(scheduledNotification, notify) {
```
Cruikshanks, line 23: Does it need to be passed in the `scheduledNotification`, only to then just `...scheduledNotification` in the result?

I'd argue the purpose of this presenter becomes more specific and clearer, plus, it solves the naming problem: `notify-update.presenter.js`.

app/presenters/notifications/setup/notify-update.presenter.js
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @param {object } notify - the response from notify
+ *
+ * @returns {object} - the Notify response formatted to save as part of a 'water.scheduled_notifications'
+ */
+function go(notify) {
+  if (notify.status !== CREATED) {
+    return {
+      ..._error(notify)
+    }
```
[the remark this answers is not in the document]
jonathangoulding: Yeah my bad

PR #1803, opened 2025-03-12
Update notifications setup batch notifications to batch by amount

jonathangoulding:

The Notify service imposes a rate limit of 3,000 requests per minute and restricts the number of statuses returned per API call to a maximum of 250. Given these constraints, this update introduces a batch processing mechanism to handle the lowest of the two limitations, which in this case is the 250-message status retrieval limit from Notify.

Thus, the system will process a maximum of 250 recipients per batch.

To allow for flexibility and future adjustments, we’ve added a configurable setting that enables us to modify the batch size if necessary. This also aids in testing the solution with smaller batches for validation purposes.

Additionally, this approach helps us stay within the constraints of PostgreSQL’s batch insert limit. While PostgreSQL supports larger batches, the Notify API’s limitation on status checks (250 per call) remains the primary constraint. Therefore, the Notify limit is the driving factor for determining the batch size.factor.


app/services/notifications/setup/batch-notifications.service.js
```diff
@@ -34,6 +35,27 @@ const CreateNotificationsService = require('./create-notifications.service.js')
  * @returns {object} - the number of sent and errored notifications
  */
 async function go(recipients, determinedReturnsPeriod, referenceCode, journey, eventId) {
+  let sent = 0
+  let error = 0
+
+  const size = batchSize
```
jonathangoulding, line 41: I will work on the load limitation / stalling in a separate change.

config/notify.config.js
```diff
@@ -5,13 +5,16 @@
...
+const defaultBatchSize = 250
+
 // We require dotenv directly in each config file to support unit tests that depend on this subset of config.
 // Requiring dotenv in multiple places has no effect on the app when running for real.
 require('dotenv').config()
 
 const config = {
   apiKey: process.env.GOV_UK_NOTIFY_API_KEY,
-  stubNotify: process.env.STUB_NOTIFY || true // Used to perform integration tests with notify
+  batchSize: process.env.NOTIFICATIONS_BATCH_SIZE || defaultBatchSize,
```
Cruikshanks, line 16: I'm fine to put this inline. Where we have defaults in other places in the config, we do this.

```suggestion
  batchSize: process.env.NOTIFICATIONS_BATCH_SIZE || 250,
```

Else, if you want to keep the named param we should change the param to reflect it is a constant; `DEFAULT_BATCH_SIZE`.

I also think it would be helpful to have add a `// NOTE: ` above the line to copy the bit from your explanation on where 250 comes from.

test/services/notifications/setup/batch-notifications.service.test.js
```diff
@@ -32,6 +33,10 @@ describe('Notifications Setup - Batch notifications service', () => {
   let testRecipients
 
   beforeEach(async () => {
+    // By setting the batch size to 1 we can prove that all the batches are run, as we should have all the scheduled
+    // notifications still saved in the database regardless of batch size
```
Cruikshanks, line 37: ❤️

Cruikshanks (approved):

config/notify.config.js
jonathangoulding: Sonar complained about the 'magic number' hence it being moved up.

jonathangoulding: inline added and comment added based on commit message with link provided to docs.

Cruikshanks (approved):
Mean it this time!

PR #1808, opened 2025-03-13
Update notification setup batch notifications service to have a delay

jonathangoulding:

The Notify service imposes a rate limit of 3,000 requests per minute and restricts the number of statuses returned per API call to a maximum of 250. Given these constraints, this update introduces a batch processing delay mechanism to keep out request below the rate limit.

The system will process a maximum of 1506 notifications per minute.

To allow for flexibility and future adjustments, we’ve added a configurable setting that enables us to modify the batch delay if necessary. This also aids in testing the solution with shorter batch delays for validation purposes.

We are limiting ourselves to 1,500 requests per minute. Given that each status request returns up to 250 statuses, this translates to a maximum of 6 batches per minute (1,500 requests / 250 statuses per request = 6 batches).

To ensure we stay within this 6-batch-per-minute constraint, we introduce a **10-second delay** between each batch. This delay is calculated as: 60 seconds (1 minute) / 6 batches = 10 seconds per batch.

This delay ensures we don’t exceed 1,500 requests per minute, leaving enough room for other parts of the system to make requests to the service.

The call to get the statues will count towards the rate limit. But to keep the explanation simple it has been ignored. In reality the total calls per minute will be 1506 (6 additional calls to get the statues for each batch).


app/services/notifications/setup/batch-notifications.service.js
```diff
@@ -6,11 +6,12 @@
  */
 
 const CreateNotificationsService = require('./create-notifications.service.js')
+const NotifyConfig = require('../../../../config/notify.config.js')
 const NotifyEmailService = require('../../notify/notify-email.service.js')
 const NotifyLetterService = require('../../notify/notify-letter.service.js')
 const NotifyUpdatePresenter = require('../../../presenters/notifications/setup/notify-update.presenter.js')
 const ScheduledNotificationsPresenter = require('../../../presenters/notifications/setup/scheduled-notifications.presenter.js')
-const { batchSize } = require('../../../../config/notify.config.js')
+const { setTimeout } = require('node:timers/promises')
```
Cruikshanks, line 14: As it's an external package/library it needs to go above our own ([Top of .js files](https://github.com/DEFRA/water-abstraction-team/blob/main/coding_conventions.md#top-of-js-files))

app/services/notifications/setup/batch-notifications.service.js
```diff
@@ -6,11 +6,12 @@
  */
 
 const CreateNotificationsService = require('./create-notifications.service.js')
+const NotifyConfig = require('../../../../config/notify.config.js')
```
Cruikshanks, line 9: As it's config it needs to go below our modules ([Top of .js files](https://github.com/DEFRA/water-abstraction-team/blob/main/coding_conventions.md#top-of-js-files))

test/services/notifications/setup/batch-notifications.service.test.js
```diff
@@ -37,6 +36,10 @@ describe('Notifications Setup - Batch notifications service', () => {
     // notifications still saved in the database regardless of batch size
     NotifyConfig.batchSize = 1
 
+    // By setting the delay to 100ms we can keep the tests fast whilst assuring our batch mechanism is delaying
+    // correctly, we do not want increase the timeout for the test as we want them to fail if a timeout occurs
+    NotifyConfig.delay = ONE_HUNDRED_MILLISECONDS
```
Cruikshanks, line 41: Apologies, this applies to the `batchSize` as well (didn't clock it on first pass).

We normally stub changes to config. See `test/presenters/bill-runs/review/base-review.presenter.test.js` for an example.

```javascript
// Things we need to stub
const NotifyConfig = require('../../../../config/notify.config.js')

describe('Awesome service', () => {
  beforeEach(() => {
    Sinon.stub(NotifyConfig, 'batchSize').value(1)
    Sinon.stub(NotifyConfig, 'delay').value(ONE_HUNDRED_MILLISECONDS)
  })

  afterEach(() => {
    Sinon.restore()
  })
})

Cruikshanks (changes requested):
Just NITS!

Cruikshanks (approved):

PR #1811, opened 2025-03-14
Add Notify status service

jonathangoulding:

This change adds the 'NotifyStatusService' which checks the status of a single notification in notify.

Similar to previous Notify services it's tests can be used against the live GOV.UK Notify.

However, to do so would require a valid notification id stored with notify. As this would require the test to use one of our other Notify services (within the test itself) to create the notification it has been decided to have this as a manual step for the time being.

### DIFF
--- a/test/services/notify/notify-status.service.test.js
+++ b/test/services/notify/notify-status.service.test.js
@@ -1,0 +1,95 @@
+'use strict'
+
+// Test framework dependencies
+const Lab = require('@hapi/lab')
+const Code = require('@hapi/code')
+const Sinon = require('sinon')
+
+const { describe, it, afterEach, beforeEach } = (exports.lab = Lab.script())
+const { expect } = Code
+
+// Test helpers
+const { NotifyClient } = require('notifications-node-client')
+const { stubNotify } = require('../../../config/notify.config.js')
+
+// Thing under test
+const NotifyStatusService = require('../../../app/services/notify/notify-status.service.js')
+
+describe('Notify - Status service', () => {
+  let notifyStub
+  let notificationId
+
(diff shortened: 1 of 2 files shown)


jonathangoulding:
There was a previous assumption we could use the notify client to get multiple statuses at once - https://docs.notifications.service.gov.uk/node.html#get-the-status-of-multiple-messages. 

However, this assumption was wrong as we can not provide multiple notification id's to this endpoint. 

This will mean we need to address the assumed 250 batch limit. These changes will be made in a separate pr and will require some discussion. 

I have intentionally left this out of the PR commit as i don't think it add's anything necessary to this specific change. 

My initial thought is to have the batches decrease to 100 and update the delay to give us a similar total batch within a minute. 

We would create all the notifications in one go (without getting the statuses) and then have a separate status service ( called after BatchNotificationsService - with the same limitations on batch / delay) which can focus on updating the statuses. 

This service can then be used via a cron job / direct call to check for updates with notify (letters can take 72 hours to process).

test/services/notify/notify-status.service.test.js
```diff
@@ -0,0 +1,95 @@
...
+  describe('when the call to "notify" is successful', () => {
+    beforeEach(() => {
+      notifyStub = _stubSuccessfulNotify()
+    })
+
+    it('should call notify', async () => {
+      const result = await NotifyStatusService.go(notificationId)
+
+      expect(result).to.equal({
+        status: 'received'
```
jonathangoulding, line 40: This is a letter status. Should we highlight this or have a email and letter test.

Cruikshanks (approved):

PR #1822, opened 2025-03-18
Notifications - View sent notices page

robertparkinson:

As part of our work to replace the legacy UI we're slowly porting over pages to the new system. This PR migrates the notifications tab.

screencapture-localhost-8008-system-notices-2025-05-21-17_05_18

screencapture-localhost-8008-system-notices-2025-05-21-17_08_26


app/services/notifications/index.service.js
```diff
@@ -0,0 +1,136 @@
...
+  const _filterNotificationTypes = filters.filterNotificationTypes
+    ? _prepareNotifications(filterNotificationTypes)
+    : filterNotificationTypes
+
+  const validateResult = _validate(filters)
+
+  const filter = {
+    notifications: _filterNotificationTypes,
+    ...filters
+  }
```
Cruikshanks, line 33: TBH, I had to play with this section a bit before I understood what it was doing.

I suggest moving this into a function so we can give it a name to help our future selves remember what is going on! Perhaps, `_determineSelectedNotificationTypes()` and `_prepareNotifications()` becomes `_convertTypesToBooleanFlags()`? 🤷

app/services/notifications/index.service.js
```diff
@@ -0,0 +1,136 @@
...
+  const sentByError = validation.error.details.find((detail) => {
+    return detail.context.key === 'sentBy'
+  })
+
+  return {
+    text: 'There was a problem with your filters.',
+    ...(fromDateError && { fromFullDate: { message: fromDateError.message } }),
+    ...(toDateError && { toFullDate: { message: toDateError.message } }),
+    ...(sentByError && { sentBy: { message: sentByError.message } })
+  }
```
Cruikshanks, line 131: Could we not just iterate details and add the errors found? This might also solve the problem I spotted: if you add a to date before a from date, it correctly validates there is an error, but this current logic misses displaying that in the summary.

Screenshot 2025-04-13 at 11 41 45

Cruikshanks: In fact, validation is going to need a bit more work. Spotted now we are just showing the message `There was a problem with your filters` when we should be showing the specific error messages from Joi. And those errors should when clicked take the user to the problem field.

[GOV.UK Error summary](https://design-system.service.gov.uk/components/error-summary/)

I think we can also avoid most of the prep work in the validator itself, and just validate the `filters` object directly.

Cruikshanks (changes requested):
My time is up and so is my data, so I'm going to have to leave it there.

If I can I'll try and add some feedback, else we can pick this up when I return.

Cruikshanks (dismissed):

Cruikshanks (approved):

PR #1824, opened 2025-03-18
Update notification setup default event status

jonathangoulding:

There was a previous assumption of using setting the event status to 'started' when it is created. This resulted in the event not showing on the legacy on the 'notifications/report' page.

This change updates the default status to 'completed'. This will not change in our code base.

This lead to exploring where the 'sent' and 'error' values on the options object are used in the legacy code. They are not used as far as we can tell.

The recipients count displayed on the 'notifications/report' page is taken from 'metadata.recipients'.

The errors are 'counted' based on a query in the legacy code and does not use the 'metadata.error' key.

### DIFF
--- a/app/presenters/notifications/setup/create-event.presenter.js
+++ b/app/presenters/notifications/setup/create-event.presenter.js
@@ -11,6 +11,9 @@ const { transformStringOfLicencesToArray } = require('../../../lib/general.lib.j
 /**
  * Formats a notification `SessionModel` instance into the data needed for a 'EventModel' record
  *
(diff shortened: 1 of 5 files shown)


jonathangoulding:
> Double-check with @robertparkinson about what is or isn't being used.
> 
> For example, the legacy code might have opted to query the notifications to get an error count. But we're rebuilding that page, and I'm fairly certain we couldn't see the point of querying the notifications table when everything we need is in the event. 🤷

They have done a query to get the updated errors.

Cruikshanks (approved):

PR #1829, opened 2025-03-19
Update notifications setup submit ad-hoc licence

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4962

The 'ad-hoc' journey does not exist in the legacy code.

This change updates the notification setup submit ad-hoc- licence service to provide a 'determinedReturnsPeriod' in the session.

This is required to allow the 'ad-hoc' journey to use the existing 'reminders' GOV.UK Notify templates to send the notifications.

A previous change has implemented the ability to send notifications for ad-hoc licences -  https://github.com/DEFRA/water-abstraction-system/pull/1828.


app/services/notifications/setup/submit-ad-hoc-licence.service.js
```diff
@@ -45,6 +46,34 @@ async function go(sessionId, payload) {
   return {}
 }
 
+/**
+ * Ad-hoc licence notifications has highlighted an issue with using the returns period dates.
+ *
+ * This code is intended as a temporary 'fix' to allow ad-hoc returns to use the returns reminders Notify template
```
Cruikshanks, line 52:
```suggestion
 * This code is intended as a temporary 'fix' to allow ad-hoc returns to use the returns invitations Notify template
```

app/services/notifications/setup/submit-ad-hoc-licence.service.js
```diff
@@ -45,6 +46,34 @@ async function go(sessionId, payload) {
...
+ * (which requires the 'determinedReturnsPeriod' dates)
+ *
+ * This function alters the 'dueDate' but adding 28 days (this may not be the final state and should be regarded as a
```
Cruikshanks, line 55:
```suggestion
 * This function alters the 'dueDate' by adding 28 days to the current date (this may not be the final state and should be regarded as a
```

app/services/notifications/setup/submit-ad-hoc-licence.service.js
```diff
@@ -45,6 +46,34 @@ async function go(sessionId, payload) {
...
+ * placeholder).
+ *
+ * This also, manually, sets gets the 'startDate' and 'endDate' for 'allYear' which represents the current financial
+ * kyear.
```
Cruikshanks, line 59: Apologies, but I don't understand what this comment refers to (plus typos!)

Cruikshanks (changes requested):

app/services/notifications/setup/submit-ad-hoc-licence.service.js
jonathangoulding: The current financial year is calculated by the 'allYear'. So the start and end are used from that returns period. 

No we know theses are going i can temporarily set them to the same as due date and remove this part ?

app/services/notifications/setup/submit-ad-hoc-licence.service.js
```diff
@@ -45,6 +46,34 @@ async function go(sessionId, payload) {
...
+  const dueDate = new Date()
+  const daysToAdd = 28
+  dueDate.setDate(dueDate.getDate() + daysToAdd)
+
+  return {
+    ...returnsPeriod,
+    summer,
+    dueDate
+  }
+}
```
Cruikshanks, line 75: Perhaps...?

```suggestion
  const dueDate = new Date()
  
  dueDate.setDate(dueDate.getDate() + 28)
  
  // TODO: Remove this use of DetermineReturnsPeriodService(). Ad-hoc will use the same Notify template as invitations,
  // but currently it expects period start and end date values to be provided. We don't have these on the ad-hoc journey
  // so for now are using whatever the current return period is. Once we have confirmation the template has been updated
  // we can drop this call. 
  const returnsPeriod = DetermineReturnsPeriodService.go('allYear')

  return {
    dueDate,
    ...returnsPeriod
  }
}
```

app/services/notifications/setup/submit-ad-hoc-licence.service.js
Cruikshanks: Made a suggested change to the function, which means I think we can drop this comment altogether.

app/services/notifications/setup/submit-ad-hoc-licence.service.js
jonathangoulding: This will still need summer for the return as it will be set to false to not affect the legacy code queries.

Cruikshanks (approved):

PR #1831, opened 2025-03-20
Return "not due yet" status

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4954
https://eaflood.atlassian.net/browse/WATER-4968

There is a need for a status "Not due yet" for a return that is visible to the user so it is clear to them which return is due for submission. This status should be shown for those returns (Quarterly and Annual) that are not due. This PR is adding that status to the view returns page.

### DIFF
--- a/app/presenters/licences/view-licence-returns.presenter.js
+++ b/app/presenters/licences/view-licence-returns.presenter.js
@@ -98,27 +98,26 @@ function _status(returnLog) {
   // 'today' would be flagged as overdue when it is still due (just!)
   today.setHours(0, 0, 0, 0)
 
-  // For all status other than due we can just return the status and the return-status-tag macro will know how to
-  // display it
-  if (status !== 'due') {
-    return status
+  if (status === 'due') {
+    if (today > dueDate) {
+      return 'overdue'
+    }
+
+    // A return is considered "due" for 28 days, starting 28 days before the due date
+    // Any date before this period should be marked as "not due yet"
+    const notDueUntil = new Date(dueDate)
+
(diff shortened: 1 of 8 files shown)


app/presenters/licences/view-licence-returns.presenter.js
```diff
@@ -96,8 +98,22 @@ function _status(returnLog) {
   // 'today' would be flagged as overdue when it is still due (just!)
   today.setHours(0, 0, 0, 0)
 
-  if (status === 'due' && dueDate < today) {
-    return 'overdue'
+  if (status === 'due') {
```
StuAA78, line 101: We could invert this condition to make it a guard clause so we don't need to have nested ifs, ie:

```js
if (status !== 'due') {
  return status
}
```

app/presenters/licences/view-licence-returns.presenter.js
```diff
@@ -96,8 +98,22 @@ function _status(returnLog) {
...
+    if (dueDate < today) {
```
StuAA78, line 102: If we swap this to be `if (today > dueDate) { ... }` then it's in line with the comparison we make on line 114 and (for me anyway) the logic flow is a little clearer.

StuAA78 (changes requested):

app/presenters/licences/view-licence-returns.presenter.js
Beckyrose200: I want to sometimes return the status when it's due as well so this change doesn't remove the need to still return status at the end

StuAA78 (approved):

PR #1834, opened 2025-03-20
Add notifications setup update notifications

jonathangoulding:

When a notification has been sent to GOV.UK Notify we need to check the status of the notification. We receive various statuses from Notify.

This change introduces an 'UpdateNotificationsService'. This service will update the 'status' and 'notifyStatus' for a 'scheduledNotification'.

### DIFF
--- a/app/services/notifications/setup/update-notifications.service.js
+++ b/app/services/notifications/setup/update-notifications.service.js
@@ -14,12 +14,32 @@ const ScheduledNotificationModel = require('../../../../app/models/scheduled-not
  * https://www.postgresql.org/docs/current/postgres-fdw.html#POSTGRES-FDW-OPTIONS-REMOTE-EXECUTION. The calling
  * service will handle the batch process.
  *
+ * We have used an insert with onConflict to allow us to provide an array of notifications to reduce the calls needed to
+ * the database.
+ *
+ * ```javascript
+ * [
+ *   {
+ *     notifyId: '1234',
+ *     notifyStatus: 'received',
+ *     status: 'sent'
+ *   },
+ *   {
+ *     notifyId: '123',
+ *     status: 'sent'
+ *   }
+ * ]
+ * ```
+ *
  * @param {object} notifications
  *
  * @returns {object} - the Updated notifications
  */
 async function go(notifications) {
-  return ScheduledNotificationModel.query().insert(notifications).onConflict('id').merge(['status', 'notifyStatus'])
+  return ScheduledNotificationModel.query()
+    .insert(notifications)
+    .onConflict('notifyId')
+    .merge(['status', 'notifyStatus'])
 }
 
 module.exports = {
--- a/app/services/notifications/setup/update-notifications.service.js
+++ b/app/services/notifications/setup/update-notifications.service.js
@@ -16,10 +16,9 @@ const ScheduledNotificationModel = require('../../../../app/models/scheduled-not
  *
  * @param {object} notifications
  *
- * @returns {object} - the Updated notifications
  */
 async function go(notifications) {
-  return ScheduledNotificationModel.query().insert(notifications).onConflict('id').merge(['status', 'notifyStatus'])
+  await ScheduledNotificationModel.query().insert(notifications).onConflict('id').merge(['status', 'notifyStatus'])
 }
 
 module.exports = {
--- a/app/services/notifications/setup/update-notifications.service.js
+++ b/app/services/notifications/setup/update-notifications.service.js
@@ -14,32 +14,12 @@ const ScheduledNotificationModel = require('../../../../app/models/scheduled-not
  * https://www.postgresql.org/docs/current/postgres-fdw.html#POSTGRES-FDW-OPTIONS-REMOTE-EXECUTION. The calling
  * service will handle the batch process.
  *
- * We have used an insert with onConflict to allow us to provide an array of notifications to reduce the calls needed to
- * the database.
- *
- * ```javascript
- * [
- *   {
- *     notifyId: '1234',
- *     notifyStatus: 'received',
- *     status: 'sent'
- *   },
- *   {
- *     notifyId: '123',
- *     status: 'sent'
- *   }
- * ]
- * ```
- *
  * @param {object} notifications
  *
  * @returns {object} - the Updated notifications
  */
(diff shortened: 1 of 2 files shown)


test/services/notifications/setup/update-notifications.service.tests.js
```diff
@@ -0,0 +1,163 @@
...
+// Test helpers
+const EventHelper = require('../../../support/helpers/event.helper.js')
+const ScheduledNotificationHelper = require('../../../support/helpers/scheduled-notification.helper.js')
+const ScheduledNotificationModel = require('../../../../app/models/scheduled-notification.model.js')
+const { timestampForPostgres } = require('../../../../app/lib/general.lib.js')
+
+// Thing under test
+const UpdateNotificationsService = require('../../../../app/services/notifications/setup/update-notifications.service.js')
+
+describe('Notifications Setup - Update notification service', () => {
```
Cruikshanks, line 19:
```suggestion
describe('Notifications Setup - Update Notifications service', () => {
```

test/services/notifications/setup/update-notifications.service.tests.js
```diff
@@ -0,0 +1,163 @@
...
+const Code = require('@hapi/code')
+
+const { describe, it, beforeEach } = (exports.lab = Lab.script())
+const { expect } = Code
+
+// Test helpers
+const EventHelper = require('../../../support/helpers/event.helper.js')
+const ScheduledNotificationHelper = require('../../../support/helpers/scheduled-notification.helper.js')
+const ScheduledNotificationModel = require('../../../../app/models/scheduled-notification.model.js')
+const { timestampForPostgres } = require('../../../../app/lib/general.lib.js')
```
Cruikshanks, line 14:
```suggestion
const EventHelper = require('../../../support/helpers/event.helper.js')
const { timestampForPostgres } = require('../../../../app/lib/general.lib.js')
const ScheduledNotificationHelper = require('../../../support/helpers/scheduled-notification.helper.js')
const ScheduledNotificationModel = require('../../../../app/models/scheduled-notification.model.js')
```

app/services/notifications/setup/update-notifications.service.js
```diff
@@ -0,0 +1,27 @@
...
+ * This query uses postgresql's ability to insert with batches. There is a limitation on the batch size
+ * https://www.postgresql.org/docs/current/postgres-fdw.html#POSTGRES-FDW-OPTIONS-REMOTE-EXECUTION. The calling
+ * service will handle the batch process.
+ *
+ * @param {object} notifications
+ *
+ * @returns {object} - the Updated notifications
+ */
+async function go(notifications) {
+  return ScheduledNotificationModel.query().insert(notifications).onConflict('id').merge(['status', 'notifyStatus'])
```
Cruikshanks, line 22: This is just a question.

Will we be using the scheduled notifications returned from this service somewhere else? If not, this should just be `await ScheduledNotificationModel`.

test/services/notifications/setup/update-notifications.service.tests.js
```diff
@@ -0,0 +1,163 @@
...
+        {
+          createdAt: new Date(scheduledNotification.createdAt),
+          id: scheduledNotification.id,
+          notifyStatus: 'received',
+          status: 'sent'
+        }
+      ]
+    })
+
+    it('should update a single notification (and only update the required values)', async () => {
```
Cruikshanks, line 52:
```suggestion
    it("updates only the notification's required values", async () => {
```

test/services/notifications/setup/update-notifications.service.tests.js
```diff
@@ -0,0 +1,163 @@
...
+        },
+        {
+          createdAt: new Date(scheduledNotification2.createdAt),
+          id: scheduledNotification2.id,
+          status: 'sent'
+        }
+      ]
+    })
+
+    it('should return the saved notifications', async () => {
```
Cruikshanks, line 108:
```suggestion
    it('updates the notifications required values', async () => {
```

Cruikshanks (changes requested):
Some NITS and a question

app/services/notifications/setup/update-notifications.service.js
jonathangoulding: correct it should be await i will update

jonathangoulding:
I will have to rethink this as we would want to update based on the notify id to save mapping notify responses to our scheduled notifications (it would add unnecessary complexity)

jonathangoulding:
Marked as draft to update the 'notifyId' to be unique

test/services/notifications/setup/update-notifications.service.tests.js
```diff
@@ -0,0 +1,163 @@
...
+        }
+      ]
+    })
+
+    it("updates only the notification's required values", async () => {
+      await UpdateNotificationsService.go(notifications)
+
+      const createdResult = await ScheduledNotificationModel.query().where('eventId', eventId)
+
+      expect(createdResult).equal([
```
Cruikshanks, line 57: Two things. We're updating not creating so `createdResult` doesn't read right in this context. To avoid the confusion altogether, just stick to our convention of `result` or `results` (depending on whether one or more results are being returned) in tests.

Secondly, I think you can make it even more verbose that your scheduled notification got updated by using a different syntax to get it.

```suggestion
      const result = await scheduledNotification.$query()

      expect(result).equal(
```

test/services/notifications/setup/update-notifications.service.tests.js
```diff
@@ -0,0 +1,163 @@
...
+      ]
+    })
+
+    it('updates the notifications required values', async () => {
+      await UpdateNotificationsService.go(notifications)
+
+      const createdResult = await ScheduledNotificationModel.query().where('eventId', eventId)
+
+      expect(createdResult).equal([
+        {
```
Cruikshanks, line 114: Same comment as above applies here

```suggestion
      let result = await scheduledNotification.$query()

      expect(result).equal(
        {
```

Then repeat for `scheduledNotification2`.

Cruikshanks (changes requested):
Just a couple of suggested changes for how to get the records in the tests.

Cruikshanks (approved):

PR #1810, opened 2025-03-14
Migrate legacy communications view page

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4959

Having migrated the view licence page to our new code base, we are starting to look at various pages linked from it. Although the communications tab has been created to show all notifications sent to licence holders, the page links to a legacy view communications page. As part of our work to replace legacy views to better support the service, this page needs migrating.

This PR is focused on rebuilding the communications view page from the legacy code base.

PR #1835, opened 2025-03-21
Add `marked` npm package for markdown rendering

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4959

As part of the work to migrate the legacy [view licence communications page](https://github.com/DEFRA/water-abstraction-system/pull/1810), we needed a method to parse the `plaintext` returned by the [Notify](https://www.notifications.service.gov.uk/) service into HTML.

The `plaintext` is the content from the template used for the notification and any personalisation the service provides. **Notify** allows us to [format the content](https://www.notifications.service.gov.uk/using-notify/formatting) using something that is [Markdown](https://www.markdownguide.org/) in all but name.

The legacy team used [marked](https://www.npmjs.com/package/marked) to parse **Notify** `plaintext` for the legacy licence communications view. They made it available for use in views as a [custom Nunjucks filter](https://mozilla.github.io/nunjucks/api#custom-filters). This also allowed them to deal with the one discrepancy: **Notify** uses `^` for inset text, whereas **Markdown** uses `>`.

To be honest, we think for once the legacy code nailed it! 😱 :trollface:  

So, this PR does the same

- Adds [marked](https://www.npmjs.com/package/marked) as a new dependency
- Adds a [custom Nunjucks filter](https://mozilla.github.io/nunjucks/api#custom-filters)
- Deals with the `^` / `>` discrepency


app/plugins/views.plugin.js
```diff
@@ -134,6 +136,8 @@ function prepare(config, next) {
   // This gets assigned to Vision's compileOptions which is passed into `compile()` above as `options`.
   const environment = Nunjucks.configure(paths)
 
+  environment.addFilter('markdown', markdown)
```
Cruikshanks, line 139:
```suggestion
  // Add custom filter to support rendering Notify notifications as HTML
  environment.addFilter('markdown', markdown)
```

Cruikshanks (changes requested):
One extra comment, and I also think it would be prudent to add a basic test specifically for the filter (our first!) just to confirm that

- Our logic for replacing `^` is working
- That we are calling `marked` correctly

Shout if you need any help with that.

Cruikshanks (approved):

PR #1844, opened 2025-03-25
Feature job notificaitons status update

jonathangoulding:

As part of the work to migrate the legacy notifications into the system repo, we need to have a mechanism to check and update the notifications statuses. 

This change updates the ProcessNotificationsStatusUpdatesService to get the events (which statuses needed to be updated), call notify to get a status update and then persists the updated status. 

When an error occurs the error count will be updated for the event.


app/services/jobs/notifications/notifications-status-updates.service.js
```diff
@@ -5,10 +5,44 @@
...
+  }
+}
+
+async function _processEvent(event) {
+  const notifications = []
+
+  for (const scheduledNotification of event.scheduledNotifications) {
+    const notifyResponse = await NotifyStatusService.go(scheduledNotification.notifyId)
+
+    if (!notifyResponse.errors) {
```
jonathangoulding, line 31: How should we handle the error. Shall we log each error as it comes. 

It will be 4xx / 5xx.

Cruikshanks (approved):

PR #1846, opened 2025-03-25
Update notification setup error count

jonathangoulding:

The legacy code keeps track of errors by using a key on the 'events.metadata.error' field.

This change updates the batch notifications logic to update the error count in the 'events' table.

This error count is used on the reporting page to show how many errors have occurred when sending the notifications.

This change also adds a new 'UpdateEventService'. The current use case is only for updating the metadata but there may be other updates required in the future so this naming has been left as open as possible to facilitate change.


test/services/notifications/setup/batch-notifications.service.test.js
```diff
@@ -367,6 +369,30 @@ describe('Notifications Setup - Batch notifications service', () => {
         }
       ])
     })
+
+    it('should update the "event.metadata.error"', async () => {
+      await BatchNotificationsService.go(testRecipients, determinedReturnsPeriod, referenceCode, journey, eventId)
+
+      const updatedResult = await EventModel.query().where('id', eventId).first()
```
Cruikshanks, line 376: Use `findById()`. It would be more idiomatic for this context, and returns just a single result automatically.

```suggestion
      const updatedResult = await EventModel.query().findById(eventId)
```

test/services/notifications/setup/update-event.service.tests.js
```diff
@@ -0,0 +1,56 @@
...
+// Test helpers
+const EventHelper = require('../../../support/helpers/event.helper.js')
+const EventModel = require('../../../../app/models/event.model.js')
+
+// Thing under test
+const UpdateEventService = require('../../../../app/services/notifications/setup/update-event.service.js')
+
+describe('Notifications Setup - Update event service', () => {
+  let event
+  const errorCount = 5
```
Cruikshanks, line 19:
```suggestion
  const errorCount = 5
  
  let event
```

app/services/notifications/setup/update-event.service.js
```diff
@@ -0,0 +1,35 @@
...
+ * related 'scheduledNotifications'.
+ *
+ * @param {string} id
+ * @param {number} errorCount - The number of errors that occurred for the related 'scheduledNotifications'.
+ *
+ */
+async function go(id, errorCount) {
+  await EventModel.query()
+    .findById(id)
+    .patch({
```
Cruikshanks, line 28: We need to be patching the `updatedAt` property as well.

test/services/notifications/setup/update-event.service.tests.js
```diff
@@ -0,0 +1,56 @@
...
+
+  it('should update the event', async () => {
+    await UpdateEventService.go(event.id, errorCount)
+
+    const updatedResult = await EventModel.query().findById(event.id)
+
+    expect(updatedResult.metadata.error).to.equal(errorCount)
+
+    expect(updatedResult).equal({
+      createdAt: updatedResult.createdAt,
```
Cruikshanks, line 40: If we are going to check this, I think it would be more precise to say it should match the date on the source record.

```suggestion
      createdAt: event.createdAt,
```

Cruikshanks (changes requested):
Just some NITS.

test/services/notifications/setup/batch-notifications.service.test.js
```diff
@@ -367,6 +369,30 @@ describe('Notifications Setup - Batch notifications service', () => {
...
+
+    it('should update the "event.metadata.error"', async () => {
+      await BatchNotificationsService.go(testRecipients, determinedReturnsPeriod, referenceCode, journey, eventId)
+
+      const updatedResult = await EventModel.query().findById(eventId)
+
+      expect(updatedResult.metadata.error).to.equal(1)
+
+      expect(updatedResult).to.equal({
+        createdAt: updatedResult.createdAt,
```
Cruikshanks, line 381: I hadn't spotted it on this one.

```suggestion
        createdAt: event.createdAt,
```

Cruikshanks (changes requested):

Cruikshanks (approved):

PR #1850, opened 2025-03-27
Add error logging when a call to Notify fails

jonathangoulding:

As part of getting the status updates for notifications from GOV.UK Notify we have identified an issue in how we handle errors.

For some errors we store them in the 'scheduledNotifications' to align with the legacy and for the status this is no required.

This change adds the logging to the Notify services when an error occurs in Notify. We format this error (strip out the relevant information).

Notify returns failed notification within 'response.data.status' this would be something like 'technical-failure'. This is an error when Notify has tried to send the notification, and it has failed. Not a failure in the call to Notify.

### DIFF
--- a/app/services/notify/notify-email.service.js
+++ b/app/services/notify/notify-email.service.js
@@ -53,11 +53,15 @@ async function _sendEmail(notifyClient, templateId, emailAddress, options) {
       statusText: response.statusText.toLowerCase()
     }
   } catch (error) {
-    return {
+    const formattedError = {
       status: error.status,
       message: error.message,
       errors: error.response.data.errors
     }
+
+    global.GlobalNotifier.omfg('Notify send email failed', null, formattedError)
+
+    return formattedError
   }
 }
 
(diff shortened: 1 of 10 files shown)


app/services/notify/notify-email.service.js
```diff
@@ -53,11 +53,15 @@ async function _sendEmail(notifyClient, templateId, emailAddress, options) {
...
-    return {
+    const formattedError = {
       status: error.status,
       message: error.message,
       errors: error.response.data.errors
     }
+
+    global.GlobalNotifier.omfg('Notify send email failed', null, formattedError)
+
+    return formattedError
```
Cruikshanks, line 64: Unfortunately, this won't work with how `omfg()` is formulated in `app/lib/base-notifier.lib.js`.

It expects something of type 'Error' for the 3rd argument. What you are providing is data, so should be passed as the second argument.

It is fine to call `omfg()` without an actual error, in which case you just leave the third argument off.

```suggestion
    const errorDetails = {
      status: error.status,
      message: error.message,
      errors: error.response.data.errors
    }

    global.GlobalNotifier.omfg('Notify send email failed', errorDetails)

    return errorDetails
```

This will apply to all the other places you are calling `omfg()` in this way.

Cruikshanks (changes requested):

Cruikshanks (approved):

PR #1854, opened 2025-03-28
Update notifications status update to batch scheduled notifications

jonathangoulding:

There was a previous assumption to use the scheduled notifications event as part of the batch process. This idea introduced some complex logic to handle counting how many request we have made to Notify.

This change updates the existing fetch logic to return an array of scheduled notifications rather than an event with scheduled notifications.

This allows the batch mechanism to function the same as it does in other places.

This change also updates the notification status update service to use the batch mechanism previously mentioned.

As we are using an upsert style query (for batch inserting) the 'water.scheduledNotifications' schema wants us to provide a 'createdAt' even though all our 'scheduledNotifications' should have a 'createdAt' date.

We should not be creating any new 'scheduledNotifications' just updating them. So we provide the 'createdAt' date to be reinserted.


app/services/jobs/notifications/notifications-status-updates.service.js
```diff
@@ -5,58 +5,85 @@
...
+
+async function _delay(delay) {
+  return setTimeout(delay)
+}
+
+async function _notificationStatus(scheduledNotification) {
   const notifyResponse = await NotifyStatusService.go(scheduledNotification.notifyId)
 
-  if (!notifyResponse.errors) {
+  if (notifyResponse.errors) {
```
jonathangoulding, line 61: This is necessary to avoid complicated logic to remove any 'undefined' statuses when the promise resolves.

test/services/jobs/notifications/fetch-scheduled-notifications.service.test.js
```diff
@@ -12,13 +12,12 @@ const EventHelper = require('../../../support/helpers/event.helper.js')
 const ScheduledNotificationHelper = require('../../../support/helpers/scheduled-notification.helper.js')
 
 // Thing under test
-const FetchEventNotificationsService = require('../../../../app/services/jobs/notifications/fetch-event-notifications.service.js')
+const FetchScheduledNotificationsService = require('../../../../app/services/jobs/notifications/fetch-scheduled-notifications.service.js')
 
 describe('Job - Notifications - Fetch event notifications service', () => {
```
jonathangoulding, line 17: ``
```suggestion
describe('Job - Notifications - Fetch scheduled notifications service', () => {
```

app/services/jobs/notifications/fetch-scheduled-notifications.service.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+const ScheduledNotificationModel = require('../../../models/scheduled-notification.model.js')
+
+const SEVEN_DAYS = 7
+
+/**
+ * Fetches the scheduled notifications with the status 'sending'.
+ *
+ * The function returns a list of 'scheduledNotifications' that have a status of 'sending' and are lees than 7 days old.
```
Cruikshanks, line 15:
```suggestion
 * The function returns a list of 'scheduledNotifications' that have a status of 'sending' and are less than 7 days old.
```

Cruikshanks (changes requested):
One NIT!

Cruikshanks (approved):

PR #1856, opened 2025-03-31
Add update event error count service

jonathangoulding:

As part of the work to get status updates from notify, we need to calculate the error count for failed scheduled notifications.

This logic follows the legacy codes approach to calculating errors, but we will do this count only when we check for status updates (if there are any).

We will get a status update from notify at least once.

This change adds the logic to iterate over an array of event id's and updates the events error count in the metadata ('event.metadata.error)'.


jonathangoulding:
I will add this to the batch notification service in a separate change as we will need to flat map the event id's.

app/services/jobs/notifications/update-event-error-count.service.js
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * **Reference**:
+ * - For more information about error statuses from Notify, refer to the documentation:
+ * [Get the status of multiple messages](https://docs.notifications.service.gov.uk/node.html#get-the-status-of-multiple-messages-response).
+ *
+ * @param {string[]} eventIds - an array of event ids to update the error count for
+ *
+ */
+async function go(eventIds) {
+  const query = _query(eventIds)
```
jonathangoulding, line 39:
```suggestion
  const query = _query()
```

test/services/jobs/notifications/update-event-error-count.service.test.js
```diff
@@ -0,0 +1,129 @@
...
+      it('updates the error count"', async () => {
+        await UpdateEventErrorCountService.go([event.id])
+
+        const result = await EventModel.query().findById(event.id)
+
+        expect(result.metadata.error).to.equal(2)
+      })
+    })
+
+    describe('and the "scheduledNotification" have existing errors', () => {
```
Cruikshanks, line 69:
```suggestion
    describe('and the "scheduledNotifications" have existing errors', () => {
```

test/services/jobs/notifications/update-event-error-count.service.test.js
```diff
@@ -0,0 +1,129 @@
...
+        eventId: event.id,
+        status: 'error'
+      })
+    })
+
+    describe('and the "scheduledNotifications" have a single error', () => {
+      it('updates the error count"', async () => {
+        await UpdateEventErrorCountService.go([event.id])
+
+        const result = await EventModel.query().findById(event.id)
```
Cruikshanks, line 46: Will this not work?

I appreciate it's the same result, but IMHO it makes it explicit we testing that the event we created got updated by the service.

```suggestion
        const refreshEvent = await event.$query()
```

(Same suggestion applies to all applicable tests).

Cruikshanks (changes requested):

test/services/jobs/notifications/update-event-error-count.service.test.js
jonathangoulding: I keep forgetting this.

Cruikshanks (approved):

PR #1864, opened 2025-04-03
Update Notify client proxy

jonathangoulding:

Previous work has found an issue with using Notify behind our proxy - https://github.com/DEFRA/water-abstraction-import/pull/1076.

This change updates our implementation of the Notify client allow us to pass in the proxy url.

### DIFF
--- a/app/services/notify/notify-client.service.js
+++ b/app/services/notify/notify-client.service.js
@@ -1,0 +1,63 @@
+'use strict'
+
+/**
+ * Creates a Notify Client
+ * @module NotifyClientService
+ */
+
+const NotifyClient = require('notifications-node-client').NotifyClient
+const axios = require('axios')
+
+const { HttpsProxyAgent } = require('hpagent')
+
+const NotifyConfig = require('../../../config/notify.config.js')
+const RequestConfig = require('../../../config/request.config.js')
+
+/**
+ * Creates a Notify Client
+ *
+ * Returns an instance of {@link https://github.com/alphagov/notifications-node-client | notifications-node-client}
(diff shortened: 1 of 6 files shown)


.env.example
```diff
@@ -82,3 +82,5 @@ ENABLE_SYSTEM_RETURNS_VIEW=true
 GOV_UK_NOTIFY_API_KEY=
 NOTIFICATIONS_BATCH_DELAY=
 NOTIFICATIONS_BATCH_SIZE=
+# The proxy url used for the Notify client
+PROXY=
```
Cruikshanks, line 86: The env var we should be using is `http_proxy`.

It's a standard convention (even the use of lowercase,) and all our environments will have this env var.

The previous team used `PROXY` in the import app, and it seemed unnecessary to add a new one or update their code to use `http_proxy`. Hence, I used `PROXY` in my change.

The good news is that it is already available: `config/request.config.js`!

package.json
```diff
@@ -43,6 +43,7 @@
     "@hapi/yar": "^11.0.2",
     "@joi/date": "^2.1.0",
     "@smithy/node-http-handler": "^2.0.4",
+    "axios": "^1.8.4",
```
Cruikshanks, line 46: I didn't need to add this explicitly in the import app. It is a dependency of Notify, and if they change to something else, we'll also need to change, so we shouldn't need to manage our own version.

Cruikshanks (changes requested):

.env.example
jonathangoulding: bugger

Cruikshanks (approved):

PR #1868, opened 2025-04-04
Update notifications status updates service

jonathangoulding:

We are seeing an issues with the same test timing out.

This test is taking just a bit longer than 2 seconds to complete. Our default timeout is 2 seconds.

This test only takes longer when run after the 'Notifications Setup - Batch notifications service' test suite.

When run before / alone the test only takes roughly 224 ms.

The change increases the timeout to 3 seconds for this specific test.

This keeps the time low to highlight further issues.


jonathangoulding:
Screenshot 2025-04-04 at 12 33 32

robertparkinson (approved):

PR #1869, opened 2025-04-04
Update Notify services to requests

jonathangoulding:

We have an established pattern of putting third party services we interact with in the requests folder and name them request.

This change refactors the Notify services into Notify Requests.

All the module names, file names will be updated.

The test folder will align with the app folder.

### DIFF
--- a/app/requests/notify/notify-client.request.js
+++ b/app/requests/notify/notify-client.request.js
@@ -2,7 +2,7 @@
 
 /**
  * Creates a Notify Client
- * @module NotifyClientService
+ * @module NotifyClientRequest
  */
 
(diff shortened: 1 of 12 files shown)


app/requests/notify/notify-client.request.js
```diff
@@ -2,7 +2,7 @@
 
 /**
  * Creates a Notify Client
- * @module NotifyClientService
+ * @module NotifyClientRequest
```
jonathangoulding, line 5: I have left this as exporting go as it is creating the client to allow the send

Cruikshanks (approved):

PR #1874, opened 2025-04-07
Refactor scheduled notification to notifications services

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4979

We have previously used a 'scheduledNotifications' view for the 'water.scheduledNotifications' table. We have decided to move to a new view called 'notifications' this is, simply, because we do not schedule notifications to be sent, we just send them.

This change update the notifications logic to use this new notification model and view. 

Anything previously using 'scheduledNotifications' or 'scheduledNotification' has been replaced with 'notifications' and 'notification' respect

### DIFF
--- a/test/fixtures/notifications.fixture.js
+++ b/test/fixtures/notifications.fixture.js
@@ -50,7 +50,7 @@ function notification() {
       '\n' +
       '# Why you are receiving this notification\n' +
       '\n',
-    sendAfter: new Date('2024-07-02T16:52:17.000Z'),
+    createdAt: new Date('2024-07-02T16:52:17.000Z'),
     event
   })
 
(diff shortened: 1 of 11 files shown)


test/fixtures/notifications.fixture.js
```diff
@@ -50,7 +50,7 @@ function notification() {
       '\n' +
       '# Why you are receiving this notification\n' +
       '\n',
-    sendAfter: new Date('2024-07-02T16:52:17.000Z'),
+    createdAt: new Date('2024-07-02T16:52:17.000Z'),
```
jonathangoulding, line 53: @rvsiyad i have assumed this date change does not affect behaviour ?

PR #1883, opened 2025-04-10
Migrate view billing account page

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4988

Having migrated the view licence page to our new code base, we are starting to look at various pages linked from it. Although the bills tab has been created to show all bills sent to licence holders, the page links to a legacy view billing account page. As part of our work to replace legacy views to better support the service, this page needs migrating.

This PR is focused on rebuilding the billing account view page from the legacy code base.


app/views/billing-accounts/view.njk
```diff
@@ -0,0 +1,194 @@
...
+          }
+        }
+      ]
+    }) }}
+  </div>
+
+  <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+
+  {# Bill details #}
+  {% if bills.length <= 0 %}
```
Jozzey, line 79: As `bills.length` can never be less than zero I would write it like this:
```suggestion
  {% if bills.length === 0 %}
```

app/views/billing-accounts/view.njk
```diff
@@ -0,0 +1,194 @@
...
+    <div class="govuk-body govuk-!-padding-top-3">
+      <p>No sent bills for this billing account</p>
+    </div>
+  {% elif bills.length > 0 %}
```
Jozzey, line 83: It shouldn't be nessessary to use an "else if" here, we should just be able to use an "else" as there are only 2 conditions that we are checking for. The length is either zero or it isn't.
```suggestion
  {% else %}
```

test/controllers/billing-accounts.controller.test.js
```diff
@@ -46,6 +47,34 @@ describe('Billing Accounts controller', () => {
     Sinon.restore()
   })
 
+  describe('/billing-accounts/{billingAccountId}', () => {
+    describe('GET', () => {
+      beforeEach(async () => {
```
Jozzey, line 52: This shouldn't need to be `async`
```suggestion
      beforeEach(() => {
```

test/controllers/billing-accounts.controller.test.js
```diff
@@ -46,6 +47,34 @@ describe('Billing Accounts controller', () => {
...
+          url: '/billing-accounts/2e71429d-3fd1-4ed1-a45e-eb5616873018',
+          auth: {
+            strategy: 'session',
+            credentials: { scope: ['billing'] }
+          }
+        }
+      })
+
+      describe('when the request succeeds', () => {
+        beforeEach(async () => {
```
Jozzey, line 64: This shouldn't need to be `async`
```suggestion
        beforeEach(() => {
```

Jozzey (changes requested):
Just a few minor things. Looks good

PR #1887, opened 2025-04-11
Check current return cycle exists

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4542

As part of the testing for the quarterly returns it was discovered that we lost the code for creating a return log if one doesn't already exist along the way with the refactors. 

This PR refactors the code to check for the existence of the current return cycles when doing the twice annual return log job as well as in the individual licence return log job. If things go as planned the new return cycles should be created as part of the annual jobs but in case those jobs are not ran we need the coverage in the individual licence return logs code to make sure we're don't miss any return logs.


app/services/jobs/return-logs/create-current-return-cycle.service.js
```diff
@@ -14,16 +14,44 @@ const {
 const ReturnCycleModel = require('../../../models/return-cycle.model.js')
 
 /**
- * Given a return cycle type, it generates the cycle data based on the current date and creates it
- *
- * @param {boolean} summer - true if creating the current summer return cycle, else false
+ * Verify that the town most recent return cycles are the most current ones. If they're not create the missing cycles.
```
Cruikshanks, line 17: The convention is that the summary for the function is the same as the module description.

I believe this will be a better summary, though, once the references to 'the town' are dealt with!

I wasn't sure where you were going with this, so I couldn't suggest anything! 😂

Cruikshanks (changes requested):
My interpretation of how we are dealing with this is that we're assuming we only need to worry about missing return cycles for the current period.

- When we run `CreateReturnCycleService` we only look at the last two entries
- We only call `CreateReturnCycleService` when no change date has been provided i.e. the current date is used

I'm not so confident we can guarantee this will always be the case. For example, what if a user decides to insert a very old return version? There would be no return cycle and we wouldn't create one.

But if we switch to checking the return cycles for a particular cycle and date, then I believe we could

- Combine `CreateReturnCycleService` and `FetchCurrentReturnCycleService` into a single `CheckReturnCycleService`
- Just check and process the return cycle needed, rather than always processing Summer and winter

Both the job and `ProcessLicenceReturnLogService` know the cycle needed, and what the change date is. So, these could be passed to `CheckReturnCycleService`, which can then check if such a cycle exists.

If it does, it simply returns it. If it doesn't it creates just the return cycle needed, then returns it.

If you are unsure about what I mean, feel free to park this until I return.

PR #1893, opened 2025-04-16
Notifications report presenter

robertparkinson:

As part of our work to replace the legacy UI we're slowly porting over pages to the new system. This PR adds the presenter and view for the index page.

### DIFF
--- a/app/presenters/notifications/index.presenter.js
+++ b/app/presenters/notifications/index.presenter.js
@@ -1,0 +1,65 @@
+'use strict'
+
+/**
+ * Formats data for the `/notifications` page
+ * @module NotificationsIndexPresenter
+ */
+
+const { formatLongDate } = require('../base.presenter.js')
+
+/**
+ * Formats data for the `/notifications` page
+ *
+ * @param {object} data - The object containing the notifcations sent
+ *
+ * @returns {object} - The data formatted for the view template
+ */
+function go(data) {
+  return {
+    backLink: '/manage',
+    headers: _tableHeaders(),
+    rows: _tableRows(data),
+    pageTitle: 'View sent notices'
+  }
+}
+
+function _tableHeaders() {
+  return [
+    {
+      text: 'Date'
+    },
+    {
+      text: 'Notification type'
+    },
+    {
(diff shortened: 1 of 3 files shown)


app/presenters/notifications/index.presenter.js
```diff
@@ -0,0 +1,65 @@
...
+      { text: formatLongDate(notification.createdAt) },
+      { html: `<a href="/notifications/report/${notification.id}">${name}</a>` },
+      { text: notification.issuer },
+      { text: notification.recipientCount, format: 'numeric' },
+      { html: notification.errorCount ? `<strong class="govuk-tag govuk-tag--orange">ERROR</strong>` : '' }
+    ])
+  }
+
+  return rows
+}
```
Jozzey, line 61: You could refactor this function to use map like this:
```suggestion
function _tableRows(data) {
  return data.map((notification) => {
    const name = notification.alertType === 'warning' ? `Warning - ${notification.name}` : notification.name

    return [
      { text: formatLongDate(notification.createdAt) },
      { html: `<a href="/notifications/report/${notification.id}">${name}</a>` },
      { text: notification.issuer },
      { text: notification.recipientCount, format: 'numeric' },
      { html: notification.errorCount ? `<strong class="govuk-tag govuk-tag--orange">ERROR</strong>` : '' }
    ]
  })
}
```

app/presenters/notifications/index.presenter.js
```diff
@@ -0,0 +1,65 @@
+'use strict'
+
+/**
+ * Formats data for the `/notifications` page
+ * @module NotificationsIndexPresenter
```
Jozzey, line 5: Regarding the name of this file `index.presenter.js` I think there was a time when we were using that filename in our services/presenters etc but I don't thinnk we do anymore. I did a search in this repo for files named `index.` and can't find any except for the one in the root. It would appear the ones that were named index have since been renamed, for example `app/presenters/bill-runs/index-bill-runs.presenter.js`. So if we are following that pattern I think this file should be renamed to `index-notifications.presenter.js`

Jozzey (approved):

PR #1895, opened 2025-04-17
Fix View Return missing lines for NALD imports

StuAA78:
https://eaflood.atlassian.net/browse/WATER-5002

View Return is missing lines for data imported from NALD. This is because it was developed and tested against data entered via the frontend, where every day/week/month has a line even if the volume/reading is 0. For data imported from NALD, we only have entires for days/weeks/months where a value was reported. Hence, only these lines are visible in the summary table.


StuAA78:
This is no longer required as the issue has been fixed in the underlying data.

PR #1900, opened 2025-04-22
Fix error summary for requirements marked quarterly

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4955

After previously add an error summary to the return check page to prevent requirements marked summer & quarterly submission. While testing the page, it was noticed that the error summary box was not red and the error message was not hyperlinked as per the [GDS](https://design-system.service.gov.uk/components/error-summary/).

This PR adds the red error summary box and the hyper linked error message to the return check page.


rvsiyad:
This bug was deal with as part of [Amend validation for returns marked quarterly submissions](https://github.com/DEFRA/water-abstraction-system/pull/1903).

rvsiyad:
This bug was deal with as part of [Amend validation for returns marked quarterly submissions](https://github.com/DEFRA/water-abstraction-system/pull/1903).

PR #1901, opened 2025-04-22
Return logs - fix weekly/monthly line label

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4813

When editing/creating a monthly or weekly return log the label on each line currently just shows an end date for the week/month e.g. 30 April 2023.

This needs to be corrected so that when the return is weekly the label would show "Week ending 30 April 2023" & when monthly the day would be dropped so that it just displays "April 2023".

This PR will correct these labels when editing a return logs readings/volumes.


app/presenters/return-logs/setup/readings.presenter.js
```diff
@@ -58,8 +58,22 @@ function _inputLines(lines) {
...
+  if (returnsFrequency === 'month') {
+    return labelDate.toLocaleDateString('en-GB', { month: 'long', year: 'numeric' })
+  }
+
+  return formatLongDate(labelDate)
+}
+
 function _pageTitle(date) {
-  const titleDate = date.toLocaleDateString('en-GB', { year: 'numeric', month: 'long' })
+  const titleDate = date.toLocaleDateString('en-GB', { month: 'long', year: 'numeric' })
```
Cruikshanks, line 76: We already have a helper for doing this in the `BasePresenter`, which you were previously using. Could we update this to use the existing helper?

```suggestion
  const titleDate = formatLongDate(date)
```

app/presenters/return-logs/setup/volumes.presenter.js
```diff
@@ -59,8 +59,22 @@ function _inputLines(lines) {
...
+  if (returnsFrequency === 'month') {
+    return labelDate.toLocaleDateString('en-GB', { month: 'long', year: 'numeric' })
+  }
+
+  return formatLongDate(labelDate)
+}
+
 function _pageTitle(date) {
-  const titleDate = date.toLocaleDateString('en-GB', { year: 'numeric', month: 'long' })
+  const titleDate = date.toLocaleDateString('en-GB', { month: 'long', year: 'numeric' })
```
Cruikshanks, line 77:
```suggestion
  const titleDate = formatLongDate(date)
```

Cruikshanks (changes requested):

app/presenters/return-logs/setup/readings.presenter.js
Jozzey: This one is a bit different, it just displays the month and year and drops the day.

As discussed will move it into the base presenter.

Cruikshanks (approved):

PR #1902, opened 2025-04-22
Add a feature toggle to the water abstraction alert page 'create alert' link

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4979

We are migrating the legacy water abstraction alert pages into this repo.

This change adds a new feature toggle 'ENABLE_MONITORING_STATIONS_ALERT_NOTIFICATIONS'. This will be used to hide the 'new' journey as it goes through our development process.

This change also adds another 'NOTIFICATION_TYPES' called ' 'abstraction-alert'. This key will be used to reuse existing parts of our notification setup journey.

We do need to use the monotoning station id in the upcoming journey. We will add this id to the session with the basic data we need from the database but this will be done in another change as it feels too much for this change. But we are highlighting it as we have added it to the query string.


app/presenters/monitoring-stations/view.presenter.js
```diff
@@ -84,6 +86,18 @@ function _alert(status, statusUpdatedAt) {
   return sentenceCase(status)
 }
 
+function _links(monitoringStationId) {
+  let createAlert = `/system/notifications/setup?journey=abstraction-alert&monitoringStationId=${monitoringStationId}`
+
+  if (!FeatureFlagsConfig.enableMonitoringStationsAlertNotifications) {
```
jonathangoulding, line 92: I have done it this way so we can just remove this code when ready.

app/services/notifications/setup/initiate-session.service.js
```diff
@@ -9,22 +9,29 @@ const SessionModel = require('../../../models/session.model.js')
...
-    redirectPath: 'ad-hoc-licence'
+    prefix: 'ADHC-',
+    redirectPath: 'ad-hoc-licence',
+    type: 'Ad hoc'
+  },
+  'abstraction-alert': {
+    journey: 'abstraction-alert',
+    prefix: 'WAA-',
+    redirectPath: 'abstraction-alert',
+    subType: 'waterAbstractionAlerts',
```
jonathangoulding, line 33: I have aded the subType and will update the others in a separate change. I can remove some unnecessary upstream logic.

PR #1908, opened 2025-04-23
Add abstraction alerts monitoring stations service

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5012

We need to use the monitoring station id to get the information needed to present in the notification setup abstraction alerts journey.

This change adds the Monitoring Station Service to fetch and format the data required for the notification setup abstraction alerts journey.

When the monitoring station id is passed as an arg to the initiate session service we will add the additional data to the session.

The query has been taken from the legacy and updated to use the the 'views'. Some fields have been omitted as we do not see a need for them currently, but they can be added back in as we build the journey. 

The additional data will be used across the journey and so a fixture has been added for all the tests to use so we have a single representation of the expected initial session data.


app/services/notices/setup/abstraction-alerts/fetch-monitoring-station.service.js
```diff
@@ -0,0 +1,110 @@
...
+ * @module FetchMonitoringStationService
+ */
+
+const { db } = require('../../../../../db/db.js')
+
+/**
+ * Orchestrates fetching the data needed for the Monitoring station journey
+ *
+ * @param {string} id
+ * @returns {Promise<{object}>}
```
Cruikshanks, line 14: I think this should be

```suggestion
 * @returns {Promise<object[]>}
```

app/services/notices/setup/abstraction-alerts/fetch-monitoring-station.service.js
```diff
@@ -0,0 +1,110 @@
...
+        join water.licences l on
+        (l.licence_id = lms.licence_id)
+        join public.licence_version_purpose_conditions lvpc on
+        lvpc.id = lms.licence_version_purpose_condition_id
+        join public.licence_version_purposes lvp on
+        lvp.id = lvpc.licence_version_purpose_id
+where
+    lms.licence_version_purpose_condition_id is not null
+  and lms.deleted_at is null
+  and  ms.id=?`
```
Cruikshanks, line 105: Just a NIT. Can we uppercase the keywords in the query? It's such a convention in SQL, you've even used uppercase in your examples 😂 

I'd also add a `;` to the end of the query for good measure.

app/services/notices/setup/abstraction-alerts/fetch-monitoring-station.service.js
```diff
@@ -0,0 +1,110 @@
+'use strict'
+
+/**
+ * Orchestrates fetching the data needed for the Monitoring station journey
```
Cruikshanks, line 4: I think we can just say...

```suggestion
 * Fetches the data needed for the monitoring station abstraction alert journey
```

Orchestration is more the purview of the calling service.

app/services/notices/setup/abstraction-alerts/fetch-monitoring-station.service.js
```diff
@@ -0,0 +1,110 @@
...
+ * @module FetchMonitoringStationService
+ */
+
+const { db } = require('../../../../../db/db.js')
+
+/**
+ * Orchestrates fetching the data needed for the Monitoring station journey
```
Cruikshanks, line 11:
```suggestion
 * Fetches the data needed for the monitoring station abstraction alert journey
```

app/services/notices/setup/abstraction-alerts/monitoring-station.service.js
Cruikshanks, line 1: This is a hard one, because we're early in the journey, and its purpose may change.

But if you check, you should find that all our services start with a verb: create, generate, determine, calculate, etc.

This is sticking out at the moment because it does not follow that convention.

For now, can I suggest `determine-licence-monitoring-stations.service.js` 🤷

app/services/notices/setup/abstraction-alerts/fetch-monitoring-station.service.js
Cruikshanks, line 1: As this is not fetching a monitoring station, but the licence monitoring stations linked to it, I'd suggest renaming this as `fetch-licence-monitoring-stations.service.js`.

Cruikshanks (changes requested):
Hope these suggested changes make sense as a first pass.

Cruikshanks (approved):

PR #1913, opened 2025-04-25
Seed historic return cycles for tests

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-5007

As part of our work to replace NALD as the main source of managing returns we need to be able to handle not only new returns but historic changes to returns. This PR updates the seed function we use to populate return cycles to include the historic ones so when we're testing we have match for what is on the deployed environments.


db/seeds/16-return-cycles.seed.js
```diff
@@ -10,19 +10,23 @@ const {
...
 
+    if (allYearReturnCycle.startDate === new Date('2019-04-01')) {
+      allYearReturnCycle.dueDate = new Date('2020-10-16')
+    }
+
     await _upsert(summerReturnCycle)
     await _upsert(allYearReturnCycle)
+
+    year++
+    determinationDate = new Date(`${year}-04-01`)
```
Cruikshanks, line 29: If you made this the first line of the `while` loop, could you not then convert line 15 to just be `let determinationDate`?

Cruikshanks (changes requested):
Done as a 'request for change' but I could be wrong!

db/seeds/16-return-cycles.seed.js
robertparkinson: sure, I can switch it to a do while loop

Cruikshanks (approved):

PR #1917, opened 2025-04-25
Add code scaffolding script for services and presenters

jonathangoulding:
We regularly create new services and presenters with consistent structure, we have style guides to follow.

This script automates that repetitive setup, reduces human error, and saves time by enforcing conventions and reusing templates.

Includes support for:
- Nested module paths
- Auto-generation of service and presenter files in app/ and test/ directories
- Template placeholders (__MODULENAME__, __REQUIRE_PATH__, __DESCRIBE_LABEL__)
- Skipping existing files to prevent overwrites
- Separate test templates for services and presenters


jonathangoulding:
An additional step would be to add an additional filed / command to add a scaffolded 'fetch'. 

We can add a new fetch service with the corresponding test, and as we tend to stub the fetch in the calling service we can add this stub to the service.test.js template.

templates/service.js
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * @module __MODULENAME__
+ */
+
+/**
+ * @returns {object} - The data formatted for the view template
+ */
+function go() {
+  return {}
```
jonathangoulding, line 11: Add the code for the presenter to be required and called.

jonathangoulding:
Another additional step would be to add the controller (or create the code in a separate tmp file which will need to be manually dropped into an existing controller). This would save having to write and require the controller to the service. 

We can also do this for the router.

jonathangoulding:
The above comments are some improvements to make the scaffolding more complete in functionality. 

We can have a default, say the service / presenter. And then have a select option (might not be a great option for the bash script). Maybe just an additional 'all' yes command input or something.

jonathangoulding:
> An additional step would be to add an additional filed / command to add a scaffolded 'fetch'.
> 
> We can add a new fetch service with the corresponding test, and as we tend to stub the fetch in the calling service we can add this stub to the service.test.js template.

Update - this has been added.

templates/presenter.js
```diff
@@ -0,0 +1,16 @@
+'use strict'
+
+/**
+ * @module __MODULENAME__
```
jonathangoulding, line 4: We can add the common description here 

```
Orchestrates presenting the data for `/notices/setup/{sessionId}/abstraction-alert/alert-type` page
```

templates/presenter.test.js
```diff
@@ -0,0 +1,25 @@
...
+// Thing under test
+const __MODULENAME__ = require('__REQUIRE_PATH__')
+
+describe('__DESCRIBE_LABEL__', () => {
+  let session
+
+  beforeEach(async () => {})
+
+  describe('when called', () => {
+    it('returns page data for the view', async () => {
```
jonathangoulding, line 19: remove redundant await

jonathangoulding:
We should could add the njk file aswell.

templates/service.js
```diff
@@ -0,0 +1,16 @@
+'use strict'
+
+/**
+ * @module __MODULE_NAME__
+ */
+
+/**
+ * @returns {object} - The data formatted for the view template
```
jonathangoulding, line 8: these should be 

```
{Promise<object>}
```

templates/helper.js
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable camelcase, no-unused-vars, strict */
+
+// Controller code
+const __SERVICE_NAME__ = require('__SERVICE_PATH__')
+
+async function view__NAME__(request, h) {
```
jonathangoulding, line 6: Add submit and POST

templates/presenter.js
```diff
@@ -0,0 +1,19 @@
+'use strict'
+
+/**
+ * Formats data for the `` page
```
jonathangoulding, line 4: We could maybe generate this from the folder paths. But it looks a but complicated. 

We could put a placeholder for the route and then it's a simple find and replace.

scaffold.sh
```diff
@@ -0,0 +1,331 @@
...
+  else
+    SOURCE_OUTPUT="app/${SUBFOLDER}/${SOURCE_FILE}"
+    TEST_OUTPUT="test/${SUBFOLDER}/${TEST_FILE}"
+  fi
+
+  # Helper
+  RELATIVE_UP_PATH=$(build_up_path)
+
+  # Test
+  DESCRIBE_LABEL="$(echo "$RAW_NAME" | sed -E 's/[-_]+/ /g' | awk '{for(i=1;i<=NF;++i) $i=toupper(substr($i,1,1)) substr($i,2)}1') $TYPE" # Test describe block
```
jonathangoulding, line 245: This could do with a but of work

robertparkinson (approved):

Cruikshanks (changes requested):
I've requested a change regards the README.

I've not checked the rest, but I'm specifically keeping an 'arms-length' approach.

This idea has come from the team, so I'd like the team to own it and keep it maintained. A haven from the rest of the service where ...

templates/helper.js
```diff
@@ -0,0 +1,52 @@
...
+
+  const pageData = await __SERVICE_NAME__.go(sessionId)
+
+  return h.view(`__VIEW_PATH__`, pageData)
+}
+
+async function submit__NAME__(request, h) {
+  const { sessionId } = request.params
+
+  await __SUBMIT_NAME__.go(sessionId)
```
jonathangoulding, line 18: This needs

```
  const {
    payload,
    params: { sessionId }
  } = request
```

jonathangoulding: same for the redirect 

return h.redirect(redirectUrl)

jonathangoulding: and the render data for the error

templates/validator.test.js
```diff
@@ -0,0 +1,33 @@
...
+const __MODULE_NAME__ = require('__REQUIRE_PATH__')
+
+describe('__DESCRIBE_LABEL__', () => {
+  let session
+
+  beforeEach(async () => {})
+
+  describe('when called with valid data', () => {
+    it('returns with no errors', () => {
+      const result = __MODULE_NAME__.go(session.id)
```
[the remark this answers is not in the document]
jonathangoulding: no need for async in the bfe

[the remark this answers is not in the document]
jonathangoulding: test need to be updated to simialr:

```
  describe('when called with valid data', () => {
    it('returns with errors', () => {
      const result = AlertTypeValidator.go(payload)

    })
  })

  describe('when called with valid data', () => {
    it('returns with no errors', () => {
      const result = AlertTypeValidator.go(payload)

      expect(result.error.details[0].message).to.equal('Enter a licence number')
    })
  })
```

templates/submit.service.test.js
```diff
@@ -0,0 +1,40 @@
...
+  let sessionData
+
+  beforeEach(async () => {
+    sessionData = {}
+    session = await SessionHelper.add({ data: sessionData })
+  })
+
+  describe('when called', () => {
+    it('continues the journey', async () => {
+      const result = await __MODULE_NAME__.go(session.id)
```
jonathangoulding, line 27: add a payload object

Cruikshanks (approved):

PR #1925, opened 2025-04-29
Enter readings/volumes - maximum number errors

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4870

Other tickets effected are the volumes https://eaflood.atlassian.net/browse/WATER-4813 and start reading https://eaflood.atlassian.net/browse/WATER-4848

During QA it was spotted that when an unsafe number is entered for the reading, the default error message displayed e.g. `"2024-04-01T00:00:00.000Z" must be a safe number`.

Whilst discussing this issue we have decided that we should set an actual maximum reading & volume rather than just relying on the maximum safe number, which is far higher than any value we would expect a user to enter.

After examining the data in production for previously submitted readings and volumes. We have come up with a maximum reading of: `99,999,999,999` and a maximum volume of `9,999,999,999` based on the existing data.

In this PR the validation messages for the readings and volumes are going to be updated to reflect this change.


Jozzey:
SQL query used to identify the highest meter reading in the DB:
```sql
SELECT MAX(value::NUMERIC) AS highest_reading
FROM (
  SELECT value
  FROM public.return_submissions,
       jsonb_each_text(metadata->'meters'->0->'readings')
) readings;
``` 

Query used to identify the highest volume in the DB:
```sql
select MAX(rsl.quantity)
from public.return_submission_lines rsl
```

Beckyrose200 (approved):

PR #1927, opened 2025-04-29
Add scaffolding pages for notices service

robertparkinson:

As part of our work to replace the legacy UI we're slowly porting over pages to the new system. This PR migrates the notices report page. This is the basic scaffolding of the different needed services.


app/views/notices/index.njk
```diff
@@ -0,0 +1,217 @@
...
+        name: "filterNotificationTypes",
+        fieldset: {
+          legend: {
+            text: "Notification type",
+            isPageHeading: false
+          }
+        },
+        items: [
+          {
+            value: "legacy-notifications",
```
jonathangoulding, line 43: I think this is missing filter by ad-hoc ?

robertparkinson: it's not on the list I have in the ticket but can double check with alan on standup

jonathangoulding (approved):

PR #1930, opened 2025-04-30
Add notices setup abstraction alerts - alert type submit

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5014

This change adds the alert-type submit service and validator.

It also adds the route and controller to for the submit alert types.


test/validators/notices/setup/abstraction-alerts/alert-type.validator.test.js
```diff
@@ -0,0 +1,44 @@
...
+  describe('when called with valid data', () => {
+    it('returns with no errors', () => {
+      const result = AlertTypeValidator.go(payload)
+
+      expect(result.value).to.exist()
+      expect(result.error).not.to.exist()
+    })
+  })
+
+  describe('when called with valid data', () => {
```
jonathangoulding, line 31:
```suggestion
  describe('when called with invalid data', () => {
```

Cruikshanks (approved):

PR #1939, opened 2025-05-02
Add notifications setup abstraction thresholds submit

jonathangoulding:

This change updates the alert-threshold submit service, view and validator. The selected thresholds will be added to the session.


app/services/notices/setup/abstraction-alerts/submit-alert-thresholds.service.js
```diff
@@ -38,6 +40,12 @@ async function go(sessionId, payload) {
 }
 
 async function _save(session, payload) {
+  const alertThresholdsKey = 'alert-thresholds'
+
+  session.alertThresholds = Array.isArray(payload[alertThresholdsKey])
+    ? payload[alertThresholdsKey]
+    : [payload[alertThresholdsKey]]
```
Cruikshanks, line 47: Personally, if I can't fit a ternary on a single line, then I stick with a standard if/else.

But for this specific scenario, we already have a pattern in place. Check out `app/services/return-versions/setup/submit-points.service.js` and `app/services/return-versions/setup/submit-purpose.service.js`.

The benefit is that it clarifies why the bit of logic exists.

> 😩 I say pattern, I can see they are slightly different 😭 .
>
> I think `submit-points.service.js` is the one to follow (and if you have a mo, please feel free to update `submit-purpose.service.js`

app/validators/notices/setup/abstraction-alerts/alert-thresholds.validator.js
```diff
@@ -8,6 +8,8 @@
 
 const Joi = require('joi')
 
+const errorMessage = 'Select applicable threshold(s)'
```
Cruikshanks, line 11: If this is a top level `const` then the variable name [should be declared in uppercase](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const#basic_const_usage)

```suggestion
const ERROR_MESSAGE = 'Select applicable threshold(s)'
```

Cruikshanks (changes requested):
Just NITS

jonathangoulding (commented):
-

Cruikshanks (approved):

PR #1952, opened 2025-05-06
Update return version start data validation

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-5031

When creating a return version you will create return logs which use the existing return cycles. We found an issue with missing returns cycles and have now fixed that but it only goes back to the oldest licence we have in the system. 

We are 99.999% certain no one will add a new licence with a start date before 1959-04-01. But just in case, for that 0.001% chance, we will extend the start date validation in the return version setup journey.

The validator already ensures that the start date entered equals or exceeds the licence’s start date. But should we encounter a licence with a start date before 1959-04-01 we need to stop it from being created as no return logs will be added.

### DIFF
--- a/app/validators/return-versions/setup/start-date.validator.js
+++ b/app/validators/return-versions/setup/start-date.validator.js
@@ -68,15 +68,11 @@ function _fullDate(payload) {
  * @returns {object} The date and messsage to use for earliest allowable date
  */
 function _minimumDateDetails(licenceStartDate) {
-  let minDate = licenceStartDate
-  let minMessage = 'Start date must be on or after the original licence start date'
-
   if (new Date(licenceStartDate) < new Date('1959-04-01')) {
-    minDate = '1959-04-01'
-    minMessage = 'Start date must be on or after 1 April 1959'
+    return { minDate: '1959-04-01', minMessage: 'Start date must be on or after 1 April 1959' }
   }
 
-  return { minDate, minMessage }
+  return { minDate: licenceStartDate, minMessage: 'Start date must be on or after the original licence start date' }
 }
 
 function _validateAnotherStartDate(payload, licenceStartDate, licenceEndDate) {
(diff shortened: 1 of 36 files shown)


app/validators/return-versions/setup/start-date.validator.js
```diff
@@ -53,16 +53,24 @@ function _fullDate(payload) {
 }
 
 function _validateAnotherStartDate(payload, licenceStartDate, licenceEndDate) {
+  let minDate = licenceStartDate
+  let minMessage = 'Start date must be on or after the original licence start date'
+
+  if (new Date(licenceStartDate) < new Date('1959-04-01')) {
+    minDate = '1959-04-01'
+    minMessage = 'Start date must be on or after 1 April 1959'
+  }
```
Cruikshanks, line 62: Could I ask that this be put into a function of its own? Perhaps `_minimumDateDetails()`, which returns `{ minDate, minMessage }`.

Not only does this move this 'noise' out of the validation method, it also gives us somewhere to add the same not we've used in the PR. So, should we come by in 6 months, we'll have an explanation for why 1959-04-01 is being used for this validation.

Cruikshanks (changes requested):

app/validators/return-versions/setup/start-date.validator.js
```diff
@@ -52,17 +52,46 @@ function _fullDate(payload) {
...
+function _minimumDateDetails(licenceStartDate) {
+  let minDate = licenceStartDate
+  let minMessage = 'Start date must be on or after the original licence start date'
+
+  if (new Date(licenceStartDate) < new Date('1959-04-01')) {
+    minDate = '1959-04-01'
+    minMessage = 'Start date must be on or after 1 April 1959'
+  }
+
+  return { minDate, minMessage }
```
Cruikshanks, line 79: Sorry, one more refactor to align more with guard clauses.

```suggestion
  if (new Date(licenceStartDate) < new Date('1959-04-01')) {
    return { minDate: '1959-04-01', minMessage: 'Start date must be on or after 1 April 1959' }
  }

  return { minDate: licenceStartDate, minMessage = 'Start date must be on or after the original licence start date' }
```

Cruikshanks (changes requested):

Cruikshanks (approved):

PR #1953, opened 2025-05-06
Update notifications setup abstraction thresholds view

jonathangoulding:

This change updates a previously wrong assumption of showing all thresholds for all alert type. This is not the case.

There are two restriction types 'stop' and 'reduce'. When either of these are selected as the alert type then we need to only show those restrictions, when the alert type is something else then show all the thresholds.

### DIFF
--- a/app/presenters/notices/setup/abstraction-alerts/alert-thresholds.presenter.js
+++ b/app/presenters/notices/setup/abstraction-alerts/alert-thresholds.presenter.js
@@ -19,7 +19,7 @@ function go(session) {
     backLink: `/system/notices/setup/${session.id}/abstraction-alerts/alert-type`,
     caption: session.monitoringStationName,
     pageTitle: 'Which thresholds do you need to send an alert for?',
-    thresholdOptions: _thresholdOptions(session.licenceMonitoringStations, session.alertThresholds, session.alertType)
+    thresholdOptions: _thresholdOptions(session.licenceMonitoringStations, session.alertType, session.alertThresholds)
   }
 }
 
@@ -40,7 +40,7 @@ function _relevantLicenceMonitoringStations(stations, alertType) {
   })
 }
 
-function _thresholdOptions(licenceMonitoringStations, alertThresholds = [], alertType) {
+function _thresholdOptions(licenceMonitoringStations, alertType, alertThresholds = []) {
   const relevantLicenceMonitoringStations = _relevantLicenceMonitoringStations(licenceMonitoringStations, alertType)
 
   return relevantLicenceMonitoringStations.map((licenceMonitoringStation) => {
--- a/app/presenters/notices/setup/abstraction-alerts/alert-thresholds.presenter.js
+++ b/app/presenters/notices/setup/abstraction-alerts/alert-thresholds.presenter.js
@@ -30,13 +30,12 @@ function go(session) {
  *
  * @private
  */
-function _relevantLicenceMonitoringStations(stations, alertType) {
-  return stations.filter((station) => {
-    if (alertType === 'stop' || alertType === 'reduce') {
-      return station.restriction_type === alertType
-    }
-
-    return Boolean(station.restriction_type)
+function _relevantLicenceMonitoringStations(licenceMonitoringStations, alertType) {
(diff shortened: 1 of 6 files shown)


app/presenters/notices/setup/abstraction-alerts/alert-thresholds.presenter.js
```diff
@@ -19,12 +19,31 @@ function go(session) {
...
+ */
+function _relevantLicenceMonitoringStations(stations, alertType) {
+  return stations.filter((station) => {
+    if (alertType === 'stop' || alertType === 'reduce') {
+      return station.restriction_type === alertType
+    }
+
+    return Boolean(station.restriction_type)
+  })
+}
```
Cruikshanks, line 41: Could I suggest this as an alternative (assuming the tests still pass)?

```suggestion
function _relevantLicenceMonitoringStations(licenceMonitoringStations, alertType) {
  if (alertType !== 'stop' && alertType !== 'reduce') {
    return licenceMonitoringStations
  }

  return licenceMonitoringStations.filter((licenceMonitoringStation) => {
    return licenceMonitoringStation.restriction_type === alertType
  })
}
```

In this version we only iterate the `licenceMonitoringStations` if the alert type is 'stop' or 'reduce' (rather than always). Plus we avoid the use of `Boolean()` which [MSDN says](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean/Boolean)

> Warning: You should rarely find yourself using `Boolean` as a constructor.

test/presenters/notices/setup/abstraction-alerts/alert-thresholds.presenter.test.js
```diff
@@ -64,6 +68,134 @@ describe('Notices Setup - Abstraction Alerts - Alert Thresholds Presenter', () =
...
+
+      describe('and the "alertType" is "reduce" ', () => {
+        beforeEach(() => {
+          session = {
+            ...AbstractionAlertSessionData.monitoringStation(),
+            alertType: 'reduce'
+          }
+        })
+
+        it('returns page data for the view, with only the thresholds with stop restrictions', () => {
```
Cruikshanks, line 129:
```suggestion
        it('returns page data for the view, with only the thresholds with reduce restrictions', () => {
```

test/presenters/notices/setup/abstraction-alerts/alert-thresholds.presenter.test.js
```diff
@@ -64,6 +68,134 @@ describe('Notices Setup - Abstraction Alerts - Alert Thresholds Presenter', () =
...
+                text: 'Level thresholds for this station (m)'
+              },
+              text: '100 m',
+              value: '2'
+            }
+          ])
+        })
+      })
+
+      describe('and the "alertType" is "resume" ', () => {
```
Cruikshanks, line 185: TL;DR; happy for "warning" and "resume" to be combined into "and the alert type is not stop or reduce".

I'm fine in cases like this where we do something if `x`, else do if not `x` do something else to say that in the test description and just use have one test that is not `x`.

For example, if there were 30 different threshold types, and we just do something special for a couple of them, I wouldn't expect to see tests for the other 28.

Cruikshanks (changes requested):

app/presenters/notices/setup/abstraction-alerts/alert-thresholds.presenter.js
```diff
@@ -19,12 +19,30 @@ function go(session) {
...
+ *
+ * When the alert type is not 'stop' or 'reduce', all stations with any restriction are included.
+ *
+ * @private
+ */
+function _relevantLicenceMonitoringStations(licenceMonitoringStations, alertType) {
+  if (alertType !== 'stop' && alertType !== 'reduce') {
+    return licenceMonitoringStations
+  }
+  return licenceMonitoringStations.filter((licenceMonitoringStation) => {
```
Cruikshanks, line 37: My bad!

```suggestion
  }

  return licenceMonitoringStations.filter((licenceMonitoringStation) => {
```

Cruikshanks (changes requested):

Cruikshanks (approved):

PR #1957, opened 2025-05-07
Update notifications setup abstraction alerts fetch / determine session data

jonathangoulding:

There was a previous assumption that we needed to do a raw query to get the licence monitoring stations. This is not the case as the table and data we need already exist in the monitoring station code.

This change updates the fetch licence mounting station to use a similar fetch to the one that already exists (with some columns removed).

This change also updates the determine licence monitoring stations service to use the newly fetched data.

### DIFF
--- a/app/services/notices/setup/abstraction-alerts/fetch-licence-monitoring-stations.service.js
+++ b/app/services/notices/setup/abstraction-alerts/fetch-licence-monitoring-stations.service.js
@@ -5,104 +5,64 @@
  * @module FetchLicenceMonitoringStationsService
  */
 
-const { db } = require('../../../../../db/db.js')
+const MonitoringStationModel = require('../../../../models/monitoring-station.model.js')
 
 /**
  * Fetches the data needed for the monitoring station abstraction alert journey
  *
- * @param {string} id
- * @returns {Promise<object[]>}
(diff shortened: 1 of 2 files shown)


app/services/notices/setup/abstraction-alerts/fetch-licence-monitoring-stations.service.js
Cruikshanks, line 1: As this is actually fetching an instance of `MonitoringStationModel`, it should be called `fetch-monitoring-station.service.js`. And if its just a few extra columns that separate this from the other, I'm okay with the other being reused.

I think when I suggested creating a separate version, it was with the intent that the service was returning an array of `LicenceMonitoringStationModels`. But this would mean you need to get the label via another route.

So, if it's just easier to use the existing fetch, go right ahead. 🤷

Cruikshanks (changes requested):
Just need to align the name of the fetch service with what it is actually returning.

That's my only NIT!

app/services/notices/setup/abstraction-alerts/fetch-licence-monitoring-stations.service.js
jonathangoulding: Updated to FetchMonitoringStationService

PR #1815, opened 2025-03-17
Implement Confirm button on Check Return Details page

StuAA78:
https://eaflood.atlassian.net/browse/WATER-4847

PR #1959, opened 2025-05-07
Create `CreateReturnSubmissionService`

StuAA78:
https://eaflood.atlassian.net/browse/WATER-4847

Broken out of PR https://github.com/DEFRA/water-abstraction-system/pull/1815

This PR implements `CreateReturnSubmissionService `. The service creates a new return submission in the db for the given return log ID and using the data passed to it.

The version number will be 1 if no previous submission exists for this return log ID; otherwise, the version number will be incremented by 1. If a previous submission does exist then it will be marked as superseded by setting its `current` property to `false`.


app/services/return-logs/setup/create-return-submission.service.js
```diff
@@ -0,0 +1,69 @@
...
+ * ID; otherwise, the version number will be incremented by 1
+ *
+ * If a previous submission does exist then it will be marked as superseded by setting its `current` property to false
+ *
+ * @param {string} returnLogId - ID of the return log (typically something like v1:6:01/23:987654:2024-11-01:2025-10-31)
+ * @param {string} userId - ID of the user creating the return submission (typically an email address)
+ * @param {string} userType - Type of user (eg. 'internal' or 'external')
+ * @param {object} metadata - Metadata for the return submission
+ * @param {boolean} nilReturn - Indicates if the return is a nil return
+ * @param {objection.transaction} [trx] - Optional transaction object
```
Cruikshanks, line 22: We've been sticking primitive types in the JSDoc comments since we started linting and building the docs.

So, perhaps this if you are trying to be explicit about what it is...

```suggestion
 * @param {object} [trx=null] - Optional {@link https://vincit.github.io/objection.js/guide/transactions.html#transactions | transaction object}
```

app/services/return-logs/setup/create-return-submission.service.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  if (previousVersion) {
+    await _markPreviousVersionAsSuperseded(returnLogId, previousVersion, trx)
+  }
+
+  return ReturnSubmissionModel.query(trx).insert(returnSubmission)
+}
+
+async function _getVersionNumber(returnLogId, trx) {
```
Cruikshanks, line 48: Admittedly, it's a personal opinion from my Ruby days, but I encourage folks to drop prefixing methods with `get` and `set`.

In this case, we're doing more than just getting the version number: we're working it out and returning both it and the previous number.

So, might I suggest ...

```suggestion
async function _determineVersionNumbers(returnLogId, trx) {
```

app/services/return-logs/setup/create-return-submission.service.js
```diff
@@ -0,0 +1,69 @@
...
+  const { previousVersion } = await ReturnSubmissionModel.query(trx)
+    .max('version as previousVersion')
+    .whereLike('returnLogId', `${returnLogId}%`)
```
Cruikshanks, line 51: This should be a standard `where`. There shouldn't ever be a scenario where we have received only a partial return log and/or are using it to identify the corresponding return submissions in this context.

```suggestion
    .where('returnLogId', returnLogId)
```

app/services/return-logs/setup/create-return-submission.service.js
```diff
@@ -0,0 +1,69 @@
...
+  return {
+    previousVersion: previousVersion || null,
+    version: previousVersion + 1 || 1
+  }
+}
+
+async function _markPreviousVersionAsSuperseded(returnLogId, version, trx) {
+  await ReturnSubmissionModel.query(trx)
+    .patch({ current: false })
+    .whereLike('returnLogId', `${returnLogId}%`)
```
Cruikshanks, line 63:
```suggestion
    .where('returnLogId', returnLogId)
```

app/services/return-logs/setup/create-return-submission.service.js
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Creates a new return submission. The version number will be 1 if no previous submission exists for this return log
+ * ID; otherwise, the version number will be incremented by 1
+ *
+ * If a previous submission does exist then it will be marked as superseded by setting its `current` property to false
+ *
+ * @param {string} returnLogId - ID of the return log (typically something like v1:6:01/23:987654:2024-11-01:2025-10-31)
+ * @param {string} userId - ID of the user creating the return submission (typically an email address)
+ * @param {string} userType - Type of user (eg. 'internal' or 'external')
```
Cruikshanks, line 19: I think you can skip this param. In this context, it will always be `internal`, so it can be defined within this service.

test/services/return-logs/setup/create-return-submission.service.test.js
```diff
@@ -0,0 +1,72 @@
...
+        const previousVersion = await ReturnSubmissionModel.query()
+          .where('returnLogId', returnLogId)
+          .where('version', 1)
+          .first()
+
+        expect(previousVersion.current).to.be.false()
+      })
+    })
+  })
+})
```
Cruikshanks, line 72: I've made a series of suggested changes here. It was easier to suggest the change as a single thing, but I'll try and break down what they are.

- **Main describe** - I know we haven't been consistent in `/return-logs`, but it's on the TODO list!
- **Declaring values at the top** - This is more about getting the call to `ReturnSubmissionHelper.defaults` out of the `it()` blocks in order to stick to the 'Arrange, Act, Assert' test pattern. Therefore, if the call was being pushed into `beforeEach()`, we'd need to declare some variables so they can be accessed in the `it()` blocks. And if we're doing that, we may as well declare them as `const`s at the top with values we can see. `returnLogId` we do need to treat a little differently, hence it becomes a `let`. I opted to bring in `ReturnLogHelper` so we could see we are using its `generateReturnLogId()` method explicitly, rather than the implicitly via `defaults`. But that's not a hill I'm prepared to die on!
- **Test dropped** I've dropped the top-level separate `can create an instance` test as this can be incorporated into the other two. This follows our attempts to limit interactions with the DB to a minimum, where possible.
- **Equal over include** - I figured this was a more explicit test of what we get back and would more likely fail and expose an issue should that change.
- **InstanceOf** - Where a service declares that we are returning an instance of something, I try to include an assertion of that (if I remember!)

Hope that all makes sense!

```suggestion
const ReturnLogHelper = require('../../../support/helpers/return-log.helper.js')
const ReturnSubmissionHelper = require('../../../support/helpers/return-submission.helper.js')
const ReturnSubmissionModel = require('../../../../app/models/return-submission.model.js')

// Thing under test
const CreateReturnSubmissionService = require('../../../../app/services/return-logs/setup/create-return-submission.service.js')

describe('Return Logs - Setup - Create Return Submission service', () => {
  const userId = 'admin-internal@wrls.gov.uk'
  const userType = 'internal'
  const metadata = {}
  const nilReturn = false

  let returnLogId

  describe('when called with valid data', () => {
    describe('and no previous submission exists for this return log', () => {
      beforeEach(() => {
        returnLogId = ReturnLogHelper.generateReturnLogId()
      })

      it('creates a new return submission and sets the version to 1', async () => {
        const result = await CreateReturnSubmissionService.go(returnLogId, userId, userType, metadata, nilReturn)

        expect(result).to.equal(
          {
            returnLogId,
            userId,
            userType,
            version: 1,
            metadata,
            nilReturn,
            current: true
          },
          { skip: ['id', 'createdAt'] }
        )
        expect(result).to.be.instanceOf(ReturnSubmissionModel)
      })
    })

    describe('and a previous submission exists for this return log', () => {
      beforeEach(async () => {
        const returnSubmission = await ReturnSubmissionHelper.add()

        returnLogId = returnSubmission.returnLogId
      })

      it('creates a new return submission and sets the version to 2', async () => {
        const result = await CreateReturnSubmissionService.go(returnLogId, userId, userType, metadata, nilReturn)

        expect(result).to.equal(
          {
            returnLogId,
            userId,
            userType,
            version: 2,
            metadata,
            nilReturn,
            current: true
          },
          { skip: ['id', 'createdAt'] }
        )
        expect(result).to.be.instanceOf(ReturnSubmissionModel)
      })

      it('marks the previous version as superseded', async () => {
        await CreateReturnSubmissionService.go(returnLogId, userId, userType, metadata, nilReturn)

        const previousVersion = await ReturnSubmissionModel.query()
          .where('returnLogId', returnLogId)
          .where('version', 1)
          .first()

        expect(previousVersion.current).to.be.false()
      })
    })
  })
})
```

Cruikshanks (changes requested):
Give me a ping if this raises any questions, issues or comments!

app/services/return-logs/setup/create-return-submission.service.js
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @param {string} returnLogId - ID of the return log (typically something like v1:6:01/23:987654:2024-11-01:2025-10-31)
+ * @param {string} userId - ID of the user creating the return submission (typically an email address)
+ * @param {object} metadata - Metadata for the return submission
+ * @param {boolean} nilReturn - Indicates if the return is a nil return
+ * @param {object} [trx=null] - Optional {@link https://vincit.github.io/objection.js/guide/transactions.html#transactions | transaction object}
+ *
+ * @returns {Promise<module:ReturnSubmissionModel>} - The created return submission
+ */
+async function go(returnLogId, userId, metadata, nilReturn, trx = null) {
```
Jozzey, line 25: From just looking at this PR I'm not really sure why `trx = null` is there. Is it possible to do a unit test that includes a transaction just to clarify?

Cruikshanks (dismissed):

app/services/return-logs/setup/create-return-submission.service.js
StuAA78: Good shout -- I've added tests to his plus the other PR to verify that the `trx` is correctly used so if there are any errors then nothing is persisted in the db (which is why it's there -- the orchestrating service cals each of the services within a transaction so we don't end up with some stuff persisted and some not if it errors)

Cruikshanks (approved):

PR #1982, opened 2025-05-13
Update notifications setup abstraction alert types error message

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5033

When a user selects an alert type that is not warning or resume. We need to check that the selected alert type is available as a restriction type to send the alert notification.

This change updates the alert type submit service and validator to check if the selected alert type is available.


jonathangoulding:
I honestly don't quite understand how the 'warning' and 'resume' should at least be there. But this is best effort from my understanding.

Cruikshanks (approved):

PR #1903, opened 2025-04-22
Amend validation for returns marked quarterly submissions

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-4955

While working on a [bug fix](https://github.com/DEFRA/water-abstraction-system/pull/1900) for the returns journey check page, a problem was identified with the current validation. Other check services within WRLS do not currently involve validation as they are meant to only contain validated data from the user. The returns check page breaks this and involves validation to prevent requirements marked summer and returns checked for quarterly submission.

In response to this, we have decided to incorporate validation into both the 'additional submission options' and 'returns cycle' page. This will prevent users from attaining incorrect data onto the check page, removing unnecessary validation.

This PR will add validation to the 'additional submission options' and 'returns cycle' pages to prevent requirements marked summer and returns checked for quarterly submission.

PR #1983, opened 2025-05-13
Missing validation for quarterly return & summer combination

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-5052

During [Amend validation for returns marked quarterly submissions](https://github.com/DEFRA/water-abstraction-system/pull/1903), a fix was added to prevent requirements for returns with a combination of summer and marked quarterly return submission. While being tested by the QA team, it was noticed that the current implementation does not take into account licences marked as water undertakers.

When a licence is an undertaker, the setup automatically sets quarterly return submissions to true, allows invalid data to be submitted. In order to fix this, when a licence is a water undertaker the user is prevented from selecting `summer` on the returns cycle page.

This PR adds additional validation to the return cycle page to prevent licences marked as water undertakers from submitting a combination of summer return cycle and quarterly return submission.


app/validators/return-versions/setup/returns-cycle.validator.js
```diff
@@ -46,9 +46,8 @@ function _noSummerCycleWithQuarterlyReturns(value, helpers, session) {
 
   const isSummer = returnsCycle === 'summer'
   const hasQuarterlyReturns = session.data.quarterlyReturns === true
-  const checkPageVisited = session.data.checkPageVisited === true
 
-  if (checkPageVisited && hasQuarterlyReturns && isSummer) {
+  if (hasQuarterlyReturns && isSummer) {
```
rvsiyad, line 50: By removing the `checkPageVisited` conditional, we are able to compare if the user has submitted `summer` for the return cycle and if the licence is marked for quarterly return submission. Since the quarterly submission flag is automatically set for water undertaker licences—regardless of user input—this ensures that validation works.

PR #1988, opened 2025-05-14
Add notifications setup abstraction check licence remove service

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5028

We have changed the legacy check licence matches page to allow individual thresholds / licence monitoring stations to be removed.

This change adds the remove threshold service. It uses the licence monitoring station id to remove the threshold from the list.

This is not removed from the session but is instead applied on top of the existing list similar the current filter for relevant threshold to display.


jonathangoulding:
We will add the notification / flash in another change.

rvsiyad (approved):

PR #2006, opened 2025-05-19
Fix meter readings not correctly being persisted on submission

StuAA78:
https://eaflood.atlassian.net/browse/WATER-4847

During QA it was found that volumes were not being persisted for meter readings. On investigation, it was found that although we calculate volumes from meter readings for display, these calculated volumes aren't put in the session. This was overlooked as when the session is initialised from a previous return, we do populate it with volumes.

We fix this by ensuring that when the user submits a return, we re-calculate volumes based on meter readings prior to persisting the lines.


app/services/return-logs/setup/create-return-lines.service.js
```diff
@@ -44,6 +66,36 @@ async function go(lines, returnSubmissionId, returnsFrequency, units, meterProvi
...
+ * quantity directly. If it contains meter readings, it calculates the difference between the current and previous
+ * readings. If the meter has a 10x display, it multiplies the difference by 10.
+ *
+ * @param {object} line - The line to process
+ * @param {number} previousReading - The last meter reading we encountered
+ * @param {boolean} volumes - Indicates if the line contains volumes (vs meter readings)
+ * @param {boolean} meter10TimesDisplay - Indicates if the meter is a 10x display
+ * @returns {object} - An object containing the calculated quantity and the new previous reading
+ */
+function calculateLineQuantity(line, previousReading, volumes, meter10TimesDisplay) {
```
Jozzey, line 80:
```suggestion
function _calculateLineQuantity(line, previousReading, volumes, meter10TimesDisplay) {
```

Jozzey (changes requested):
Just one little thing

PR #2013, opened 2025-05-21
Tag a licence - Enter licence number

StuAA78:
https://eaflood.atlassian.net/browse/WATER-4995

This PR adds the "Enter licence number" page as part of the tag a licence flow.

Note that until the next page is implemented, the Continue button will result in a 404.


Jozzey (changes requested):
Just one tiny thing

Jozzey (approved):

PR #2014, opened 2025-05-21
Spike to see the impact of abstraction alerts on recipients

jonathangoulding:


app/lib/notify-templates.lib.js
```diff
@@ -20,6 +20,24 @@ const notifyTemplates = {
       returnsAgentEmail: '41c45bd4-8225-4d7e-a175-b48b613b5510',
       licenceHolderLetter: '4fe80aed-c5dd-44c3-9044-d0289d635019',
       returnsToLetter: '0e535549-99a2-44a9-84a7-589b12d00879'
+    },
+    'abstraction-alert': {
+      reduceWarning: '27499bbd-e854-4f13-884e-30e0894526b6',
```
jonathangoulding, line 25: All of these will require a template match.

app/presenters/notices/setup/create-notice.presenter.js
```diff
@@ -25,14 +25,14 @@ function go(session, recipients, auth) {
 
   return {
     issuer: auth.credentials.user.username,
-    licences: _licences(recipients),
+    licences: session.journey !== 'abstraction-alert' ? _licences(recipients) : null,
```
jonathangoulding, line 28: This needs to be rejigged to allow for both types of message

app/presenters/notices/setup/notifications.presenter.js
```diff
@@ -177,9 +177,10 @@ function _letterTemplate(contactType, journey) {
 
 function _returnsPeriod(returnsPeriod) {
   return {
-    periodEndDate: formatLongDate(new Date(returnsPeriod.endDate)),
-    periodStartDate: formatLongDate(new Date(returnsPeriod.startDate)),
-    returnDueDate: formatLongDate(new Date(returnsPeriod.dueDate))
+    // periodEndDate: formatLongDate(new Date(returnsPeriod.endDate)),
+    // periodStartDate: formatLongDate(new Date(returnsPeriod.startDate)),
+    // returnDueDate: formatLongDate(new Date(returnsPeriod.dueDate))
+    // flow_or_level, monitoring_station_name, licence_ref, issuer_email_address, source, threshold_value, threshold_unit, condition_text
```
jonathangoulding, line 183: This is the most required template structure. 

What is condition text ?

jonathangoulding: To get the thresholds we need to use the same logic we had previously to determine the thresholds to use. Maybe lift the logic out of the presenter into a determine service.

app/presenters/notices/setup/notifications.presenter.js
```diff
@@ -243,6 +244,15 @@ function _messageRef(journey, messageType, contactType) {
         both: 'ad-hoc_licence_holder_letter',
         'Returns to': 'ad-hoc_returns_to_letter'
       }
+    },
+    'abstraction-alert': {
```
jonathangoulding, line 248: check these.

PR #2017, opened 2025-05-21
Add notices abstraction alert notification presenter

jonathangoulding:

We previously implemented the notices journey based on the use of return periods, with the intention of adding abstraction alerts later down the line.

That time has come and we need to adjust some of the existing structures in place to allow us reuse alot of the complex logic.

To do this we will be using the session.journey to trigger any alternative logic to the returns periods.

This means passing the session object around rather than the previous attempt of restricting the session to the submit service.

This change adds a presenter to handle the abstraction alerts. This will take the 'relevant' licence monitoring stations from the session and create a notification for any recipient associated with that licence / licence monitoring station.

This could / will result in multiple recipient receiving the same notification, as well as a recipient receiving a notification per licence monitoring station / threshold.


app/lib/notify-templates.lib.js
```diff
@@ -21,6 +21,22 @@ const notifyTemplates = {
       licenceHolderLetter: '4fe80aed-c5dd-44c3-9044-d0289d635019',
       returnsToLetter: '0e535549-99a2-44a9-84a7-589b12d00879'
     }
+  },
+  'abstraction-alerts': {
+    reduceWarning: '27499bbd-e854-4f13-884e-30e0894526b6',
```
jonathangoulding, line 26: How do we work these out. 

There are `restriction_type` and `status` in the licence monitoring station in the db. 

Statuses in my db show - 'reduce', 'resume' and 'warning'

Do we need to look at the last type of alert sent ? 

The assumption would be: 

['chosen-alert-type']['status']

This would mean a 'reduce''reduce' = reduce ?

app/presenters/notices/setup/abstraction-alerts-notifications.presenter.js
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * @param recipients
+ * @param session
+ * @param eventId
+ */
+function go(recipients, session, eventId) {
+  const notifications = []
+
+  for (const recipient of recipients) {
+    // loop each threshold we need to use ?
```
jonathangoulding, line 22: We need to iterate over the 'relevant' licence monitoring stations to send the notification to the recipient per station.

app/presenters/notices/setup/abstraction-alerts-notifications.presenter.js
```diff
@@ -0,0 +1,121 @@
...
+const { notifyTemplates } = require('../../../lib/notify-templates.lib.js')
+const { contactName, contactAddress } = require('../../crm.presenter.js')
+
+/**
+ *
+ * @param recipients
+ * @param session
+ * @param eventId
+ */
+function go(recipients, session, eventId) {
```
jonathangoulding, line 18: Add this into service call - use the journey to trigger.

app/presenters/notices/setup/abstraction-alerts-notifications.presenter.js
```diff
@@ -0,0 +1,115 @@
...
+ * Formats recipients into notifications for an abstraction alert
+ * @module AbstractionAlertsNotificationsPresenter
+ */
+
+const { transformStringOfLicencesToArray, timestampForPostgres } = require('../../../lib/general.lib.js')
+const { notifyTemplates } = require('../../../lib/notify-templates.lib.js')
+const { contactName, contactAddress } = require('../../crm.presenter.js')
+
+/**
+ * Formats recipients into notifications for an abstraction alert
```
jonathangoulding, line 13: I will add docs when the func is finished

Cruikshanks (changes requested):
Just some NITS.

app/presenters/notices/setup/abstraction-alerts-notifications.presenter.js
```diff
@@ -0,0 +1,115 @@
...
+    })
+
+    for (const recipient of rec) {
+      if (recipient.email) {
+        notifications.push(_email(recipient, referenceCode, eventId))
+      } else {
+        notifications.push(_letter(recipient, referenceCode, eventId))
+      }
+    }
+  }
```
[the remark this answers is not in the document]
jonathangoulding: my bad

app/presenters/notices/setup/abstraction-alerts-notifications.presenter.js
```diff
@@ -0,0 +1,115 @@
...
+  return addressLines
+}
+
+function _email(recipient, referenceCode, eventId) {
+  const templateId = _emailTemplate()
+
+  const messageType = 'email'
+
+  return {
+    ..._common(referenceCode, templateId, eventId),
```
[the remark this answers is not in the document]
jonathangoulding: done.

Cruikshanks (approved):

PR #2018, opened 2025-05-21
Add notices determines relevant licence monitoring stations

jonathangoulding:

We previously added functionality to save the licence refs to the session in the 'SubmitCheckLicenceMatchesService'. This was to make the code simpler on the recipients part of the journey.

We have found another use case of adding the 'relevant' licence monitoring stations to the session.

This change adds a new session object 'relevantLicenceMonitoringStations' to the session for the 'abstraction-alert' journey. This will be used by the recipient to send notifications to the stations recipients.

As this has complex functionality and needed in two places it has been added / tested as its own service.

A few tweaks to the wording has been made to better align with what it is actually doing. They are not thresholds but licence monitoring stations.

### DIFF
--- a/app/services/notices/setup/abstraction-alerts/determine-relevant-licence-monitoring-stations.service.js
+++ b/app/services/notices/setup/abstraction-alerts/determine-relevant-licence-monitoring-stations.service.js
@@ -1,0 +1,43 @@
+'use strict'
+
+/**
+ * Determines relevant licence monitoring stations for the `abstraction-alerts` journey
+ *
+ * @module DetermineRelevantLicenceMonitoringStationsService
+ */
+
+/**
+ * Determines relevant licence monitoring stations for the `abstraction-alerts` journey
+ *
+ * When building the `abstraction-alerts` journey there is a starting array of 'licenceMonitoringStations'. The user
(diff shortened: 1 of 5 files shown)


app/services/notices/setup/abstraction-alerts/determine-relevant-licence-monitoring-stations.service.js
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * We keep the original array intact.
+ *
+ * @param {Array<object>} licenceMonitoringStations
+ * @param {Array<string>} selectedLicenceMonitoringStations
+ * @param {Array<string>} removedLicenceMonitoringStations
+ *
+ * @returns {Array<object>}
+ */
+function go(licenceMonitoringStations, selectedLicenceMonitoringStations, removedLicenceMonitoringStations) {
```
jonathangoulding, line 27: If possible i will go back an refactor the use of thresholds to the relevant args here. 

In another change.

jonathangoulding:
These changes have come from the work here - https://github.com/DEFRA/water-abstraction-system/pull/2017

This simplifies the need to find the relevant licence monitoring station the recipients.

Cruikshanks (approved):

PR #2029, opened 2025-05-27
Add notices abstraction alert notification personalisation

jonathangoulding:

We previously implemented the notices journey based on the use of return periods, with the intention of adding abstraction alerts later down the line.

That time has come and we need to adjust some of the existing structures in place to allow us reuse alot of the complex logic.

To do this we will be using the session.journey to trigger any alternative logic to the returns periods.

This means passing the session object around rather than the previous attempt of restricting the session to the submit service.

This change updates the existing presenter to create the personalisation for the notify templates.

They all expect the same data. Some are just structured slight different.

Some personalisation is missing and will be added in a later chang.

To make sure the emails are received the batch notification has been update in this change to allow the abstraction journey to use the correct presenter. This follows our existing strategy using the session.journey

### DIFF
--- a/app/presenters/notices/setup/abstraction-alerts-notifications.presenter.js
+++ b/app/presenters/notices/setup/abstraction-alerts-notifications.presenter.js
@@ -29,10 +29,12 @@ function go(recipients, session, eventId) {
     })
 
     for (const matchingRecipient of matchingRecipients) {
+      const commonPersonalisation = _commonPersonalisation(licenceMonitoringStation, session.monitoringStationName)
+
       if (matchingRecipient.email) {
-        notifications.push(_email(matchingRecipient, referenceCode, eventId))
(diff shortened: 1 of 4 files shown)


app/presenters/notices/setup/abstraction-alerts-notifications.presenter.js
```diff
@@ -52,7 +72,69 @@ function _addressLines(contact) {
...
+ * 'condition_text' can be empty, the templates is structured so when it's empty is does not affect the copy.
+ *
+ * In the case of a letter, the address is also required.
+ *
+ * @private
+ */
+
+function _commonPersonalisation(licenceMonitoringStation, monitoringStationName) {
+  return {
+    condition_text: '',
```
jonathangoulding, line 100: The empty parts still need to be implemented. 

This will be done in other changes.

PR #2030, opened 2025-05-27
Update paginator presenter to use URLSearchParams

jonathangoulding:
Use URLSearchParams for proper query encoding

Replaced manual query string construction with the built-in URLSearchParams API.

This change improves the robustness of pagination URLs by ensuring that all query parameters are properly URL-encoded.

This prevents potential issues with special characters (like spaces, slashes, ampersands) breaking the URL structure or causing unexpected behavior.


Cruikshanks (changes requested):
Just a NIT

Cruikshanks (approved):

PR #2034, opened 2025-05-29
Update abstraction alert email address - Submit

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-5058

Previously we added the logic for viewing the abstraction alert email address page. We now have to add the logic to validation user input and saving it to the session.

This PR adds the submit logic for the abstraction alert email address.


app/services/notices/setup/abstraction-alerts/submit-alert-email-address.service.js
```diff
@@ -49,10 +57,12 @@ function _validate(payload) {
     return null
   }
 
-  const { message } = validation.error.details[0]
+  const { message, path } = validation.error.details[0]
 
   return {
-    text: message
+    text: message,
+    radioFormElement: path[0] === 'alertEmailAddress' ? { text: message } : null,
```
jonathangoulding, line 64: path[0] ? 

Needs a better name or explanation. 

Should be another way to do it if they are both the same for different thing.

app/validators/notices/setup/abstraction-alerts/alert-email-address.validator.js
```diff
@@ -17,14 +23,23 @@ const Joi = require('joi')
  * any errors are found the `error:` property will also exist detailing what the issues were
  */
 function go(payload) {
-  const errorMessage = 'Select an email address to include in the alerts'
-
   const schema = Joi.object({
-    alertEmailAddress: Joi.string().required().messages({
-      'any.required': errorMessage
+    alertEmailAddress: Joi.string()
```
jonathangoulding, line 27: .label('otherUser')

jonathangoulding (approved):

PR #2044, opened 2025-06-02
Update determine relevant licence monitoring stations

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5082

There was a previous assumption that the restriction type "stop_or_reduce" made a station relevant for both the "stop" and "reduce" alert types. This assumption was incorrect.

When the alert type is "stop", only stations with a restrictionType of "stop" should be considered.

When the alert type is "reduce", stations with a restrictionType of "reduce" or "stop_or_reduce" are relevant.

For other alert types ("warning", "resume"), all stations should be included.

This change corrects the logic to accurately determine the relevant licence monitoring stations for the selected alert type.

It also streamlines the process by consolidating the filtering logic into a single, cohesive function/ service.


jonathangoulding:
blocked by https://github.com/DEFRA/water-abstraction-system/pull/2033

PR #2046, opened 2025-06-03
Tag a licence - Licence conditions

StuAA78:
https://eaflood.atlassian.net/browse/WATER-4996

The next page in the tag a licence journey enables the user to select a condition on their licence. Of note are the following points:

- We amend `SubmitLicenceNumberService` (the previous page) to save the licence number to the session, as we overlooked this.
- If the licence has conditions then we display a set of radio buttons, one per condition. If the user selects one of these, they are directed to the next page in the journey -- since this is the last page, this means they go to the `/check` page.
- There is also a "condition not listed" radio button. If the user selects this, they are directed to the `/abstraction-period` page which, when developed, will allow them to manually enter the abstraction period.
- If the licence has no conditions then no radio buttons are displayed. Instead, we show explanatory text; when the user clicks Continue, they are taken to the `/abstraction-period` page.
- The legacy code filtered out any conditions which had an empty `note` field. We believe this may be the reason behind some conditions not being shown when users otherwise expected them to be. We therefore do not filter such conditions. **This means that some attention may need to be paid to notifications which expect the `note` field to be present.**


app/services/licence-monitoring-station/setup/submit-full-condition.service.js
```diff
@@ -0,0 +1,95 @@
...
+    ...pageData
+  }
+}
+
+async function _fetchAbstractionPeriod(conditionId) {
+  if (conditionId === 'not_listed') {
+    return null
+  }
+
+  return LicenceVersionPurposeConditionModel.query()
```
Jozzey, line 55: Do we need to worry about extracting the abstraction period from the linked condition? From looking at existing data in the `licence_monitoring_stations` view it appears that if linked to a condition then the abstraction period is not saved in this table. So there should be no need to add it to the session when linked to a condition?

Jozzey (changes requested):

test/services/licence-monitoring-station/setup/full-condition.service.test.js
```diff
@@ -0,0 +1,87 @@
...
+
+const CES_CONDITION_TYPE = LicenceVersionPurposeConditionTypeHelper.select(20)
+
+describe('Full Condition Service', () => {
+  let condition
+  let licence
+  let session
+
+  beforeEach(async () => {
+    licence = await LicenceHelper.add()
```
Jozzey, line 30: One of the reasons why we like to pull any significant fetch queries out of the main services is to make testing the main services simpler as we can just stub the "fetch" services.

For consistency I'd suggest that rather than using the helpers to actually create the test data, that the fetch service is stubbed instead. This is generally how we would do this, here's an example `test/services/monitoring-stations/view.service.test.js`

Jozzey (changes requested):
Just 1 little thing

PR #2051, opened 2025-06-05
Fix abstraction alerts download recipients

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5065

We have previously implemented the download and check recipient functionality. A bug has been found in on both the check page and downloaded csv.

The issues occur when a recipient is for multiple licences. We did not account for this in the original code (we thought we did, but this change proves we did not).

When a recipient belongs to multiple licences, we receive the 'licence_refs' in a comma seperated string: 'licenceOne,licenceTwo'. We were not accounting for this at all.

This change adds logic to check if the recipients licence refs (think of them as an array now) includes the relevant licence monitoring stations licence ref.

If this array does then it is a recipient.

We also found an issue with duplicate licence refs when fetching the recipients. This was due to multiple additional contacts not being distinct. This has also been fixed.


app/services/notices/setup/fetch-abstraction-alert-recipients.service.js
```diff
@@ -120,11 +120,11 @@ function _query() {
     WHERE
       ldh.licence_ref = ANY (?)
     UNION ALL
-    SELECT
+    SELECT DISTINCT
```
jonathangoulding, line 123: I am not 100% on why this works. 

Could do with a discussion to clarify.

rvsiyad (approved):

PR #2070, opened 2025-06-10
Update abstraction alert thresholds to be ordered

rvsiyad:
https://eaflood.atlassian.net/browse/WATER-5094

Previously we added the logic to display abstraction alert threshold values. After comparing against the legacy journey, we noticed that the thresholds values where being sorted into a specific order. As a result, we want to replicate this sorted order within the new abstraction alerts threshold page.

This PR will display the abstraction alert thresholds in sorted order.


app/presenters/notices/setup/abstraction-alerts/alert-thresholds.presenter.js
```diff
@@ -58,12 +116,14 @@ function _thresholdOptions(licenceMonitoringStations, alertType, alertThresholds
...
+  const sortedThresholds = _sortedThresholds(relevantThresholds)
+
+  return sortedThresholds.map((threshold) => {
+    const [measureType, thresholdValue, thresholdUnit] = threshold.split('-')
 
     return {
-      checked: alertThresholds.includes(relevantThreshold),
-      value: relevantThreshold,
+      checked: alertThresholds.includes(threshold),
+      value: threshold,
```
jonathangoulding, line 126: this could be better named as thresholdGroup 

i should have named it this initially.

PR #2084, opened 2025-06-13
Add notices ad hoc select notice type scaffold

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5011

This change introduced the scaffolded code for the notices setup ad-hoc select notice type page.


jonathangoulding:
ignore this - 



this should be

PR #2094, opened 2025-06-17
Update notices abstraction alerts status update

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5100

We have a requirement to update the licence monitoring stations 'Last type of alert sent and date issued'. To do this and keep the exising notification code / logic generic we need to add the licence monitoring station id and the users chosen alert type into the personalisation.

This will not be shown in the template. We initially built the notices functionality around the legacy code. This mean we store the personalisation data as json for each notice we send.

This change updates the notification job to check if the notifications id for an abstraction alert. When it is we update licence monitoring stations 'status_updated_at' and 'status' fields.

This is known as the 'Last alert sent'

### DIFF
--- a/app/services/jobs/notifications/update-abstraction-alerts.service.js
+++ b/app/services/jobs/notifications/update-abstraction-alerts.service.js
@@ -19,7 +19,7 @@ const { timestampForPostgres } = require('../../../lib/general.lib.js')
  */
 async function go(notifications) {
   for (const notification of notifications) {
-    if (notification.messageRef?.includes('water_abstraction_alert'))
+    if (notification.messageRef?.includes('water_abstraction_alert') && notification.status !== 'error')
       await _update(notification.personalisation.licenceMonitoringStationId, notification.personalisation.alertType)
   }
 }
--- a/app/services/jobs/notifications/update-abstraction-alerts.service.js
+++ b/app/services/jobs/notifications/update-abstraction-alerts.service.js
@@ -5,8 +5,8 @@
  * @module UpdateAbstractionAlertsService
  */
 
-const LicenceMonitoringStationModel = require('../../../models/licence-monitoring-station.model.js')
 const { timestampForPostgres } = require('../../../lib/general.lib.js')
+const LicenceMonitoringStationModel = require('../../../models/licence-monitoring-station.model.js')
 
 /**
  * Orchestrates the process of updating the licence monitoring stations last abstraction alert.
--- a/app/services/jobs/notifications/update-abstraction-alerts.service.js
+++ b/app/services/jobs/notifications/update-abstraction-alerts.service.js
@@ -1,0 +1,36 @@
+'use strict'
+
+/**
+ * Orchestrates the process of updating the licence monitoring stations last abstraction alert.
+ * @module UpdateAbstractionAlertsService
+ */
+
+const LicenceMonitoringStationModel = require('../../../models/licence-monitoring-station.model.js')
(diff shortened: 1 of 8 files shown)


app/services/jobs/notifications/update-abstraction-alerts.service.js
```diff
@@ -0,0 +1,51 @@
+'use strict'
+
+/**
+ * Orchestrates the process of updating the licence monitoring stations last abstraction alert.
+ * @module UpdateAbstractionAlertsService
+ */
+
+const LicenceMonitoringStationModel = require('../../../models/licence-monitoring-station.model.js')
+const { timestampForPostgres } = require('../../../lib/general.lib.js')
```
Cruikshanks, line 9: NIT

```suggestion
const { timestampForPostgres } = require('../../../lib/general.lib.js')
const LicenceMonitoringStationModel = require('../../../models/licence-monitoring-station.model.js')
```

app/services/jobs/notifications/update-abstraction-alerts.service.js
```diff
@@ -0,0 +1,51 @@
...
+ */
+function _stations(notifications) {
+  const stationsToUpdate = notifications.filter((notification) => {
+    return notification.messageRef?.includes('water_abstraction_alert') && notification.status !== 'error'
+  })
+
+  return stationsToUpdate.map((notification) => {
+    return _update(notification.personalisation.licenceMonitoringStationId, notification.personalisation.alertType)
+  })
+}
```
Cruikshanks, line 47: NIT

```suggestion
/**
 * Determine which notifications are for the water abstraction alerts and have not failed to send.
 * @private
 */
function _stations(notifications) {
  const stationsToUpdate = notifications.filter((notification) => {
    return notification.messageRef?.includes('water_abstraction_alert') && notification.status !== 'error'
  })

  return stationsToUpdate.map((notification) => {
    return _update(notification.personalisation.licenceMonitoringStationId, notification.personalisation.alertType)
  })
}

async function _update(id, status) {
  await LicenceMonitoringStationModel.query().findById(id).patch({
    status,
    statusUpdatedAt: timestampForPostgres()
  })
}
```

test/services/jobs/notifications/notifications-status-updates.service.test.js
```diff
@@ -37,6 +39,10 @@ describe('Job - Notifications - Process notifications status updates service', (
     // correctly, we do not want increase the timeout for the test as we want them to fail if a timeout occurs
     Sinon.stub(NotifyConfig, 'delay').value(ONE_HUNDRED_MILLISECONDS)
 
+    if (NotifyClient.prototype.getNotificationById.restore) {
+      NotifyClient.prototype.getNotificationById.restore()
+    }
```
Cruikshanks, line 44: I'm guessing this is here for a reason, and it's related to `Sinon.restore()` not functioning as expected.

But it's a guess. Hence, this could really do with a `// NOTE: `.

Cruikshanks (changes requested):
Just NITS

test/services/jobs/notifications/notifications-status-updates.service.test.js
jonathangoulding: Yes. 

When we remove the notify stubs toggle this will be needed in all the tests. Unless we leave the functions as they where are remove the toggle.

jonathangoulding: Cancel my last. 

It works without this. But i can not explain why it didn't before.

PR #2102, opened 2025-06-19
Fix permissions for creating & viewing return ver.

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-5104

Don't ask me how we've ended up here, but we have the permissions the wrong way round!

Viewing a return is scoped to only those with the `billing` role, whereas we never put any auth on the buttons. We think it is because to access the 'Licence setup' tab, you need the role `view_charge_versions`, and we assumed that if you can see the tab, it's because you need to administer stuff on it.

However, dipping back into the legacy version, and our own code, I can see that they differentiate between viewing 'licence setup' (Charge information on the legacy page) and editing stuff in it.

We need to check the roles of the user, specifically whether they have `billing` as part of determining whether the return version setup buttons are shown.

WE also need to change the scope on viewing a return version to be `view_charge_versions`. This aligns with being able to access the licence setup tab and viewing charge versions.

### DIFF
--- a/app/presenters/licences/view-licence-purposes.presenter.js
+++ b/app/presenters/licences/view-licence-purposes.presenter.js
@@ -57,7 +57,7 @@ function _formatAbstractionAmounts(licenceVersionPurpose) {
   }
 
   if (instantQuantity) {
-    details.push(`${parseFloat(instantQuantity).toFixed(2)} cubic metres per second`)
+    details.push(`${parseFloat(instantQuantity).toFixed(2)} litres per second`)
   }
 
   return details
--- a/app/views/licences/tabs/summary.njk
+++ b/app/views/licences/tabs/summary.njk
@@ -162,7 +162,7 @@
       <dd class="govuk-summary-list__value">
         {% if purposesCount > 1 %}
           <div>Multiple abstractions</div>
-          <a href="/licences/{{ documentId }}/purposes">View details of the amounts</a>
+          <a href="/system/licences/{{ licenceId }}/purposes">View details of the amounts</a>
         {% else %}
           <ul class="govuk-list govuk-!-margin-bottom-0">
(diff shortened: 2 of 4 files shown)


test/presenters/licences/view-licence-set-up.presenter.test.js
```diff
@@ -990,6 +990,52 @@ describe('View Licence Set Up presenter', () => {
...
+            credentials: {
+              user: { id: 123 },
+              roles: ['view_charge_versions'],
+              groups: [],
+              scope: ['view_charge_versions'],
+              permissions: { abstractionReform: false, billRuns: true, manage: true }
+            }
+          }
+          describe('and the "enableRequirementsForReturns" feature toggle is true', () => {
+            it('return the associated links', () => {
```
Cruikshanks, line 1006:
```suggestion
            it('returns no return version setup links', () => {
```

test/presenters/licences/view-licence-set-up.presenter.test.js
```diff
@@ -990,6 +990,52 @@ describe('View Licence Set Up presenter', () => {
...
+              expect(result.links.returnVersions).to.equal({})
+            })
+          })
+
+          describe('and the "enableRequirementsForReturns" feature toggle is false', () => {
+            beforeEach(() => {
+              Sinon.stub(FeatureFlagsConfig, 'enableRequirementsForReturns').value(false)
+            })
+
+            it('return no returnVersions links', () => {
```
Cruikshanks, line 1025:
```suggestion
            it('returns no return version setup links', () => {
```

Cruikshanks (commented):
It is my own PR, so I can't approve or reject it.

Just got a couple of NITS. Once those are done, please could you 'approve' all my hard work!? 😁

robertparkinson (approved):

PR #2105, opened 2025-06-19
Notices - check recipients - preview

Jozzey:
https://eaflood.atlassian.net/browse/WATER-5066

Add functionality to allow the user to preview a notification before it is sent to the recipient.


app/presenters/notices/setup/preview.presenter.js
```diff
@@ -0,0 +1,70 @@
...
+    caption: _caption(licences),
+    contents: await _notifyPreview(personalisation, templateId),
+    messageType,
+    pageTitle: titleCase(messageRef.replace(/_/g, ' '))
+  }
+}
+
+function _address(personalisation) {
+  const addressLines = [
+    personalisation['address_line_1'],
```
jonathangoulding, line 34: Just double check this includes he name as the top line. 

The two presenters should already do this.

app/requests/notify/notify-preview.request.js
```diff
@@ -0,0 +1,51 @@
...
+ * Generate a preview template using GOV.UK Notify
+ *
+ * https://docs.notifications.service.gov.uk/node.html#generate-a-preview-template
+ *
+ * @param {string} templateId
+ * @param {object} personalisation - Optional personalisation data to be used in the template
+ *
+ * @returns {Promise<object>}
+ */
+async function send(templateId, personalisation = {}) {
```
jonathangoulding, line 20: Missing test ?

app/presenters/notices/setup/preview.presenter.js
Jozzey: It does

app/requests/notify/notify-preview.request.js
```diff
@@ -0,0 +1,51 @@
...
+  const notifyClient = NotifyClientRequest.go()
+
+  return _generatePreview(notifyClient, templateId, personalisation)
+}
+
+async function _generatePreview(notifyClient, templateId, personalisation) {
```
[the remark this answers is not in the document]
Jozzey: Currently I expect it will just go pop. I'll look into sorting out the error handling.

app/views/notices/setup/preview.njk
```diff
@@ -0,0 +1,31 @@
...
+{% block content %}
+  {# Main heading #}
+  <div class="govuk-body">
+    <span class="govuk-caption-l">{{ caption }}</span>
+    <h1 class="govuk-heading-l govuk-!-margin-bottom-3">{{ pageTitle }}</h1>
+  </div>
+
+  <div class="message-preview govuk-body">
+    <div class="govuk-!-margin-bottom-5">
+      {% if messageType === 'letter' %}
```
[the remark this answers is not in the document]
Jozzey: Thats a good spot. You're right, it should be. I'll blame it on a copy and paste error from another view 😁

Jozzey:
> We still need a licence select page ?
> 
> is that in the next change.

Following on from the chat we had. I'll hide the preview link when the notification is for an abstraction alert as they will require some kind of select page as there could be multiple notices per recipient.

robertparkinson (dismissed):

app/requests/notify/notify-preview.request.js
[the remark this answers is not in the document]
Jozzey: I've added some error handling so now if notify fails to retreive the template this is displayed in place of the preview:

Screenshot 2025-07-02 at 10 33 07

app/views/notices/setup/preview.njk
```diff
@@ -0,0 +1,35 @@
...
+{% block content %}
+  {# Main heading #}
+  <div class="govuk-body">
+    <span class="govuk-caption-l">{{ caption }}</span>
+    <h1 class="govuk-heading-l govuk-!-margin-bottom-3">{{ pageTitle }}</h1>
+  </div>
+
+  <div class="message-preview govuk-body">
+    {% if contents === 'error' %}
+      Failed to generate a preview for this notice. Click <a href="{{ refreshPageLink }}" class="govuk-link">here</a> to try again.
```
robertparkinson, line 24: this should be in a `<p></p>`

robertparkinson (changes requested):

app/views/notices/setup/preview.njk
Jozzey: Just to check I asked Copilot for it's opinion on the use of the `<p>` tag. This is what it came out with:

> It is not always necessary to use a `<p>` tag in HTML. The `<p>` tag is specifically for marking up paragraphs of text. If your content is not a paragraph (for example, headings, lists, tables, or other block elements), you should use the appropriate HTML tags for those elements.
> 
> In many cases, especially when using frameworks or templating languages (like Nunjucks in your example), you might use `<div>`, `<span>`, or other semantic tags instead of `<p>`, depending on the structure and meaning of your content. However, for blocks of plain text that form a paragraph, using `<p>` is recommended for semantic correctness and accessibility.

robertparkinson: you're example from copilot proves my point... your text is a paragraph, admittedly a small one

This text is not a heading, nor any of the other options it lists. If it was one of those other things you'd be using a `<h1></h>` or other semantic tags. 

`However, for blocks of plain text that form a paragraph, using <p> is recommended for semantic correctness and accessibility.`

Basically divs should be for sectioning parts of the page and applying classes to control how it looks, spacing, etc, on the page. the <p> tag is a semantic tag that lets screen readers and other automated tools know that the text inside of it is to be read out or allows them to skip between blocks of content with their tools. 

I'm not sure what actually gets returned from notify.. but if it's just a big block of text maybe it should be in <p> as well? 🤔 

so it should be: 
`{% if contents === 'error' %}
      <p>Failed to generate a preview for this notice. Click <a href="{{ refreshPageLink }}" class="govuk-link">here</a> to try again.</p>
{% else %}
<div class="message-preview govuk-body">
    {% if messageType === 'letter' %}
        <div class="govuk-!-margin-bottom-5">
          {{ newLineArrayItems(address) }}
        </div>
      {% endif %}

      {{ contents | markdown | safe }}
    {% endif %}
`

but if @Cruikshanks fine with it as is then ok 😄

Jozzey: Before Nunjucks works it magic on the contents from notify `{{ contents | markdown | safe }}` the raw notify output looks like this: 

```
'^ You must submit a record of your water abstraction from 1 April 2025 to 30 June 2025. \r\n' +
'\r\n' +
'^ You’ll need to submit your returns by 28 July 2025.\r\n' +
'\r\n' +
'# Why you need to submit a return\r\n' +
'\r\n' +
'You must tell us how much water you abstract to comply with your licence. We may take enforcement action if you do not submit your returns.\r\n'
```

PR #2118, opened 2025-06-23
Update notices setup notice type

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5011

We previously scaffolded the select a notice type page.

This change adds the logic to render the options for a notice type and captures the users input.

We will need to determine the notice type using the previously created 'DetermineNoticeTypeService'. This will be done in another change.


jonathangoulding:
Depends on - https://github.com/DEFRA/water-abstraction-system/pull/2114

PR #2122, opened 2025-06-24
Scaffold the notices setup check notice page

jonathangoulding:

This is the initial scaffolded code for the 'Check the notice type' page.

The page title has been added.

We have assumed the page has no requirement to check validation or do any additional processing.


Cruikshanks (approved):

app/routes/notices-setup.routes.js
```diff
@@ -233,6 +233,18 @@ const routes = [
       }
     }
   },
+  {
+    method: 'GET',
+    path: basePath + `/{sessionId}/check-notice-type`,
```
robertparkinson, line 238: can we remove the basePath as part of this PR? to standardise it with the rest of the controllers?

robertparkinson (changes requested):

PR #2124, opened 2025-06-24
Update notice setup ad-hoc routing

jonathangoulding:

We have enough of the journey built to plumb the routing all together.

This change adds the necessary change to allow the single licence journey to be traversed.


jonathangoulding:
Depends on - https://github.com/DEFRA/water-abstraction-system/pull/2122

PR #2126, opened 2025-06-24
Update the notice setup ad-hoc journey to handle paper forms

jonathangoulding:

This change does not implement paper forms.

It implements the basic changes to the notices journey to allow an individual licence to be entered and the notice type selected and the journey to be progressed.

A standard single licence will work as before.

But a paper form will 'appear' to work in the UI based on the set metadata. So it looks and feels like it works but nothing is sent.

This is required to allow the 'check the notice type' page to work and route correctly.


jonathangoulding:
depends on https://github.com/DEFRA/water-abstraction-system/pull/2124

PR #2136, opened 2025-06-25
Create select address page scaffold

robertparkinson:

As part of our work to replace the legacy ad-hoc notifications journey we will need to lookup addresses. We want to make the address lookup journey reusable across several journeys.

This PR creates the scaffold structure for the select an address page.

### DIFF
--- a/app/controllers/address.controller.js
+++ b/app/controllers/address.controller.js
@@ -30,7 +30,7 @@ async function submitPostcode(request, h) {
   return h.redirect(`/system/address/${sessionId}/select`)
 }
 
-async function select(request, h) {
+async function viewSelect(request, h) {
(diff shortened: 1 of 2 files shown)


app/controllers/address.controller.js
```diff
@@ -25,10 +27,32 @@ async function submitPostcode(request, h) {
     return h.view('address/postcode.njk', pageData)
   }
 
-  return h.redirect(`/system/address/${sessionId}/lookup`)
+  return h.redirect(`/system/address/${sessionId}/select`)
+}
+
+async function select(request, h) {
```
jonathangoulding, line 33: I'm not sure if these should be prefixed view ?

jonathangoulding (approved):

PR #2155, opened 2025-07-01
Update notice setup returns paper forms notificaiton

jonathangoulding:
Update notice setup returns paper forms notification

This change adds a notification message when the user has come from the check page and makes a change. 

If the previously selected returns do not match the currently selected returns then something has changed, we show the notification.


app/services/notices/setup/submit-returns-for-paper-forms.service.js
```diff
@@ -39,6 +45,37 @@ async function go(sessionId, payload) {
...
+  const sortedA = [...a].sort()
+  const sortedB = [...b].sort()
+
+  for (let i = 0; i < sortedA.length; i++) {
+    if (sortedA[i] !== sortedB[i]) {
+      return true
+    }
+  }
+
+  return false
```
Cruikshanks, line 76: It was the dependence on needing to `sort()` the arrays first that caught my eye.

I didn't think this was necessary, so I tried it myself. This implementation should yield the same result without requiring the arrays to be sorted first.

```javascript
function _arraysDiffer(arrayA, arrayB) {
  if (arrayA.length !== arrayB.length) {
    return true
  }

  let result = false

  for (const valueA of arrayA) {
    const match = arrayB.find((valueB) => {
      return valueA === valueB
    })

    if (!match) {
      result = true
      break
    }
  }

  return result
}
```

This is a place where performance doesn't matter. But to confirm to myself that was another reason to avoid the `sort()` calls, I put both solutions into <https://jsbench.me/>

Screenshot 2025-07-01 at 16 49 30

So, unless there's something I'm missing, I'd push for my alternate solution, or something like it.

Cruikshanks (approved):

PR #2156, opened 2025-07-01
Update notice setup returns paper forms change link

jonathangoulding:

This change adds the returns data to the view and the associated change link.

Returns only shows for the 'paper-forms' journey and at least one will have been selected to show.

When multiple are selected we need to render the returns in a list in the UI.


app/presenters/notices/setup/check-notice-type.presenter.js
```diff
@@ -18,56 +20,97 @@ const NOTICE_TYPE_TEXT = {
...
+function _returns(sessionId, selectedReturns, dueReturns) {
+  const selectedDueReturns = dueReturns.filter((dueReturn) => {
+    return selectedReturns.includes(dueReturn.returnId)
+  })
+
+  const selectedDueReturnsNewLineString = selectedDueReturns
+    .map((returnItem) => {
+      return `${returnItem.returnReference} - ${formatLongDate(new Date(returnItem.startDate))} to ${formatLongDate(new Date(returnItem.endDate))}`
+    })
+    .join('<br>')
```
Cruikshanks, line 41: I think I got told off for using a `<br>` for separating stuff!

But also this brings HTML layout into the code, when we should always be trying to go the other way.

So, we should just be returning the array of strings and letting the view handle the formatting. I'm pretty sure someone even [created a macro](https://github.com/DEFRA/water-abstraction-system/pull/1630) specifically for this. 🤔 😁

Cruikshanks (changes requested):

PR #2168, opened 2025-07-04
Add view select page

robertparkinson:

This PR adds in the functionality to query OS places to return a list of addresses and display them to the user.

### DIFF
--- a/app/presenters/address/select.presenter.js
+++ b/app/presenters/address/select.presenter.js
@@ -28,7 +28,7 @@ function _addresses(addresses) {
     }
   })
 
-  displayAddresses.push({
+  displayAddresses.unshift({
     value: 'select',
     selected: true,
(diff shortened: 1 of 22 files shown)


app/presenters/address/select.presenter.js
```diff
@@ -8,10 +8,33 @@
...
+
+function _addresses(addresses) {
+  const displayAddresses = addresses.map((address) => {
+    return {
+      value: address.uprn,
+      text: address.address
+    }
+  })
+
+  displayAddresses.push({
```
Cruikshanks, line 31: Have this at the start of the options in the drop-down. I believe `unshift()` will at it to the start.

```suggestion
  displayAddresses.unshift({
```

Cruikshanks (changes requested):

PR #2169, opened 2025-07-04
Address lookup requests

robertparkinson:

This PR adds in the functionality to query OS places by either postcode or an individual UPRN.

### DIFF
--- a/app/requests/address-lookup/lookup-postcode.request.js
+++ b/app/requests/address-lookup/lookup-postcode.request.js
@@ -1,0 +1,53 @@
+'use strict'
+
+/**
+ * Use for making http requests to the address facade
+ * @module LookupPostcodeRequest
+ */
+
+const BaseRequest = require('../base.request.js')
+
+const requestConfig = require('../../../config/request.config.js')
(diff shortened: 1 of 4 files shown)


app/requests/address-lookup/lookup-postcode.request.js
```diff
@@ -0,0 +1,53 @@
+'use strict'
+
+/**
+ * Use for making http requests to the address facade
```
Cruikshanks, line 4:
```suggestion
 * Sends a request to the address facade for the provided postcode
```

app/requests/address-lookup/lookup-uprn.request.js
```diff
@@ -0,0 +1,53 @@
+'use strict'
+
+/**
+ * Use for making http requests to the address facade
```
Cruikshanks, line 4:
```suggestion
 * Sends a request to the address facade for the provided uprn
```

Cruikshanks (changes requested):
Just two NITS.

It should be that the module summary lines up with the function summary, because the module will only expose the one function.

PR #2174, opened 2025-07-08
Add view manual address page

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-5046

This PR adds in the functionality to view the manual address page and if an address had been previously saved, show the saved values.


app/presenters/address/manual.presenter.js
```diff
@@ -8,10 +8,19 @@
...
+ * @param {object} address - An object containing the saved address
+ *
  * @returns {object} - The data formatted for the view template
  */
-function go() {
-  return {}
+function go(address = {}) {
+  return {
+    pageTitle: 'Enter the address',
+    ...(address?.addressLine1 && { addressLine1: address.addressLine1 }),
```
jonathangoulding, line 18: could these be 'addressLine1' with defaults of null ? 

Would be clearer in the return object what each is called ? 

Feel free to ignore but i don't think this is a common presenter return object pattern.

app/services/address/manual.service.js
```diff
@@ -19,13 +19,23 @@ const SessionModel = require('../../models/session.model.js')
 async function go(sessionId) {
   const session = await SessionModel.query().findById(sessionId)
 
-  const pageData = ManualAddressPresenter.go(session)
+  const pageData = ManualAddressPresenter.go(session.address)
 
   return {
+    activeNavBar: 'search',
+    backLink: _backLink(session.address, sessionId),
```
jonathangoulding, line 26: is the back link suppose to be in the presenter ? 

it does make sense here as well.

app/presenters/address/manual.presenter.js
jonathangoulding: are these better all on an address object ? 

address.addressLine1
address.postcode

PR #2179, opened 2025-07-09
Update notice setup fetch download recipients

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5143

We previously implemented an update ad-hoc journey allowing the user to select a return to send the notice for.

This has highlighted an issue with our logic where we did not limit the returns logs due to 'today'. This means we were getting all the return logs due even future return logs.

This change update the 'FetchDownloadRecipientsService' to only return recipients where return logs are due and the endDate is in the pass.

To get to the page where this logic is due there is already a check on the return logs 'endDate'. This change aligns the two.

### DIFF
--- a/test/services/notices/setup/fetch-download-recipients.service.test.js
+++ b/test/services/notices/setup/fetch-download-recipients.service.test.js
@@ -26,7 +26,7 @@ describe('Notices - Setup - Fetch Download Recipients service', () => {
   before(async () => {
     dueDate = `${year}-04-28` // This needs to differ from any other returns log tests
 
-    testRecipients = await LicenceDocumentHeaderSeeder.seed(true, dueDate)
+    testRecipients = await LicenceDocumentHeaderSeeder.seed(true, dueDate, endDate)
   })
 
   describe('when the "journey" is for notifications', () => {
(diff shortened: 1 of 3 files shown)


test/services/notices/setup/fetch-download-recipients.service.test.js
```diff
@@ -358,6 +358,20 @@ describe('Notices - Setup - Fetch Download Recipients service', () => {
...
       })
+
+      describe('when the end date is greater than today', () => {
+        beforeEach(async () => {
+          testRecipients = await LicenceDocumentHeaderSeeder.seed(true, dueDate, '3000-01-01')
+
+          session = { licenceRef: testRecipients.primaryUser.licenceRef }
+        })
+
+        it('correctly an empty array - no return logs found', async () => {
```
Cruikshanks, line 369:
```suggestion
        it('correctly returns an empty array (no return logs found)', async () => {
```

test/support/seeders/licence-document-header.seeder.js
```diff
@@ -15,19 +15,24 @@ const ReturnLogHelper = require('../helpers/return-log.helper.js')
  * @param {boolean} enableReturnLog - defaulted to true, this needs to be false if you do not want the `licenceDocumentHeader`
  * to be included in the recipients list
  * @param {string} returnLogDueDate - defaulted to the same due date set by the returnsLogHelper
+ * @param {string} returnLogEndDate - defaulted to the same due date set by the returnsLogHelper
```
Cruikshanks, line 18:
```suggestion
 * @param {string} returnLogEndDate - defaulted to the same end date set by the returnsLogHelper
```

Cruikshanks (changes requested):
NITS

PR #2186, opened 2025-07-11
Refactor return version set up journey views

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-5139

Now that we have completed the return version set up journey it was time to revisit the views and standardise the code across them. As multiple people were working on the different pages it is always the case that there will be minor variations in how things are implemented. 

This PR tries to standardise the approach across the views.


app/views/return-versions/setup/note.njk
```diff
@@ -3,9 +3,9 @@
 {% from "govuk/components/button/macro.njk" import govukButton %}
 {% from "govuk/components/character-count/macro.njk" import govukCharacterCount %}
 {% from 'govuk/components/error-summary/macro.njk' import govukErrorSummary %}
+{% from "macros/licence-reference-page-heading.njk" import pageHeading %}
```
Cruikshanks, line 6:
```suggestion

{% from "macros/licence-reference-page-heading.njk" import pageHeading %}
```

Cruikshanks (changes requested):
This is awesome! Learnt so many new things I didn't know you could do.

The changes are just NITS.

Cruikshanks (approved):

PR #2192, opened 2025-07-14
Add basic notice setup return form PDF

jonathangoulding:

This changes uses the previously implemented 'GenerateReturnFormRequest' to convert HTML into a PDF. 

This change adds a new route to allow users to view the generated PDF, this is not only useful for testing, but we use this on the notice setup check preview link to allow users to preview the letter before sending. (This uses the browser's built-in PDF viewer.)

We have created the PrepareReturnFormsService to handle orchestrating fetching the data and returning the file (ArrayBuffer) to a calling service. We have implemented the preview page as part of this change. 

We will need to add another service to call the PrepareReturnFormsService when se send the PDF to notify.


app/requests/gotenberg/send-return-form.request.js
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Creates and sends a html string to Gotenberg to convert it into a pdf
+ *
+ * > Important - 'preferCssPageSize' tells Goetnberg to rely on our styling and to not add any default margin / padding.
+ * > This is required in conjunction with setting 'marginTop' etc.
+ * @param pageData
+ */
+async function go(pageData) {
+  try {
```
jonathangoulding, line 26: add tests for this

test/controllers/notices-setup.controller.test.js
```diff
@@ -809,6 +810,39 @@ describe('Notices Setup controller', () => {
     })
   })
 
+  describe('notices/setup/{sessionId}/preview-return-forms', () => {
+    describe('GET', () => {
+      let buffer
+      beforeEach(async () => {
```
Jozzey, line 816: I don't think this needs to be asyncronous. Same for a lot of the other `beforeEach` in this test file.
```suggestion
      beforeEach(() => {
```

Jozzey, line 816: We'd normally have a space between the `let` and the `beforeEach`

Jozzey (dismissed):
Just a couple of little things

app/presenters/notices/setup/preview-return-forms.presenter.js
```diff
@@ -0,0 +1,35 @@
...
+  }
+}
+
+/*
+ * The data for the cover page
+ */
+function _cover() {
+  return {
+    title: `Water abstraction day return`
+  }
```
robertparkinson, line 30: can we add something to the preview version to water mark it so it says "EXAMPLE" or something else? my worry is staff just downloading the preview and sending it to users outside the system

robertparkinson (changes requested):

robertparkinson (dismissed):

Jozzey (dismissed):

PR #2206, opened 2025-07-21
Submit international address

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-5047

As part of the journey to create ad-hoc letters to be sent out to users we need to be able to provide an address.

This PR adds in the functionality to submit the international address page. This is for when an address is outside the uk and a country must be provided.


StuAA78 (approved):
👍

PR #2207, opened 2025-07-21
Update bill run setup views to 'template' standard

StuAA78:
https://eaflood.atlassian.net/browse/WATER-5152

We use the changes made in https://github.com/DEFRA/water-abstraction-system/pull/2186 as a model for updating the "Create a bill run" journey.

Going by this PR, we're looking at implementing the following principles:

### Errors

Where we see things like this:

```njk
{% if error.text.startResult %}
    {% set startResultErrorClass = " govuk-input--error" %}
    {% set startResultErrorMessage = { text: error.text.startResult } %}
{% else %}
    {% set startResultErrorClass = "" %}
    {% set startResultErrorMessage = null %}
{% endif %}

{# ... #}

{{ govukDateInput({
    id: "abstraction-period-start",
    namePrefix: "abstraction-period-start",
    errorMessage: { text: error.text.startResult } if error.text.startResult,
    items: [
        {
            classes: "govuk-input--width-2" + startResultErrorClass,
            {# ... #}
        },
        {# ... #}
    ]
}) }}
```

We can simply do this:

```njk
{{ govukDateInput({
    id: "abstraction-period-start",
    namePrefix: "abstraction-period-start",
    errorMessage: startResultErrorMessage,
    items: [
        {
            classes: "govuk-input--width-2 {{ 'govuk-input--error' if error.text.startResult }}",
            {# ... #}
        },
        {# ... #}
    ]
}) }}
```

In other words, we ditch the first block entirely that sets the error class and message, and instead set them conditionally within whatever uses them.

### Section comments

We seem to be ditching entirely section comments, eg:

```njk
{# Main heading #}

{# Back link #}
```

etc.

### `div`s

As a general point, we usually don't need to wrap things in `<div>` tags. Wrap text in `<p>` tags; individual components don't need anything wrapping them.

### `govuk-body` class

We remove the `govuk-body` class wherever we see it. So for example. if we have `<p class="govuk-body">...</p>` anywhere then we retain the `<p>` tags but remove the class.

If we have `<div class="govuk-body">` then we would remove the `<div>` entirely in line with the previous principle.

### Page headings

If every page shares a common heading, we can extract it to a suitable file in `views/macros`. As we update more and more pages, we will likely accumulate common page headings which we can re-use across journeys (for example, there is already `views/macros/licence-reference-page-heading.njk` which shows the page title with the licence ref smaller above it; this is common in our service). When we do import a page heading, ensure we have a separating line between it and the govuk component imports, ie:

```njk
{% extends 'layout.njk' %}
{% from "govuk/components/back-link/macro.njk" import govukBackLink %}

{% from "macros/licence-reference-page-heading.njk" import pageHeading %}
```

We insert the page heading like so:

```njk
{{ pageHeading(licenceRef, pageTitle) }}

{# or if we want to set specific styling to match existing design: #}
{{ pageHeading(licenceRef, pageTitle, 'govuk-heading-xl govuk-!-margin-bottom-3') }}

```

If the page is largely comprised of a single component (eg. a page displaying radio buttons) we set the page heading:

```njk
{% set pageHeading %}
    {{ pageHeading(licenceRef, pageTitle) }}
{% endset %}
```

We can then use it in a `fieldset`:

#### `fieldset`

When the page is largely comprised of a single component, ensure we have a `fieldset` in it, eg:

```njk
{{ govukCheckboxes({
    name: "agreementsExceptions",
    fieldset: {
        legend: {
            html: pageHeading
        }
    },
    {# ... #}
}) }}
```

We also clean up any existing `fieldset.legend`.

### `govukButton`

We consistently write it on one line, eg:

```njk
{# Incorrect: #}
{{ govukButton({
    text: "Remove",
    preventDoubleClick: true
}) }}

{# Correct: #}
{{ govukButton({ text: "Continue", preventDoubleClick: true }) }}
```

### `govuk-summary-list`

If we have a summary list, we don't need the `govuk-summary-list` class, eg:

```njk
{# Incorrect: #}
{{ govukSummaryList({
    classes: 'govuk-!-margin-bottom-2',
    rows: [
        {
            classes: 'govuk-summary-list govuk-summary-list__row--no-border',
            key: {
                text: "Reason"
            },
            value: {
                text: reason
            }
        },
        {# ... #}
    ]
}) }}

{# Correct: #}
{{ govukSummaryList({
    classes: 'govuk-!-margin-bottom-2',
    rows: [
        {
            classes: 'govuk-summary-list__row--no-border', {# Removed class #}
            key: {
                text: "Reason"
            },
            value: {
                text: reason
            }
        },
        {# ... #}
    ]
}) }}
```

### `text` with `attributes` vs `html`

If we're using `html` somewhere but only to apply an attribute, use `text` instead. For example:

```njk
{# Incorrect: #}
{{ govukSummaryList({
    {# ... #}
    items: [
        {
            html: '<span data-test="change-reason">' + 'Change' + '</span>',
            href: "/system/return-versions/setup/" + sessionId + "/start-date",
            visuallyHiddenText: "the reason for the return requirement"
        }
    ]
}) }}

{# Correct: #}
{{ govukSummaryList({
    {# ... #}
    items: [
        {
              attributes: { 'data-test': 'change-reason' },
              text: 'Change',
              href: reasonLink, {# See below for why this was also changed #}
              visuallyHiddenText: "the reason for the return requirement"
        }
    ]
}) }}
```

### Pass links to template from presenter instead of hardcoding

In the example above, we see that `href` goes from a hardcoded url `"/system/return-versions/setup/" + sessionId + "/start-date"` to `reasonLink`, which we define in the presenter and pass to the template. This also applies to:

### Backlinks

We define `backlink` in the presenter and pass to the template:

```njk
{# Incorrect: #}
{{
    govukBackLink({
        text: 'Back',
        href: '/system/bill-runs/setup/' + sessionId + '/region'
    })
}}

{# Correct: #}
{{
    govukBackLink({
        text: 'Back',
        href: backlink
    })
}}
```

### Hint text

If we have a standalone hint on a page with a single main component, see if we can add it to the component. For example:

```njk
{# Incorrect: #}
<div class="govuk-hint"> Select all that apply </div>

{{ govukCheckboxes({
    name: "purposes",
    items: checkBoxItems,
    {# ... #}
}) }}

{# Correct: #}
{{ govukCheckboxes({
    name: "purposes",
    hint: {
        text: "Select all that apply"
    },
    items: checkBoxItems,
    {# ... #}
}) }}
```

### `govuk-label` classes

If we have some `label` text then we don't also need to add a `govuk-label` class. For example:

```njk
{# Incorrect: #}

{{ govukInput({
    label: {
        text: "Enter the volume",
        classes: "govuk-label"
    },
    {# ... #}
}) }}

{# Correct: #}
{{ govukInput({
    label: {
        text: "Enter the volume"
    },
    {# ... #}
}) }}
```

(Note some other changes there but the key thing is the removal of `classes: "govuk-label--l"`)

### DIFF
--- a/app/controllers/notices-setup.controller.js
+++ b/app/controllers/notices-setup.controller.js
@@ -435,10 +435,11 @@ async function submitReturnForms(request, h) {
 async function submitSelectRecipients(request, h) {
   const {
     payload,
-    params: { sessionId }
+    params: { sessionId },
+    yar
   } = request
 
-  const pageData = await SubmitSelectRecipientsService.go(sessionId, payload)
+  const pageData = await SubmitSelectRecipientsService.go(sessionId, payload, yar)
 
   if (pageData.error) {
     return h.view(`notices/setup/select-recipients.njk`, pageData)
--- a/app/services/notices/setup/submit-select-recipients.service.js
+++ b/app/services/notices/setup/submit-select-recipients.service.js
@@ -6,6 +6,7 @@
  * @module SubmitSelectRecipientsService
  */
 
+const GeneralLib = require('../../../lib/general.lib.js')
 const RecipientsService = require('./recipients.service.js')
 const SelectRecipientsPresenter = require('../../../presenters/notices/setup/select-recipients.presenter.js')
 const SelectRecipientsValidator = require('../../../validators/notices/setup/select-recipients.validator.js')
@@ -16,10 +17,11 @@ const SessionModel = require('../../../models/session.model.js')
  *
  * @param {string} sessionId
  * @param {object} payload - The submitted form data
+ * @param {object} yar - The Hapi `request.yar` session manager passed on by the controller
  *
  * @returns {Promise<object>} - The data formatted for the view template
  */
-async function go(sessionId, payload) {
+async function go(sessionId, payload, yar) {
   const session = await SessionModel.query().findById(sessionId)
 
   _handleOneOptionSelected(payload)
@@ -28,6 +30,12 @@ async function go(sessionId, payload) {
 
   if (!validationResult) {
     await _save(session, payload)
+
+    GeneralLib.flashNotification(
+      yar,
+      'Updated',
+      'The recipients have been changed. Check details before sending invitations.'
+    )
 
     return {}
   }
--- a/test/services/notices/setup/submit-select-recipients.service.test.js
+++ b/test/services/notices/setup/submit-select-recipients.service.test.js
@@ -23,6 +23,7 @@ describe('Notices - Setup - Submit Select Recipients Service', () => {
   let recipients
   let session
   let sessionData
+  let yarStub
 
   beforeEach(async () => {
     payload = { recipients: ['123'] }
@@ -33,6 +34,8 @@ describe('Notices - Setup - Submit Select Recipients Service', () => {
     recipients = RecipientsFixture.recipients()
 
     Sinon.stub(RecipientsService, 'go').resolves([recipients.primaryUser])
+
+    yarStub = { flash: Sinon.stub().returns([{ title: 'Test', text: 'Notification' }]) }
   })
 
   afterEach(() => {
@@ -41,15 +44,28 @@ describe('Notices - Setup - Submit Select Recipients Service', () => {
 
   describe('when called', () => {
     it('saves the submitted value', async () => {
-      await SubmitSelectRecipientsService.go(session.id, payload)
+      await SubmitSelectRecipientsService.go(session.id, payload, yarStub)
 
       const refreshedSession = await session.$query()
 
       expect(refreshedSession.selectedRecipients).to.equal(['123'])
     })
 
+    it('sets a flash message', async () => {
+      await SubmitSelectRecipientsService.go(session.id, payload, yarStub)
+
+      // Check we add the flash message
+      const [flashType, bannerMessage] = yarStub.flash.args[0]
+
+      expect(flashType).to.equal('notification')
+      expect(bannerMessage).to.equal({
+        text: 'The recipients have been changed. Check details before sending invitations.',
+        title: 'Updated'
+      })
+    })
+
     it('continues the journey', async () => {
-      const result = await SubmitSelectRecipientsService.go(session.id, payload)
+      const result = await SubmitSelectRecipientsService.go(session.id, payload, yarStub)
 
       expect(result).to.equal({})
     })
@@ -62,7 +78,7 @@ describe('Notices - Setup - Submit Select Recipients Service', () => {
       })
 
       it('returns page data for the view, with errors', async () => {
-        const result = await SubmitSelectRecipientsService.go(session.id, payload)
+        const result = await SubmitSelectRecipientsService.go(session.id, payload, yarStub)
 
         expect(result).to.equal({
           backLink: `/system/notices/setup/${session.id}/check`,


Cruikshanks (approved):
Oh my gosh, I ❤️ the PR description. That is _so_ being lifted and turned into a guide in our [team repo](https://github.com/DEFRA/water-abstraction-team)!

Menatime, I think this encapsulates what @robertparkinson did in the original PR, so I'm approving. But I'd also _really_ appreciate your thoughts @robertparkinson before we merge.

app/views/bill-runs/setup/no-licences.njk
```diff
@@ -3,27 +3,22 @@
...
-
+  <h1 class="govuk-heading-l govuk-!-margin-bottom-3">{{ pageTitle }}</h1>
   {{ govukWarningText({
     text: 'Check there are licences ready to be billed and try again.',
     iconFallbackText: 'Warning'
   }) }}
 
-  <p class="govuk-body">
+  <p>
     <a href="/system/bill-runs" class="govuk-link">Return to bill runs</a>
```
robertparkinson, line 22: This doesn't need to be in a `<p>` as it's not text. if we need to add styling or padding we can add it as a class in the `<a>`

app/views/bill-runs/setup/region.njk
```diff
@@ -27,35 +26,33 @@
...
-      {% for region in regions %}
-        {% set regionItem = { text: region.displayName, value: region.id, checked: region.id === selectedRegion } %}
+    {% set regionItems = [] %}
+    {% for region in regions %}
+      {% set regionItem = { text: region.displayName, value: region.id, checked: region.id === selectedRegion } %}
 
-        {# Push our item into the region items array #}
-        {% set regionItems = (regionItems.push(regionItem), regionItems) %}
-      {% endfor %}
+      {# Push our item into the region items array #}
```
robertparkinson, line 36: do we need this comment? I believe the code is descriptive enough and this doesn't add anything

app/views/bill-runs/setup/region.njk
```diff
@@ -27,35 +26,33 @@
...
+      attributes: {
+        'data-test': 'bill-run-region'
+      },
+      name: 'region',
+      errorMessage: error,
+      fieldset: {
+        legend: {
+          text: pageTitle,
+          isPageHeading: true,
+          classes: 'govuk-fieldset__legend--l govuk-!-margin-bottom-6'
```
robertparkinson, line 50: as you're setting the isPageHeading to true I don't think you need to set the legend to large `govuk-fieldset__legend--l` as that should be handled by the heading class that gets added already. would need to double check the differences on the page though

app/views/bill-runs/setup/season.njk
```diff
@@ -27,38 +26,36 @@
...
+      attributes: {
+        'data-test': 'bill-run-season'
+      },
+      name: 'season',
+      errorMessage: error,
+      fieldset: {
+        legend: {
+          text: pageTitle,
+          isPageHeading: true,
+          classes: 'govuk-fieldset__legend--l govuk-!-margin-bottom-6'
```
robertparkinson, line 42: same as above

app/views/bill-runs/setup/year.njk
```diff
@@ -27,27 +26,25 @@
...
+      attributes: {
+        'data-test': 'bill-run-year'
+      },
+      name: 'year',
+      errorMessage: error,
+      fieldset: {
+        legend: {
+          text: pageTitle,
+          isPageHeading: true,
+          classes: 'govuk-fieldset__legend--l govuk-!-margin-bottom-6'
```
robertparkinson, line 42: same as above

app/views/bill-runs/setup/region.njk
robertparkinson: Glad you tried it. For future reference the design system documentation is not the best. I don't think the people who write that are the devs.. I have found problems with it before. Best to check the component directly in the govuk frontend. but yeah it looks like it sets `govuk-fieldset__heading` in the isPageHeading code but that sounds like it's not a large font size based on your trials. 
https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/fieldset/template.njk

PR #2215, opened 2025-07-28
Show all purposes in return log pages

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-5150

A user pointed out that we're not displaying the full list of purposes in two places but did point out where it was displayed correctly. Having dug into the code and looking at where else we show the purposes I noticed we're doing it in two places already so I've pulled the code for it out of both of those places, combined it so it includes the break clause for being handed an empty array and then started calling it from the various places it needs to be seen so we are consistent across the site.

### DIFF
--- a/app/presenters/base.presenter.js
+++ b/app/presenters/base.presenter.js
@@ -219,6 +219,43 @@ function formatPounds(valueInPence) {
 }
 
 /**
+ * Format the purposes for display on screen
+ *
+ * This function assumes that when provided the purposes are in the format:
+ *
+ * ```javascript
+ * [{
+ *  alias: 'manual description',
+ *  primary: { code: 'A', description: 'Agriculture' },
+ *  tertiary: { code: '400', description: 'Spray Irrigation - Direct' },
+ *  secondary: { code: 'AGR', description: 'General Agriculture' }
+ * }]
+ * ```
+ *
+ * It is possible that a return log has zero or more purposes associated it. If this is the case we need the description
+ * of each purpose from the array of purposes to be concatenated into a single comma separated string ready to present.
+ *
+ * @param {Array} purposes - the purposes taken from the return logs metadata
+ *
+ * @returns {string} the purpose descriptions as a string, separated by commas if more than one description exists
+ */
+function formatPurposes(purposes) {
+  if (!purposes || purposes.length === 0) {
+    return ''
+  }
+
+  const purposeDescriptionArray = purposes.map((purpose) => {
+    if (purpose.alias) {
+      return `${purpose.tertiary.description} (${purpose.alias})`
+    }
+
+    return purpose.tertiary.description
+  })
+
+  return purposeDescriptionArray
+}
+
+/**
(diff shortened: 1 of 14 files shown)


app/presenters/base.presenter.js
```diff
@@ -218,6 +218,43 @@ function formatPounds(valueInPence) {
...
+ * @returns {string} the purpose descriptions as a string, separated by commas if more than one description exists
+ */
+function formatPurposes(purposes) {
+  if (!purposes || purposes.length === 0) {
+    return ''
+  }
+
+  const purposeDescriptionArray = purposes.map((purpose) => {
+    if (purpose.alias) {
+      return `${purpose.tertiary.description} (${purpose.alias})`
```
Jozzey, line 249: Just wondering where the requirement came from to concatenate the "description" and "alias" if an "alias" exists? Previously we appeared to just show one or the other. Concatenating is going to make the string considerably longer, don't know if this impacts the formatting on any of the pages this is displayed 🤷🏼‍♂️

app/presenters/return-logs/base-return-logs.presenter.js
```diff
@@ -43,32 +43,6 @@ function formatMeterDetails(meter) {
...
- * if purposes is empty
- */
-function formatPurpose(purposes) {
-  if (!purposes || purposes.length === 0) {
-    return ''
-  }
-
-  const [firstPurpose] = purposes
-
-  return firstPurpose.alias ? firstPurpose.alias : firstPurpose.tertiary.description
```
Jozzey, line 69: From looking at this removed function it looks like we very much intentionally only displayed the first purpose. I'm not sure why that was, but am not really comfortable approving without knowing why we did it that way in the first place.

app/presenters/base.presenter.js
robertparkinson: This came up in conversation with chris when we were triaging the bug last week. If the alias is there it should be displayed as it is more important as it is something provided by the abstractor. I've checked it on all of the pages I'm familiar with and added screen shots to the story but yeah I meant to ask you how to get to that bill page.

Jozzey: Fair enough. I'd still probably just display either the alias or the description rather than concatenate them together as we did previously `return purpose.alias ? purpose.alias : purpose.tertiary.description`. Unless we have been asked specifically to concatenate the two together. Just think it may make the string a bit messy.

app/presenters/return-logs/base-return-logs.presenter.js
Jozzey: This is the original ticket for info: https://eaflood.atlassian.net/browse/WATER-4444 Looks like we were just migrating an existing page. No mention of the purposes in there, guess it was just built as per the legacy page.

Jozzey: Yes maybe just double check to make sure they are clear on the impact of the change.

Also I wouldn't necessarily label this as a "bug" as the system is working as intended. I'd say this is an "enhancement" as they are requesting a change to existing funtionality.

PR #2224, opened 2025-07-31
Update notices setup select recipients to handle selected recipients

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5041

This change adds the selected recipients to the session and marks them as checked in the presenter. 

All recipients are initially set to 'checked'. When the user has unselected any recipients they will not be 'checked' until updated by the user again. 

This change does not update the common service to remove the unselected recipients from the relevant recipients. This will be done in another change.

### DIFF
--- a/app/presenters/notices/setup/select-recipients.presenter.js
+++ b/app/presenters/notices/setup/select-recipients.presenter.js
@@ -26,10 +26,10 @@ function go(session, recipients) {
 }
 
 /**
- * Determines if a contact should be marked as checked.
+ * Determines if a recipient should be marked as checked.
  *
  * If no `selectedRecipients` are provided (i.e., it's falsy), then all contacts should be marked as checked (`true`).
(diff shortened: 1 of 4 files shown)


app/presenters/notices/setup/select-recipients.presenter.js
```diff
@@ -16,23 +16,39 @@ const ContactPresenter = require('./contact.presenter.js')
...
-    recipients: _recipients(recipients)
+    recipients: _recipients(recipients, selectedRecipients)
   }
 }
 
-function _recipients(recipients) {
+/**
+ * Determines if a recipient should be marked as checked.
+ *
+ * If no `selectedRecipients` are provided (i.e., it's falsy), then all contacts should be marked as checked (`true`).
```
robertparkinson, line 31: extra , in the () I think?

robertparkinson (approved):

PR #2226, opened 2025-07-31
Fix communications notifications

jonathangoulding:
Fix communications notifications

https://eaflood.atlassian.net/browse/WATER-5188

There is a bug introduced by the new notices work we have done. We recently migrated the legacy notifications logic into 'system'. 

When the legacy notifications where sent they used a delayed sending mechanism which used a 'sendAfter' date to track when a notification was sent. 

When we migrated this logic we did not use this delayed mechanism (or an overly complicated message queue system). We instead send the notification when it has been submitted and save it in the database with the created date set (legacy notification do not have a created at date).

So we have two dates used to show notification sent. 

This change updates the logic to use both of these dates. We determined that a 'sendAfter' should be used in place of the 'createdAt' date when showing the date. 

And as the new way of sending notifications use the 'createdAt' we sort by 'createdAt' then 'sendAfter' to show the user the latest notifications (we have handled null dates in the updated query).

Another change required was updating the communications tab to use these two dates instead of the 'event.createdAt' date. This was a carry over from the legacy code.

### DIFF
--- a/app/presenters/licences/view-licence-communications.presenter.js
+++ b/app/presenters/licences/view-licence-communications.presenter.js
@@ -31,7 +31,7 @@ function _communications(communications, documentId, licenceId) {
       link: _link(communication.id, documentId, licenceId),
       type: _type(communication),
       sender: communication.event.issuer,
-      sent: formatLongDate(new Date(communication.event.createdAt)),
+      sent: _sent(communication),
       method: sentenceCase(communication.messageType)
     }
   })
@@ -45,10 +45,16 @@ function _link(communicationId, documentId, licenceId) {
   return `/licences/${documentId}/communications/${communicationId}`
 }
 
+function _sent(communication) {
+  return communication.sendAfter
+    ? formatLongDate(new Date(communication.sendAfter))
+    : formatLongDate(new Date(communication.createdAt))
+}
+
 function _type(communication) {
   return {
     label: _typeLabel(communication),
-    sentVia: `sent ${formatLongDate(new Date(communication.event.createdAt))} via ${communication.messageType}`,
+    sentVia: `sent ${_sent(communication)} via ${communication.messageType}`,
     pdf: communication.messageRef.includes('pdf')
   }
 }
--- a/app/presenters/notifications/view-notification.presenter.js
+++ b/app/presenters/notifications/view-notification.presenter.js
@@ -15,7 +15,7 @@ const { formatLongDate, sentenceCase } = require('../base.presenter.js')
  * @returns {object} The data formatted for the view template
  */
 function go(notificationData) {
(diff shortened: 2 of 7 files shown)


app/presenters/licences/view-licence-communications.presenter.js
```diff
@@ -31,7 +31,7 @@ function _communications(communications, documentId, licenceId) {
       link: _link(communication.id, documentId, licenceId),
       type: _type(communication),
       sender: communication.event.issuer,
-      sent: formatLongDate(new Date(communication.event.createdAt)),
+      sent: formatLongDate(new Date(communication.createdAt)),
```
jonathangoulding, line 34: this can be the event created at but it might not align with the notification created at / send after date.

app/presenters/notifications/view-notification.presenter.js
```diff
@@ -25,7 +25,7 @@ function go(notificationData) {
     licenceRef,
     messageType,
     pageTitle: _pageTitle(notificationData.notification),
-    sentDate: formatLongDate(sendAfter)
+    sentDate: sendAfter ? formatLongDate(sendAfter) : formatLongDate(createdAt)
```
jonathangoulding, line 28: this will fix the render issue

app/services/licences/fetch-communications.service.js
```diff
@@ -27,16 +27,16 @@ async function go(licenceRef, page) {
...
     .where('licences', '@>', `["${licenceRef}"]`)
     .andWhere('notifyStatus', 'in', ['delivered', 'received'])
     .andWhere('eventId', 'is not', null)
     .withGraphFetched('event')
     .modifyGraph('event', (builder) => {
-      builder.select(['createdAt', 'issuer', 'metadata', 'status', 'subtype', 'type'])
+      builder.select(['issuer', 'metadata', 'status', 'subtype', 'type'])
     })
     .page(page - 1, DatabaseConfig.defaultPageSize)
-    .orderBy('sendAfter', 'desc')
```
jonathangoulding, line 39: this won't sort null values ?

app/presenters/notifications/view-notification.presenter.js
```diff
@@ -15,7 +15,7 @@ const { formatLongDate, sentenceCase } = require('../base.presenter.js')
  * @returns {object} The data formatted for the view template
  */
 function go(notificationData) {
-  const { messageType, plaintext, personalisation, sendAfter } = notificationData.notification
+  const { messageType, plaintext, personalisation, sendAfter, createdAt } = notificationData.notification
```
Jozzey, line 18: Destructured functions should be listed alphabetically
```suggestion
  const { createdAt, messageType, personalisation, plaintext, sendAfter } = notificationData.notification
```

Jozzey (dismissed):

PR #2230, opened 2025-08-01
Add link to select recipients page

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-5042

As part of sending out ad-hoc notifications to users we need to be able to add a contact that is not already associated to the licence. 

This PR adds a link from the select recipients page that takes you to the select a contact type page.

### DIFF
--- a/app/views/notices/setup/select-recipients.njk
+++ b/app/views/notices/setup/select-recipients.njk
@@ -4,6 +4,7 @@
 {% from "govuk/components/button/macro.njk" import govukButton %}
 {% from "govuk/components/checkboxes/macro.njk" import govukCheckboxes %}
 {% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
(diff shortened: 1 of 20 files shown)


app/views/notices/setup/select-recipients.njk
```diff
@@ -41,6 +42,16 @@
...
   {%endif%}
 
+  {% set innerHintHtml %}
+    {{ govukHint({
+      id: "recipients",
+      classes: 'govuk-checkboxes__hint govuk-!-padding-left-0 govuk-!-margin-bottom-5',
+      text: "Select all that apply"
+      })
+    }}
+    <a class="govuk-link" href="/system/notices/setup/{{sessionId}}/contact-type">Set up a single use address or email address</a>
```
jonathangoulding, line 52: link should be in the presenter. 

Makes it testable.

jonathangoulding (approved):

PR #2231, opened 2025-08-01
Ad-hoc notice - enter postcode

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-5044

As part of sending out ad-hoc notifications to users we need to be able to add a contact that is not already associated to the licence.

This PR allows the user to enter a postcode which will be used in the address lookup service.

### DIFF
--- a/app/controllers/notices-setup.controller.js
+++ b/app/controllers/notices-setup.controller.js
@@ -332,12 +332,12 @@ async function submitContactType(request, h) {
   } = request
 
   const pageData = await SubmitContactTypeService.go(sessionId, payload)
-
+console.log(pageData)
   if (pageData.error) {
     return h.view(`notices/setup/contact-type.njk`, pageData)
   }
 
-  if (pageData.contactType === 'post') {
+  if (pageData.type === 'post') {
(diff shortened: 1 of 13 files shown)


Cruikshanks (changes requested):
The one issue I had was I got in a tizz with `contactType` and `type`. At one point I thought I'd spotted an issue but then realised it was okay.

Once I clocked this in `submit-contact-type.service.js`, I got my head round it.

```javascript
  const _contactType = {
    name: payload.name,
    type: payload.type
  }

  session.contactType = _contactType
```

We're creating a property in the session called `contactType`, that then has its own `.type` property.

A name like `contactType` to me infers it'll hold a discrete value, for example, 'email', not that it will hold multiple properties related to the 'contact'.

But when I looked again (assuming I understand what is happening!), we only store `session.contactType` if we are going down the 'letter' route.

So, I think I need to better understand what is required before I can give any opinion on what I think we should be doing.


This will definitely require a call (🤞 my internet is working better!)

Cruikshanks (approved):

PR #2233, opened 2025-08-05
Change top menu to profile details

ben-sagar:
https://eaflood.atlassian.net/browse/WATER-5171

Rename the current contact information top menu to profile details and the page title.

Move the whole thing from `water-abstraction-ui` into `water-abstraction-system`.

Put this behind a feature flag in `water-abstraction-ui`.


app/validators/users/profile-details.validator.js
```diff
@@ -0,0 +1,44 @@
...
+    tel: Joi.string().max(100).allow('').messages({
+      'string.max': 'Telephone number must be 100 characters or less'
+    }),
+    address: Joi.string().max(100).allow('').messages({
+      'string.max': 'Address must be 100 characters or less'
+    })
+  })
+
+  return schema.validate(payload, { abortEarly: false })
+}
```
ben-sagar, line 40: I've added some max length validation as there wasn't any validation on these at all - not sure what the correct business rules should be

app/services/users/submit-profile-details.service.js
```diff
@@ -0,0 +1,88 @@
...
+async function go(userId, payload, yar) {
+  const validationResult = _validate(payload)
+
+  if (!validationResult) {
+    await _save(userId, payload)
+
+    yar.flash('notification', {
+      title: 'Updated',
+      text: 'Profile details saved'
+    })
```
ben-sagar, line 41: In the existing system, when the profile details are successfully updated, the user is just returned to the home screen with no notification of whether the details have been updated.

I've added this notification and currently the controller keeps the user on this same page so they see the notification.

Not sure if that is a valid change to the journey.

app/views/users/profile-details.njk
```diff
@@ -0,0 +1,109 @@
...
+{% from "govuk/components/notification-banner/macro.njk" import govukNotificationBanner %}
+
+{% block breadcrumbs %}
+  {{
+    govukBackLink({
+      text: 'Back',
+      href: '/'
+    })
+  }}
+{% endblock %}
```
ben-sagar, line 16: I've kept the back link from the original system that just returns the user to the home page, even though that's not technically going "back" as they could have accessed this page from anywhere.

But it still felt the most appropriate way for the user to navigate elsewhere after updating their details.

app/services/users/submit-profile-details.service.js
Cruikshanks: As noted above, this is ideal for now (and better than the legacy solution).

app/validators/users/profile-details.validator.js
Cruikshanks: As discussed earlier, double-check the data to ensure we're not imposing anything that would cause existing data to be considered invalid.

Having the mechanism to allow us to validate is the key thing (because we should be). It's a minor thing to tweak it once available.

ben-sagar: I've looked through the existing records and they're all either test emails, old defra emails or actual `@environment-agency.gov.uk` so I've gone with just that as a valid domain.

I've also upped the address length to 300 as there were some that were over 100 characters.

But overall this data is very sparsely populated so this is largely guesswork to now.

test/services/users/view-profile-details.service.test.js
```diff
@@ -0,0 +1,113 @@
...
+      expect(result).to.include({ pageTitle: 'Profile details' })
+      expect(result).to.include('navigationLinks')
+      expect(result.navigationLinks).to.include([
+        {
+          active: true,
+          href: '/system/users/me/profile-details',
+          text: 'Profile details'
+        }
+      ])
+      expect(result).to.include(profileDetails)
```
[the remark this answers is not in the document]
ben-sagar: I've split out the navigation link test as that's a specific requirement unique to this page and different from the standard layout, so figured it was valuable to highlight it somehow - does this approach seem better?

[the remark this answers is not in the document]
Cruikshanks: Ahh, I see. You are have got the page that is the 'exception to the rule'.

By this I mean you don't have a 'presenter'. I believe every other page we've built has used an `app/presenter` in order to prepare the 'data' for the page.

So, the typical pattern in, for example, `view-foo.service.js` is

```javascript
async function go(fooId) {
  const foo = await FetchFooService.go(fooId)

  const formattedData = FooPresenter.go(foo)

  return {
    activeNavBar: 'search',
    ...formattedData
  }
}
```

Then we'd have a presenter test. If you take a look at some of those they also follow a general pattern.

- a general `it('correctly presents the data')` test at the top
- a `describe('the "foo" property')` scenario for each property where the output is not a straight replay (something has been calculated, generated, formatted, etc)

Your page on the overhand does not have to do _anything_ with the data you are getting out of the `user` instance. So, I agree there is no need for a presenter here. The service can return the data directly.

But I highlight 'presenters' because of the pattern we use in their tests

- the 'general' test covers _all the things_ both static and dynamic
- subsequent 'describe property' tests focus _only_ on the dynamic elements

Coming back to your service, though `navigationLinks` feels specific to this page and worthy of calling out, _every_ page has something specific to them. But we generally only call out the ones where the outcome can change. If the value is static, i.e. what goes in comes out, it's fine to cover them with a `expect(result).to.equal()` assertion.

Cruikshanks (changes requested):
I've provided a response to your comment. Hopefully makes sense.

We are now properly into 'just NITS' so you'll soon be shot of me!

test/services/users/view-profile-details.service.test.js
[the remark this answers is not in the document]
ben-sagar: I concede - done! 😆

test/services/users/view-profile-details.service.test.js
```diff
@@ -0,0 +1,98 @@
...
+      })
+    })
+  })
+
+  describe('when the service errors', () => {
+    beforeEach(() => {
+      userModelQueryStub.restore()
+      Sinon.stub(UserModel, 'query').throws(new Error('Model query error'))
+    })
+    it('throws the error', async () => {
```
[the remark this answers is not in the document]
ben-sagar: Done

PR #2245, opened 2025-08-07
Ad-hoc notice - select address

robertparkinson:

As part of sending out notifications to users we need to be able to select an address. The postcode was provided on the previous page and this page will have a list of addresses for the user to choose from. 

When chosen the information will then be converted into a recipient matching the existing structure retrieved from the database. It is then added to an array to be displayed to the user on the check page. It then returns the user to that check page.


app/services/notices/setup/create-additional-recipient.service.js
```diff
@@ -0,0 +1,85 @@
...
+    contact: {
+      name: session.contactName,
+      addressLine1: session.address.addressLine1,
+      addressLine2: session.address.addressLine2,
+      addressLine3: session.address.addressLine3,
+      addressLine4: session.address.addressLine4,
+      country: session.address.country,
+      postcode: session.address.postcode
+    },
+    contact_hash_id: _createMD5Hash(session),
```
Cruikshanks, line 34:
```suggestion
    contact_hash_id: _contactHashId(session),
```

app/services/notices/setup/create-additional-recipient.service.js
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * When querying the database on the select recipients page we return an md5 hash for each recipient which enables use to
+ * check for duplicates and remove them. In this function we do the same for the recipient being added manually.
+ *
+ * As there is only one name field in this journey we don't have the salutation or title so this may result in duplicates
+ * but they will be noticable on the screen to the user and they can adjust as needed.
+ *
+ * @param {object} session
+ *
+ * @returns {Promise<string>} - The md5 hash string of the address
```
Cruikshanks, line 67: It is not necessary to add `@params` and `@returns` properties to the comments for private functions, unless they are adding info not already covered/clear.

You just need to add the `@private` tag to avoid the linter shouting at you.

app/services/notices/setup/create-additional-recipient.service.js
Cruikshanks, line 1: We try to reserve the word 'create' for services that interact with the DB and result in a record being 'created'.

In this case, the service appears to focus on first generating, then adding an 'additional recipient' to the session. We don't need to take verbosity to the extreme, so I would suggest renaming the service as `add-additional-recipient.service.js`.

Cruikshanks (changes requested):

app/services/notices/setup/add-additional-recipient.service.js
```diff
@@ -0,0 +1,95 @@
+'use strict'
+
+/**
+ * Create an additional recipient from an address stored in session
```
Cruikshanks, line 4:
```suggestion
 * Add an 'additional recipient' to the notice setup session from address data captured by our shared address journey
```

app/services/notices/setup/add-additional-recipient.service.js
```diff
@@ -0,0 +1,95 @@
...
+ * @module AddAdditionalRecipientService
+ */
+
+const crypto = require('crypto')
+
+const SessionModel = require('../../../models/session.model.js')
+
+/**
+ * Create an additional recipient from an address stored in session
```
Cruikshanks, line 13:
```suggestion
 * Add an 'additional recipient' to the notice setup session from address data captured by our shared address journey
```

app/services/notices/setup/add-additional-recipient.service.js
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * When adding an additional recipient to a notice we need it to be in the same format as those returned from the databse.
+ * This service takes an address stored in the session from the address journey and creates an additional recipient object,
+ * then adds it to the additional recipients array if it exists or creates it if it doesn't.
+ *
+ * @param {string} sessionId
```
Cruikshanks, line 19:
```suggestion
 * @param {string} sessionId - The UUID of the current session
```

app/services/notices/setup/initiate-session.service.js
```diff
@@ -41,19 +41,36 @@ async function go(journey, noticeType = null, monitoringStationId = null) {
   const session = await SessionModel.query()
     .insert({
       data: {
+        address: {},
```
Cruikshanks, line 44:
```suggestion
```

app/services/notices/setup/submit-contact-type.service.js
```diff
@@ -39,7 +39,8 @@ async function go(sessionId, payload, yar) {
 
   const _submittedData = {
```
Cruikshanks, line 40: I know it is not in this PR. But just spotted that this var name needs to be corrected.

```suggestion
  const submittedData = {
```

test/controllers/address.controller.test.js
```diff
@@ -154,14 +155,18 @@ describe('Address controller', () => {
 
       describe('when the request succeeds', () => {
         beforeEach(() => {
-          Sinon.stub(SubmitSelectAddressService, 'go').returns({})
+          Sinon.stub(SubmitSelectAddressService, 'go').returns({
+            redirect: '/system/notices/setup/fecd5f15-bacf-4b3d-bdcd-ef279a97b061/check',
+            succeeded: true
+          })
+          Sinon.stub(AddAdditionalRecipientService, 'go').returns({})
```
Cruikshanks, line 162: The service doesn't return anything so we should reflect this in the test.

```suggestion
          Sinon.stub(AddAdditionalRecipientService, 'go')
```

PR #2246, opened 2025-08-07
Update notice setup return form PDF

jonathangoulding:

This change adds styling and structure to the upcoming return forms PDF work, with some mock data to render in the view. This makes it easier to swap out the legacy codes structure. A change will be made in the future to fetch this data. 

This change has copied the CSS from the legacy service. This was proven to work with Gotenberg in a previous spike. 

To ensure everything looks and works across multiple pages, the guidance page has been added to this change. This is just HTML with no additional data needed for the page. 

The header and footer have been added to test the layout works as expected, correct pagination has not been implemented and is mocked/ hard coded for now.


Cruikshanks (changes requested):
Just a typo, I know was in the legacy but we might as well correct.

PR #2256, opened 2025-08-12
Update Notice setup pdf return forms

jonathangoulding:

The legacy code calculates the page number manually using a Nunjucks filter. We have moved to Gotenberg to generate the PDFs, we use the chrome converter (forms/chromium/convert/html).

This enables use to get the page numbers and total count in a much simpler way, see - https://gotenberg.dev/docs/6.x/html#header-and-footer.

This change implements a generic footer using Gotenberg to generate the total page count and page number.


jonathangoulding:
I have assumed this will work when we implement the returns page which can span multiple pages woth check boxes

Cruikshanks (approved):

PR #2265, opened 2025-08-13
Update return forms PDF to use real return data

jonathangoulding:

This change updates the return forms PDF to use data from the selected return.

This change adds the return id to the path / controller. This allows the preview to be specific to the return selected on the selected returns page. When we implement the sending of the notice, we can pass in the return id to generate the PDF per return id (and recipient if required).

We needed to update the exciting fetch logic to return the additional data and check the return passed in has been selected from the journey.

Some general renaming has been done to update a previous assumption, The main one being the removal of the 'formatId' this is guaranteed to be the same as the 'returnReference' and has been removed.

The 'regionAndArea' will be updated in another change. This has some additional complexity to big for this change.

### DIFF
--- a/app/controllers/notices-setup.controller.js
+++ b/app/controllers/notices-setup.controller.js
@@ -205,9 +205,9 @@ async function viewNoticeType(request, h) {
 }
 
 async function viewPreviewReturnForms(request, h) {
-  const { sessionId } = request.params
+  const { sessionId, returnId } = request.params
 
-  const fileBuffer = await PreviewReturnFormsService.go(sessionId)
+  const fileBuffer = await PreviewReturnFormsService.go(sessionId, returnId)
 
   return h.response(fileBuffer).type('application/pdf').header('Content-Disposition', 'inline; filename="example.pdf"')
 }
--- a/app/presenters/notices/setup/prepare-return-forms.presenter.js
+++ b/app/presenters/notices/setup/prepare-return-forms.presenter.js
@@ -15,22 +15,29 @@ const { formatLongDate } = require('../../base.presenter.js')
  * Each page will be assigned a corresponding object to isolate the data to each page where possible. Those pages are:
  * - The "cover" page, this is the first page. The address is on this page.
  *
+ * @param {SessionModel} session - The session instance
+ * @param {object} dueReturn
+ *
  * @returns {object} - The data formatted for the return form
  */
-function go() {
+function go(session, dueReturn) {
+  const { licenceRef } = session
+
+  const { dueDate, endDate, purpose, returnsFrequency, returnReference, siteDescription, startDate, twoPartTariff } =
+    dueReturn
+
   return {
     address: _address(),
-    description: 'mock site',
-    dueDate: formatLongDate(new Date('2023-10-01')),
-    endDate: formatLongDate(new Date('2023-09-30')),
-    licenceRef: '123',
-    purpose: 'a purpose',
+    siteDescription,
+    dueDate: formatLongDate(new Date(dueDate)),
+    endDate: formatLongDate(new Date(endDate)),
+    licenceRef,
+    purpose,
     regionAndArea: 'A place / in the sun',
-    returnRef: '7646',
-    startDate: formatLongDate(new Date('2023-09-01')),
-    title: _title(),
-    twoPartTariff: true,
-    formatId: 'format id 123'
+    returnReference,
+    startDate: formatLongDate(new Date(startDate)),
+    pageTitle: _pageTitle(returnsFrequency),
+    twoPartTariff
   }
 }
 
@@ -43,15 +50,17 @@ function _address() {
     addressLine5: 'United Kingdom'
   }
 }
-/*
- * {{ 'Water abstraction daily return ' if frequency === 'day' }}
- * {{ 'Water abstraction weekly return ' if frequency === 'week' }}
- * {{ 'Water abstraction monthly return ' if frequency === 'month' }}
- * {{ 'Water abstraction quarterly return ' if frequency === 'quarter' }}
- * {{ 'Water abstraction yearly return ' if frequency === 'year' }}
- */
-function _title() {
-  return 'Water abstraction daily return'
+
+function _pageTitle(returnsFrequency) {
+  const mapper = {
+    day: 'daily',
+    month: 'monthly',
+    quarter: 'quarterly',
+    week: 'weekly',
+    year: 'yearly'
+  }
+
+  return `Water abstraction ${mapper[returnsFrequency]} return`
 }
 
 module.exports = {
--- a/app/routes/notices-setup.routes.js
+++ b/app/routes/notices-setup.routes.js
@@ -281,7 +281,7 @@ const routes = [
   },
   {
     method: 'GET',
-    path: '/notices/setup/{sessionId}/preview-return-forms',
+    path: '/notices/setup/{sessionId}/preview-return-forms/{returnId}',
     options: {
       handler: NoticesSetupController.viewPreviewReturnForms,
       auth: {
--- a/app/services/notices/setup/prepare-return-forms.service.js
+++ b/app/services/notices/setup/prepare-return-forms.service.js
@@ -13,18 +13,27 @@ const SessionModel = require('../../../models/session.model.js')
 /**
  * Orchestrates fetching and presenting the data for the return form
  *
- * @param {string} sessionId
+ * @param {string} sessionId - The UUID of the current session
+ * @param {string} returnId - The UUID of the return id
  *
  * @returns {Promise<ArrayBuffer>} - Resolves with the generated form file as an ArrayBuffer.
  */
-async function go(sessionId) {
+async function go(sessionId, returnId) {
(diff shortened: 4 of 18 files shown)


jonathangoulding:
> > _We needed to update the **exciting** fetch logic..._
> 
> If you say so 😝 !



updated....

app/presenters/notices/setup/prepare-return-forms.presenter.js
```diff
@@ -43,15 +50,17 @@ function _address() {
...
-  return 'Water abstraction daily return'
+
+function _pageTitle(returnsFrequency) {
+  const mapper = {
+    day: 'daily',
+    month: 'monthly',
+    quarter: 'quarterly',
+    week: 'weekly',
+    year: 'yearly'
+  }
```
Cruikshanks, line 61: This already exists in `app/lib/static-lookups.lib.js`. See `returnRequirementFrequencies`.

app/services/notices/setup/prepare-return-forms.service.js
```diff
@@ -13,20 +13,29 @@ const SessionModel = require('../../../models/session.model.js')
 /**
  * Orchestrates fetching and presenting the data for the return form
  *
- * @param {string} sessionId
+ * @param {string} sessionId - The UUID of the current session
+ * @param {string} returnId - The UUID of the return id
```
Cruikshanks, line 17: Being pedantic, but it isn't a UUID (unfortunately!)

Also, the legacy code often trips itself up by not being specific; for example, 'return' might refer to the return version, the return requirement, or the return log.

It's only a few more chars in this case, and one to bear in mind when referring to 'returns'.

```suggestion
 * @param {string} returnLogId - The return log ID
```

app/services/notices/setup/preview-return-forms.service.js
```diff
@@ -13,12 +13,13 @@ const PrepareReturnFormsService = require('./prepare-return-forms.service.js')
  *
  * This service returns the file to be display in the browser. This will likely be the built-in pdf viewer.
  *
- * @param {string} sessionId
+ * @param {string} sessionId - The UUID of the current session
+ * @param {string} returnId - The UUID of the return id
```
Cruikshanks, line 17:
```suggestion
 * @param {string} returnLogId - The return log ID
```

app/routes/notices-setup.routes.js
```diff
@@ -281,7 +281,7 @@ const routes = [
   },
   {
     method: 'GET',
-    path: '/notices/setup/{sessionId}/preview-return-forms',
+    path: '/notices/setup/{sessionId}/preview-return-forms/{returnId}',
```
Cruikshanks, line 284: This isn't going to work. Look at the returns tab and the view return log page and you'll see the problem we face trying to use the return id, for example, `v1:6:01/117:10749123:2025-04-01:2026-03-31`.

You'll either need to make this a query param, or pass the value in the body for a post request.

app/controllers/notices-setup.controller.js
```diff
@@ -205,9 +205,9 @@ async function viewNoticeType(request, h) {
 }
 
 async function viewPreviewReturnForms(request, h) {
-  const { sessionId } = request.params
+  const { sessionId, returnId } = request.params
```
Cruikshanks, line 208:
```suggestion
  const { sessionId, returnLogId } = request.params
```

app/controllers/notices-setup.controller.js
```diff
@@ -205,9 +205,9 @@ async function viewNoticeType(request, h) {
...
 
-  const fileBuffer = await PreviewReturnFormsService.go(sessionId)
+  const fileBuffer = await PreviewReturnFormsService.go(sessionId, returnId)
```
Cruikshanks, line 210:
```suggestion
  const fileBuffer = await PreviewReturnFormsService.go(sessionId, returnLogId)
```

app/presenters/notices/setup/prepare-return-forms.presenter.js
```diff
@@ -15,22 +15,29 @@ const { formatLongDate } = require('../../base.presenter.js')
  * Each page will be assigned a corresponding object to isolate the data to each page where possible. Those pages are:
  * - The "cover" page, this is the first page. The address is on this page.
  *
+ * @param {SessionModel} session - The session instance
+ * @param {object} dueReturn
```
Cruikshanks, line 19:
```suggestion
 * @param {object} dueReturnLog
```

app/presenters/notices/setup/prepare-return-forms.presenter.js
```diff
@@ -15,22 +15,29 @@ const { formatLongDate } = require('../../base.presenter.js')
...
+ *
  * @returns {object} - The data formatted for the return form
  */
-function go() {
+function go(session, dueReturn) {
+  const { licenceRef } = session
+
+  const { dueDate, endDate, purpose, returnsFrequency, returnReference, siteDescription, startDate, twoPartTariff } =
+    dueReturn
```
Cruikshanks, line 27:
```suggestion
function go(session, dueReturnLog) {
  const { licenceRef } = session

  const { dueDate, endDate, purpose, returnsFrequency, returnReference, siteDescription, startDate, twoPartTariff } =
    dueReturnLog
```

app/services/notices/setup/prepare-return-forms.service.js
```diff
@@ -13,20 +13,29 @@ const SessionModel = require('../../../models/session.model.js')
...
  *
  * @returns {Promise<ArrayBuffer>} - Resolves with the generated form file as an ArrayBuffer.
  */
-async function go(sessionId) {
+async function go(sessionId, returnId) {
```
Cruikshanks, line 21:
```suggestion
async function go(sessionId, returnLogId) {
```

app/services/notices/setup/preview-return-forms.service.js
```diff
@@ -13,12 +13,13 @@ const PrepareReturnFormsService = require('./prepare-return-forms.service.js')
...
  *
  * @returns {Promise<ArrayBuffer>} - Resolves with the generated form file as an ArrayBuffer.
  */
-async function go(sessionId) {
-  return PrepareReturnFormsService.go(sessionId)
+async function go(sessionId, returnId) {
+  return PrepareReturnFormsService.go(sessionId, returnId)
```
Cruikshanks, line 22:
```suggestion
async function go(sessionId, returnLogId) {
  return PrepareReturnFormsService.go(sessionId, returnLogId)
```

Cruikshanks (changes requested):
I was mainly focused on a NIT/pedantic point of trying to ensure we refer to 'returns' in code as 'return logs', i.e. it should be `returnLogId`.

But then I realised that I'm not sure how this is actually working, as there is no way to pass a return log ID as a path param to a POST request.

So, I've stopped there. Ping me if you need to discuss this/correct me!

app/controllers/notices-setup.controller.js
Cruikshanks: I am an idiot. We do have a UUID in the table now, and it is called `return_id`. I forgot we made this change 🤦

app/services/notices/setup/preview-return-forms.service.js
Cruikshanks: I am an idiot. We do have a UUID in the table now, and it is called `return_id`. I forgot we made this change 🤦

app/services/notices/setup/prepare-return-forms.service.js
Cruikshanks, line 17:
```suggestion
 * @param {string} returnId - The UUID of the return log
```

app/services/notices/setup/preview-return-forms.service.js
Cruikshanks, line 17:
```suggestion
 * @param {string} returnId - The UUID of the return log
```

test/controllers/notices-setup.controller.test.js
```diff
@@ -817,7 +817,7 @@ describe('Notices Setup controller', () => {
       beforeEach(() => {
         getOptions = {
           method: 'GET',
-          url: basePath + `/${session.id}/preview-return-forms`,
+          url: basePath + `/${session.id}/preview-return-forms/1234`,
```
Cruikshanks, line 820: I'd still use something more representative, in this case an actual UUID.

```suggestion
          url: basePath + `/${session.id}/preview-return-forms/95b54f97-fefb-46e7-aae8-ebf40ecb8b50`,
```

test/presenters/notices/setup/prepare-return-forms.presenter.test.js
```diff
@@ -12,14 +12,26 @@ const PrepareReturnFormsPresenter = require('../../../../app/presenters/notices/
 
 describe('Notices - Setup - Prepare Return Forms Presenter', () => {
   let session
+  let dueReturn
```
Cruikshanks, line 15:
```suggestion
  let dueReturnLog
```

Cruikshanks (changes requested):
Just NITS now.

PR #2276, opened 2025-08-14
Update the preview returns route to use the contact hash id

jonathangoulding:

We need to pass the contact hash id into the preview service to query the recipients (implemented the upcoming change to reduce complexity).

This will be used to find one recipient and will be passed into the prepare return form service. This will align with our end goal of the 'PrepareReturnFormsService' receiving a recipient to extract the address from.

This change is prep work for the upcoming change.


jonathangoulding:
Based on the required changes here - https://github.com/DEFRA/water-abstraction-system/pull/2274/files

PR #2299, opened 2025-08-15
Add dynamic meter readings to the return forms PDF

jonathangoulding:

When generating meter readings, we need to handle them spilling over onto another page.

The meter reading uses boxes in the PDF form, this is presented as rows and columns.

This change calculates the number of dates we put on a single page; when this limit is reached, a new page will be added below the existing, and this continues.

We use the legacy codes CSS and some of the HTML to show the pages correctly, legacy code has stipulated the columns and rows as the expected format when rendering the PDF.


app/presenters/notices/setup/prepare-return-forms.presenter.js
```diff
@@ -9,6 +9,21 @@ const NotifyAddressPresenter = require('./notify-address.presenter.js')
 const { formatLongDate } = require('../../base.presenter.js')
 const { naldAreaCodes, returnRequirementFrequencies } = require('../../../lib/static-lookups.lib.js')
 
+const RETURN_TYPE = {
+  week: {
+    columns: 2,
+    columnSize: 14
```
jonathangoulding, line 15: This can be called 'rows' if we can clearly illustrate that a row is not across the page but confined to the column. So a page with 2 columns would have two rows 14 vertically. 

The data is top down and not across.

test/presenters/notices/setup/prepare-return-forms.presenter.test.js
```diff
@@ -89,64 +91,220 @@ describe('Notices - Setup - Prepare Return Forms Presenter', () => {
...
-
-        it('should return the relevant title', () => {
+      describe('the "meterReadings" property', () => {
+        it('should return meter readings', () => {
           const result = PrepareReturnFormsPresenter.go(session, dueReturnLog, recipient)
 
-          expect(result.pageTitle).to.equal('Water abstraction quarterly return')
+          expect(result.meterReadings).to.equal([
+            [
+              [
```
jonathangoulding, line 127: Either adds comments to columns / rows.

Or destructure the result meterReadings into args - 
```
const [firstPage] = meterReadings

const [firstColumnRows, secondColumnsRows] = firstPage

```

Cruikshanks (changes requested):
Just some suggested tweaks and NITS.

test/lib/general.lib.test.js
```diff
@@ -380,6 +377,51 @@ describe('GeneralLib', () => {
...
+      })
+    })
+
+    describe('when the group size is 0', () => {
+      beforeEach(() => {
+        testArray = [1, 2, 3, 4, 5, 6, 7]
+        testGroupSize = 0
+      })
+
+      it('returns an empty array', () => {
```
Cruikshanks, line 402: It's not returning an empty array

Cruikshanks (approved):

PR #2308, opened 2025-08-20
Add additional recipient - use route to save details

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-5047

As part of sending out notifications to users we need to be able to submit an address. 

When the address has been submitted it is converted into a recipient object and added to the additionalRecipient array as well as the hash added to the selectedRecipient array.

This PR refactors the way the address is handled before being passed back to the check recipients page.


test/controllers/address.controller.test.js
```diff
@@ -156,17 +155,17 @@ describe('Address controller', () => {
...
           Sinon.stub(SubmitSelectAddressService, 'go').returns({
-            redirect: '/system/notices/setup/fecd5f15-bacf-4b3d-bdcd-ef279a97b061/check',
-            succeeded: true
+            redirect: '/system/notices/setup/fecd5f15-bacf-4b3d-bdcd-ef279a97b061/add-recipient'
           })
-          Sinon.stub(AddAdditionalRecipientService, 'go')
         })
 
-        it('redirects to the check page', async () => {
+        it('redirects to the add additional recipient route', async () => {
```
Cruikshanks, line 162: I think we can be generic in our test description here.

```suggestion
        it('redirects to the configured route', async () => {
```

test/controllers/address.controller.test.js
```diff
@@ -243,16 +239,17 @@ describe('Address controller', () => {
...
         beforeEach(() => {
           Sinon.stub(SubmitManualAddressService, 'go').returns({
-            redirect: '/system/notices/setup/fecd5f15-bacf-4b3d-bdcd-ef279a97b061/check'
+            redirect: '/system/notices/setup/fecd5f15-bacf-4b3d-bdcd-ef279a97b061/add-recipient'
           })
-          Sinon.stub(AddAdditionalRecipientService, 'go')
         })
 
-        it('redirects to the check page', async () => {
+        it('redirects to the add additional recipient route', async () => {
```
Cruikshanks, line 246:
```suggestion
        it('redirects to the configured route', async () => {
```

Cruikshanks (changes requested):
Just a minor NIT

PR #2309, opened 2025-08-20
Add dynamic meter readings to the return forms PDF for days

jonathangoulding:

We previously implemented the month and week meter readings.

This change adds the meter readings for the 'days' frequency.

When the return frequency is 'days', we show three columns (and set three columns in the HTML dues to the way the CSS works).

The day meter readings show the Month and Year at the top of each column, and each cell has the day and then the boxes. it looks something like this

June 2024
1 [][][][][][][][][][]
2 [][][][][][][][][][]

This differs from out established logic, so it has been implemented on its own within the presenter.

This change uses (and follows) the previously established calculation for the number of dates we put on a single page, and when this limit is reached, a new page will be added below the existing, and this continues.

We use the legacy codes CSS and some of the HTML to show the pages correctly, legacy code has stipulated the columns and rows as the expected format when rendering the PDF.

### DIFF
--- a/app/presenters/notices/setup/prepare-return-forms.presenter.js
+++ b/app/presenters/notices/setup/prepare-return-forms.presenter.js
@@ -79,7 +79,7 @@ function _address(recipient) {
 /**
  * Day meter readings are structured differently to 'month' and 'week'
  *
- * The 'days' will take up three columns per page, with the month and year are at the top of each column - 'June 2024'
+ * The 'days' will take up three columns per page, with the month and year at the top of each column - 'June 2024'
  * and with the day next to the checkbox. It will look something like this:
  *
  * ```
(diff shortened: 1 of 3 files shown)


app/presenters/notices/setup/prepare-return-forms.presenter.js
```diff
@@ -104,6 +134,37 @@ function _formatPeriodsToLongDate(periods) {
...
+ *
+ * @private
+ */
+function _groupDaysIntoMonths(dates) {
+  return dates.reduce((acc, date) => {
+    const key = date.toLocaleString('default', { month: 'long', year: 'numeric' })
+
+    let group = acc.find((group) => {
+      return group.period === key
+    })
```
Cruikshanks, line 152: SonarQube is complaining because `group` has been declared twice within this function. Here is my suggested quick fix.

```suggestion
    let group = acc.find((existingGroup) => {
      return existingGroup.period === key
    })
```

Cruikshanks (changes requested):
Sorry, approved, then checked the SonarQube issue and realised it was right.

Cruikshanks (approved):

PR #2320, opened 2025-08-22
Update the standard layout to provide the error summary, back link and page heading html.

jonathangoulding:
https://github.com/DEFRA/water-abstraction-team/issues/149

To use the updated layout and keep the old pages working, we have introduced a new block called 'pageContent'. This should generally be used to show the form content of a page.

If a page needs to control the entire content, then the 'content' block can still be used. This also allows existing pages to still function as expected.

The backlink works on the same principle where the current pages will override the 'breadcrumbs' block. When the new 'pageContent' block is used, the override can be removed, but the backlink must now provide an object with the text and href.

The error messages will be provided by the submit logic for a page, and they will provide the 'errorList'.

### DIFF
--- a/app/views/layout.njk
+++ b/app/views/layout.njk
@@ -1,5 +1,10 @@
 {% extends "govuk/template.njk" %}
+
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
 {% from "govuk/components/phase-banner/macro.njk" import govukPhaseBanner %}
+{% from 'govuk/components/error-summary/macro.njk' import govukErrorSummary %}
+
+{% from "macros/page-heading.njk" import pageHeading %}
 
 {% block pageTitle %}{{ pageTitle }} - GOV.UK{% endblock %}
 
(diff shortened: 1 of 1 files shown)


jonathangoulding:
Example for working is here - https://github.com/DEFRA/water-abstraction-system/pull/2318

Cruikshanks (changes requested):
You aligned some of the other elements in https://github.com/DEFRA/water-abstraction-system/pull/2318 like the `stylesheet` link and the `govukPhaseBanner` component.

I'd take advantage of this change to also do those housekeeping bits here.

app/views/layout.njk
```diff
@@ -46,11 +51,29 @@
...
+    [{ text: 'An error', href: "#link-to-form-item"}]  #}
+  {% if errorList %}
+    {{ govukErrorSummary({
+      titleText: "There is a problem",
+      errorList: errorList
+    }) }}
+  {% endif %}
+
+  {# We set the page heading as a variable in the view to ensure all pages are consistent use this in your form field
+    when setting 'isPageHeading: true' #}
```
robertparkinson: it's the other way around. if you don't want/need to use the default one from the macro then set `isPageHeading: true` and it will create the h1 otherwise leave it off and it will create a label

https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/label/template.njk#L12

PR #2321, opened 2025-08-22
Address lookup - handle null premises and street address

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-5046

As part of the testing for the select address page QA found some addresses that had no premises or street_address which the current code didn't handle correctly. 

This PR adds logic to check for null values for those fields.

### DIFF
--- a/app/services/address/submit-select.service.js
+++ b/app/services/address/submit-select.service.js
@@ -65,14 +65,14 @@ async function _save(session, address) {
   const premises = address.premises ?? ''
   const streetAddress = address.street_address ?? ''
 
-  const premisesStreetAddress = premises + ' ' + streetAddress
+  const premisesStreetAddress = `${premises} ${streetAddress}`.trim()
 
   if (!address.organisation) {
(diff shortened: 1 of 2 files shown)


app/services/address/submit-select.service.js
```diff
@@ -62,14 +62,17 @@ async function go(sessionId, payload) {
 async function _save(session, address) {
   session.address.uprn = address.uprn
 
-  const premiseseStreetAddress = address.premises + ' ' + address.street_address
+  const premises = address.premises ?? ''
+  const streetAddress = address.street_address ?? ''
+
+  const premisesStreetAddress = premises + ' ' + streetAddress
```
Cruikshanks, line 68: How about putting the `trim()` and using interpolation?

```suggestion
  const premisesStreetAddress = `${premises} ${streetAddress}`.trim()
```

We don't usually use concatenation, and this avoids repeating the call to `trim()` below.

Cruikshanks (approved):

PR #2322, opened 2025-08-22
Update download recipients to handle additional and selected recipients

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5223

The logic to handle the recipients has become quite complicated. We need to simplify the handling recipients. 

We now have additional and selected recipients we need to merge into the recipients array. We need to do this in multiple places. 

This change adds a new service to handle merging our recipients form the database and those manually added by the user (additional recipients). This will be the common place to get the recipients. 

As part of this change we have refactor some of our service names to align with the new change. 

This change also fixes the download by using this new'RecipientsService' to add the additional recipients and selected recipients to the CSV.


app/services/notices/setup/fetch-download-recipients.service.js
```diff
@@ -123,7 +124,12 @@ FROM (
     rl.return_reference,
     rl.start_date,
     rl.end_date,
-    rl.due_date
+    rl.due_date,
+    (md5(
```
jonathangoulding, line 128: Required when using the selected recipients.

app/services/notices/setup/recipients-add.service.js
```diff
@@ -0,0 +1,45 @@
+'use strict'
+
+/**
+ * Orchestrates the recipients with additional and selected recipients
+ * @module RecipientsAddService
```
jonathangoulding, line 5: Name obviously needs work

app/services/notices/setup/recipients-add.service.js
```diff
@@ -0,0 +1,45 @@
...
+  const recipients = _additionalRecipients(recipientsData, session)
+
+  if (allRecipients || !session.selectedRecipients) {
+    return recipients
+  }
+
+  return _selectedRecipients(session.selectedRecipients, recipients)
+}
+
+function _additionalRecipients(recipientsData, session) {
```
jonathangoulding, line 29: Additional recipients do not have the return period in them. 

We use the determinedReturnsPeriod when sending a notification. But the date from the query when the user has selected a return period. 

There might be a bug here where the download shows a different start/ end period than what is sent in the notification. 

We would need to check what is being sent as the start and end date for the notice. They are definitely different.

app/services/notices/setup/all-recipients.service.js
```diff
@@ -0,0 +1,43 @@
+'use strict'
+
+/**
+ * Orchestrates merging the recipients with additional and selected recipients
+ * @module AllRecipientsService
```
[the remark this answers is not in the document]
jonathangoulding: This needs to be used in the submit ?

Cruikshanks (approved):

PR #2341, opened 2025-09-01
Update download recipients to show a different csv for ad hoc

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5223

We previously used the same CSV for the ad hoc journey. Since introducing the selected and additional recipients, we need to update the CSV to show less than the 'full' version.

When the journey is ad hoc we have the licence already set and the return period is no required.

This change removes the return periods and return reference from the ad hoc download presenter.


Cruikshanks (changes requested):
Just NITS.

Cruikshanks (approved):

PR #2352, opened 2025-09-03
Set due date to null when creating return logs

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-5212

To accommodate new legislation we need to track when a customer has been notified to submit their return logs. To do that we need to not set the due date when we create the return logs. 

This PR adds in the logic to create a return log and leave the due date null. This will be set once we have confirmed that the user has been successfully notified of their returns.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -300,34 +300,34 @@
       }
     },
     "node_modules/@aws-sdk/client-s3": {
-      "version": "3.879.0",
-      "resolved": "https://registry.npmjs.org/@aws-sdk/client-s3/-/client-s3-3.879.0.tgz",
-      "integrity": "sha512-1bD2Do/OdCIzl72ncHKYamDhPijUErLYpuLvciyYD4Ywt4cVLHjWtVIqb22XOOHYYHE3NqHMd4uRhvXMlsBRoQ==",
+      "version": "3.882.0",
+      "resolved": "https://registry.npmjs.org/@aws-sdk/client-s3/-/client-s3-3.882.0.tgz",
+      "integrity": "sha512-0IrBUOrBepQeuH025t+b4KqgBRQT+B//JlTU3+629WUGWwsWVfFkCTkn4xK/oQP9/K6npZtfDTuO6XfXSLimmg==",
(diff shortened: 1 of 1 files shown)


app/services/jobs/return-logs/check-return-cycle.service.js
```diff
@@ -11,6 +11,7 @@ const {
   determineCycleEndDate,
   determineCycleStartDate
 } = require('../../../lib/return-cycle-dates.lib.js')
+const FeatureFlagsConfig = require('../../../../config/feature-flags.config.js')
```
Cruikshanks, line 14: This needs to go below the other `requires()`. See [Top of .js files](https://github.com/DEFRA/water-abstraction-team/blob/main/coding_conventions.md#top-of-js-files)

```suggestion
const featureFlagsConfig = require('../../../../config/feature-flags.config.js')
```

app/services/return-logs/generate-return-log.service.js
```diff
@@ -5,6 +5,7 @@
  * @module GenerateReturnLogService
  */
 
+const FeatureFlagsConfig = require('../../../config/feature-flags.config.js')
```
Cruikshanks, line 8: This needs to go below the other `requires()`. See [Top of .js files](https://github.com/DEFRA/water-abstraction-team/blob/main/coding_conventions.md#top-of-js-files)

```suggestion
const featureFlagsConfig = require('../../../config/feature-flags.config.js')
```

test/services/jobs/return-logs/check-return-cycle.service.test.js
Cruikshanks, line 1: I'm happy that we ignore testing for when a feature flag is true or false.

The flag is only temporary and the intent is to move to a place where the new feature is the norm.

So feel free to update the tests to assume the flag is set to true (by stubbing it at the top!) and drop the tests for when it is false.

Cruikshanks (changes requested):
Just some NITs, and the ok to delete a bunch of stuff! 😁

PR #2353, opened 2025-09-03
Returns tab - Fix incorrect status

Jozzey:
https://eaflood.atlassian.net/browse/WATER-5213

During QA of the work done in https://github.com/DEFRA/water-abstraction-system/pull/2346 it has been spotted that when the status is "void", and the due date is `null`. The status is incorrectly being displayed as "Not due yet".

This PR wil fix that issue.


app/presenters/licences/view-licence-returns.presenter.js
```diff
@@ -6,9 +6,7 @@
  */
 
 const FeatureFlagsConfig = require('../../../config/feature-flags.config.js')
-const { formatLongDate, formatPurposes } = require('../base.presenter.js')
-
-const DUE_PERIOD_DAYS = 27
+const { formatLongDate, formatPurposes, formatReturnLogStatus } = require('../base.presenter.js')
```
robertparkinson, line 9: can we not just import it from the base-return-logs.presenter? why do we need to move this into the base.presenter?  it's only needed when we're doing return logs so best to keep it in with all the other return log functions in the base-return-logs.presenter

robertparkinson (changes requested):

app/presenters/licences/view-licence-returns.presenter.js
Jozzey: I didn't move it for fun 😆

Cruikshanks: @robertparkinson It was me who directed @Jozzey to move it up to the `base.presenter.js` for the reason he stated. It's now used in `/return-logs` and `/licences`, which previously has dictated we bump it up to indicate it is used across domains.

PR #2355, opened 2025-09-03
Update notice setup recipient name to show name and error

jonathangoulding:
When sending a paper form, we want to allow the user to send the form to an additional recipient.

Our existing page offers a choice between email and post (a letter). We do not want to show the email option to the user.

The simplest way we can do this is by adding a standalone page to enter the recipient name. This will redirect to the generic address service the same as the existing screen.

This change updates the previously scaffolded code to function as expected.

The view will render the name if filled in and error accordingly.

The submit will redirect to the existing generic address journey. It sets the backlink so the journey knows where to come back to.

### DIFF
--- a/app/presenters/notices/setup/recipient-name.presenter.js
+++ b/app/presenters/notices/setup/recipient-name.presenter.js
@@ -8,10 +8,18 @@
 /**
  * Formats data for the '/notices/setup/{sessionId}/recipient-name' page
  *
+ * @param {SessionModel} session - The session instance
+ *
  * @returns {object} - The data formatted for the view template
  */
-function go() {
-  return {}
+function go(session) {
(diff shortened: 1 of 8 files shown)


app/presenters/notices/setup/recipient-name.presenter.js
```diff
@@ -8,10 +8,18 @@
...
  */
-function go() {
-  return {}
+function go(session) {
+  const { id: sessionId, contactName: name } = session
+
+  return {
+    backLink: { text: 'Back', href: `/system/notices/setup/${sessionId}/check` },
+    name,
+    pageTitle: "Enter the recipient's name"
```
jonathangoulding, line 21: add page caption

PR #2356, opened 2025-09-03
Amend error in layout and add new error formatter

Cruikshanks:
https://github.com/DEFRA/water-abstraction-team/issues/149

We've been working on how to be more consistent with our views and the way we apply them. As part of this, we've amended the base `app/views/layout.njk` to include an [Error summary](https://design-system.service.gov.uk/components/error-summary/) component. This will avoid the need to add the component into every view that needs it, with the risk that brings of folks implementing it differently!

The GDS component has an `errorList:` property which expects an array of objects.

```javascript
{{ govukErrorSummary({
  titleText: "There is a problem",
  errorList: [
    {
      text: "Enter your full name",
      href: "#name"
    },
    {
      text: "Enter your alias",
      href: "#alias"
    }
  ]
}) }}
```

We intended to start generating this 'error list' in our submit services and pass it through to the view as `errorList`.

```nunjucks
{% if errorList %}
  {{ govukErrorSummary({ titleText: "There is a problem", errorList: errorList }) }}
{% endif %}
```

What we've realised is that, in some cases, we would need to provide this in _addition_ to our existing `error:` property, rather than _instead of_. This is because the input components still need to know the following.

- Whether to display an error
- What message to display

For a page with a single component that often looks like this.

```nunjucks
  {% if error %}
    {{ govukErrorSummary({
      titleText: "There is a problem",
      errorList: [
        { text: error.text, href: "#frequencyCollected" }
      ]
    }) }}
  {%endif%}

  {{ govukRadios({
    name: "frequencyCollected",
    errorMessage: error,
    fieldset: {
      legend: {  html: pageHeading }
    },
    items: [
      { value: "day", text: "Daily", checked: "day" === frequencyCollected },
      { value: "week", text: "Weekly", checked: "week" === frequencyCollected },
      { value: "month", text: "Monthly", checked: "month" === frequencyCollected }
    ]
  }) }}
```

Here, switching out `error` with `errorList` would work because in the view, we could refer to `errorList[0].text`.

The problem is pages that feature multiple components. For those, the `error` object we pass in has properties we can directly reference in each component.

```nunjucks
  {% if error %}
    {{ govukErrorSummary({
      titleText: "There is a problem",
      errorList: error.errorList
    }) }}
  {% endif %}

  {{ govukInput({
    label: { text: "Name", isPageHeading: false },
    id: "name",
    name: "name",
    value: name,
    attributes: { 'data-test': 'name' },
    errorMessage: {
      text: error.name.message
    } if error.name
  }) }}

  {{ govukInput({
    label: { text: "Alias", isPageHeading: false },
    id: "alias",
    name: "alias",
    value: alias,
    attributes: { 'data-test': 'alias' },
    errorMessage: {
      text: error.alias.message
    } if error.alias
  }) }}
```

This supports only one of the components having a validation error, or both. We can't switch `error` for `errorList` because we need the named properties. `error` has an `errorList:` property, but for `layout.njk` to see it, we need to pass it in separately.

😩😢 

The good news is we believe we have a solution that works for both types of pages, and avoids needing to pass `error` and `errorList` separately.

The way we generate the `error` object from the [Joi](https://joi.dev/api/?v=17.13.3) validation result for pages with multiple components is actually generic. It can be moved to a shared helper function that all submit services could then use, regardless of whether they are validating input from one component or multiple.

This would mean `error` would have a consistent format, as well as contain an `error.errorList` property.

This change tweaks `layout.njk` and adds the helper function in preparation for new pages, and those we refactor to use this shared functionality.

### DIFF
--- a/app/views/layout.njk
+++ b/app/views/layout.njk
@@ -53,8 +53,8 @@
 {% endblock %}
 
 {% block content %}
-  {% if errorList %}
-    {{ govukErrorSummary({ titleText: "There is a problem", errorList: errorList }) }}
+  {% if error %}
+    {{ govukErrorSummary({ titleText: "There is a problem", errorList: error.errorList }) }}
   {% endif %}
 
   {% set pageHeadingHtml = pageHeading(pageTitle, pageTitleCaption) %}
--- a/app/presenters/base.presenter.js
+++ b/app/presenters/base.presenter.js
@@ -256,6 +256,117 @@ function formatPurposes(purposes) {
 }
 
 /**
+ * Format a Joi validation result into a format that can be used by our views
+ *
+ * If {@link https://joi.dev/ | Joi} found no validation errors the result will not contain an `error:` property.
+ *
+ * ```javascript
+ * {
+ *   value: {
+ *     noticeTypes: [ 'returnReminder', 'returnInvitation' ],
+ *     reference: 'RREM-VB9EFA',
+ *     sentFromDay: '01',
+ *     sentFromMonth: '09',
+ *     sentFromYear: '2025',
+ *     fromDate: 2025-09-01T00:00:00.000Z,
+ *     toDate: undefined
+ *   }
+ * }
+ * ```
+ * When this is the case we return null.
+ *
+ * If there is an `error` property, it will contain details of error found.
+ *
+ * ```javascript
+ * {
+ *   value: {
+ *     noticeTypes: [ 'returnReminder', 'returnInvitation' ],
+ *     reference: 'jkdfshfhkfhsdjkfsdjkfghadshj',
+ *     sentFromDay: '01',
+ *     sentFromMonth: '09',
+ *     fromDate: '-09-01',
+ *     toDate: undefined
+ *   },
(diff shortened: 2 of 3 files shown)


app/views/layout.njk
```diff
@@ -53,8 +53,8 @@
 {% endblock %}
 
 {% block content %}
-  {% if errorList %}
-    {{ govukErrorSummary({ titleText: "There is a problem", errorList: errorList }) }}
+  {% if error %}
+    {{ govukErrorSummary({ titleText: "There is a problem", errorList: error.errorList }) }}
```
jonathangoulding, line 57: should this be error.summary instead of error.errorList as it is specific to this component ? 

Or errorSummaryList

Cruikshanks: If pushed, I'd go with `errorSummaryList`. Or even `error.summary.errorList` (adds scope to allow us to pass other properties to the summary in the future).

However, I prefer to keep it as `errorList` for now because it matches the name of the property we're populating. 🤷

PR #2357, opened 2025-09-03
Update generic address back link

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5234

We need to handle various backlinks for the generic address journey.

Rather than have conditional logic in the address postcode presenter to determine the backlink we will instead provide the backlink in the session; this is provided by the previous page.

This change updates the backlink for the address presenter to use the provided session backlink.

We only have the contact type page redirecting to this page at the moment, and so it is the only one that's needs to be updated to provide the backlink in the session.

### DIFF
--- a/app/presenters/address/postcode.presenter.js
+++ b/app/presenters/address/postcode.presenter.js
@@ -2,7 +2,7 @@
 
 /**
  * Formats data for the `address/{sessionId}/postcode` page
- * @module ManualAddressPresenter
+ * @module PostcodePresenter
  */
 
 /**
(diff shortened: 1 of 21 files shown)


app/services/notices/setup/submit-contact-type.service.js
```diff
@@ -103,6 +103,11 @@ async function _save(session, payload, yar) {
   session.contactName = payload.name
   session.contactType = payload.type
 
+  session.backLink = {
```
robertparkinson, line 106: this should be added in the initiate session service next to where we set the redirect url: 
https://github.com/DEFRA/water-abstraction-system/blob/main/app/services/notices/setup/initiate-session.service.js#L67

robertparkinson (changes requested):

app/presenters/address/postcode.presenter.js
```diff
@@ -22,14 +22,18 @@ function go(session) {
...
+ * We achieve this by having the previous page provide the backlink in the session.
+ *
+ * The default return to the postcode page is a placeholder until the address lookup journey is expanded.
  *
  * @private
  */
 function _backLink(session) {
-  if (session.contactName) {
-    return `/system/notices/setup/${session.id}/contact-type`
+  if (session.backLink) {
```
robertparkinson, line 35: this should be `session.address.backLink`

robertparkinson (approved):

PR #2358, opened 2025-09-03
Refactor view notice page to template standard

Cruikshanks:
DEFRA/water-abstraction-team#149

We've been working on becoming more consistent in our views and the way we apply them. As part of this, we've amended the base `app/views/layout.njk` to include an [Error summary](https://design-system.service.gov.uk/components/error-summary/) component. This will avoid the need to add the component into every view that needs it, with the risk that brings of folks implementing it differently!

To support this, we've also [added a new validation result formatter](#2356) that submit services can use to ensure the validation errors we pass to views support the changes made to `layout.njk`.

Having added the helper, we wanted to provide an example of its use. So, this change refactors the view notice page to "template" standard, which incorporates using the new helper.

### DIFF
--- a/app/controllers/notices.controller.js
+++ b/app/controllers/notices.controller.js
@@ -7,7 +7,7 @@
 
 const IndexNoticesService = require('../services/notices/index-notices.service.js')
 const NoticesViewService = require('../services/notices/view.service.js')
-const SubmitIndexNoticesService = require('../services/notices/submit-index.service.js')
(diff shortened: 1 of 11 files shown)


Cruikshanks:
There were several housekeeping tasks to complete as part of this. But changes specific to the "template" standard are primarily.

- using the new `formatValidationResult()` in the submit service
- the changes to the view (use `pageContent`, dropping pointless extra divs, and how the `error` is used)
-

Jozzey (approved):
👍🏼

PR #2364, opened 2025-09-04
Update flash notification to support the updated layout

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5234

We have updated the layout to include the error summary and backlink.

We can also provide the notification banner.

This change adds the notification banner to the updated layout, and updates the existing notifications  to align with the improved notification helper. 

This should ensure all notifications follow the same pattern

We are updating the way we use the notifications by having the layout automatically provide the GDS Notification banner macro.

We used to do this in the views: 
```html
{{ govukNotificationBanner({
     titleText: notification.title,
     text: notification.text
}) }}
```

Now we provide the notification to the macro:
```
{% if notification %}
  {{ govukNotificationBanner(notification) }}
{% endif %}
```

In time we will migrate the existing pages to use the improved 'pageContent'.


app/lib/general.lib.js
```diff
@@ -101,7 +101,32 @@ function determineCurrentFinancialYear() {
...
+ * banner macro.
+ *
+ * This is how we provide the notification macro:
+ * ```
+ * {% if notification %}
+ *   {{ govukNotificationBanner(notification) }}
+ * {% endif %}
+ * ```
+ *
+ * We are in the process of migrating pages to use this improved layout. But we need to support the existing way of
```
jonathangoulding, line 116: Update this

PR #2400, opened 2025-09-16
Add return forms to batch notifications

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5226

We have implemented a service to generate the PDF file for the return forms ('DetermineReturnFormsService'). This service needs to generate the PDF file and the notification data (through a presenter).

This is different to our existing pattern where we just use the presenters to return the notification. This is necessary to generate the PDF file.

Other than this deviation, we can add the return forms to the batch notifications following the existing pattern.

This change adds the logic to send and save a return forms notification.

This change also updates the 'DetermineReturnFormsService' to return the expected data in the notification table. This aligns with our existing notification presenters. This will get its own presenter in another change.


app/services/notices/setup/determine-return-forms.service.js
```diff
@@ -32,8 +32,12 @@ async function go(session, recipients, eventId) {
 
       notifications.push({
         content: returnForm,
+        eventId,
+        licences: JSON.stringify([licenceRef]),
```
Cruikshanks, line 36: Have you tried setting `licences` as a JSON attribute on the model (app/models/notification.model.js)?

> We may as well add `metadata` whilst we're at it!

```javascript
  static get jsonAttributes() {
    return ['licences', 'metadata', 'personalisation']
  }
```

Once added, it should then be you wouldn't need to stringify objects before persisting.

```suggestion
        licences: [licenceRef],
```

But I could be wrong!

Cruikshanks (changes requested):
I could be wrong about the JSON attributes thing; `licences` is a little different just being an array.

But [the theory](https://vincit.github.io/objection.js/api/model/static-properties.html#static-jsonattributes) is if set, **Objection.js** will handle parsing things to and from the DB.

Cruikshanks (approved):

PR #2406, opened 2025-09-17
Add presenter to format the return forms data into a notification

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5226

We have implemented the logic to send and save a return forms notification. This was already saving the raw page data. But this is not what we needed to save and was a placeholder.

This change adds the presenter for the return forms notification.

A return forms notification differs from our other notification in that is sending a file. This is passed around as an array buffer. This is passed to Notify as 'content' so we map is in the presenter.

The rest of the presenter shapes the data to conform to our existing pattern (and legacy code).

This change has required an update to the fetch to provide the return log id and other region data.

We took this opportunity to introduce a 'dueReturnLog' fixture which aligns to our fetch response. This has allowed us to remove all the due return log objects in our tests.


app/presenters/notices/setup/determine-return-forms.presenter.js
```diff
@@ -0,0 +1,77 @@
...
+ * Formats recipients into notifications for a return form
+ * @module DetermineReturnFormsPresenter
+ */
+
+/**
+ * Formats recipients into notifications for a return form
+ *
+ * The 'pageData' this function receives is used when we create the PDF file. We have updated some of the keys and
+ * reduced all the complex logic set i in the view into its presenter. That means the data is not in the shape the
+ * legacy code expect. So we need to map this 'pageData' into the legacies expected 'personalisation'
```
Cruikshanks, line 13:
```suggestion
 * The 'pageData' is the result of the presenter used when creating the PDF file. It uses different keys, and returns
 * different values specific to creating the file. We cannot just save its 'personalisation' to the DB as it differs
 * from what the legacy code persisted and uses. So, we use this presenter to map that output to what is needed when
 * saving the notification.
```

test/presenters/notices/setup/determine-return-forms.presenter.test.js
```diff
@@ -0,0 +1,99 @@
...
+      returnsFrequency: 'day',
+      siteDescription: 'Water park',
+      startDate: '1 January 2025',
+      twoPartTariff: false,
+      returnLogId
+    }
+  })
+
+  describe('when called', () => {
+    it('returns a notification for each combination of recipient and selected due return log', () => {
```
Cruikshanks, line 65: No it doesn't! 😂 

```suggestion
    it('returns the data re-formatted as a notification', () => {
```

app/presenters/notices/setup/determine-return-forms.presenter.js
Cruikshanks, line 1: I know this is used by the `DetermineReturnsFormsService`. However, I don't think we should name the presenter `Determine*`. I generally save that terminology for a service that is doing work to 'determine' a result.

Perhaps `return-forms-notification.presenter.js`? 🤷

jonathangoulding: Yep thats better

PR #2414, opened 2025-09-18
Update the notification model to convert 'licences' to json

jonathangoulding:
The legacy code stringify's the licences when they save notification, we followed this pattern so the licences could be saved as jsonb.

However, we can use objection to handle this for us.

This change sets the 'licence' as a json attribute for the notification model. This allows objection to handle the input as json.


Cruikshanks (changes requested):
Teeny-weeny NIT 😁

PR #2440, opened 2025-09-24
Prevent empty return submissions

Jozzey:
https://eaflood.atlassian.net/browse/WATER-5192

Currently it is possible to step through the return submission process 'enter & submit' but to not enter any volumes/meter readings and then press confirm.  The total return volume appears as 0m3 but all the return lines within the submitted return are blank.

This is not desirable behaviour, all the lines should not be allowed to be blank. There should be at least one entry, even if it is a zero.

In this PR validation is going to be added to the check page to prevent the submission of returns when all the lines are blank.


app/services/return-logs/setup/submit-check.service.js
```diff
@@ -37,20 +68,22 @@ async function go(sessionId, user) {
...
-  await SessionModel.query(trx).deleteById(sessionId)
+  if (!validation.error) {
+    return null
+  }
+
+  const { message } = validation.error.details[0]
+
+  return {
+    errorList: [{ text: message }]
+  }
```
Cruikshanks, line 86: The convention now is to use the new helper.

```suggestion
  return formatValidationResult(validationResult)
```

app/services/return-logs/setup/submit-check.service.js
```diff
@@ -27,6 +29,35 @@ const SessionModel = require('../../../models/session.model.js')
  */
 async function go(sessionId, user) {
   const session = await SessionModel.query().findById(sessionId)
+
+  const validationResult = _validate(session)
+
+  if (!validationResult) {
```
Cruikshanks, line 35: The convention now is to just name the result `error`.


```suggestion
  const error = _validate(session)

  if (!error) {
```

app/services/return-logs/setup/submit-check.service.js
```diff
@@ -27,6 +29,35 @@ const SessionModel = require('../../../models/session.model.js')
...
+    await _save(session, user)
+
+    return { returnLogId: session.returnLogId }
+  }
+
+  const formattedData = CheckPresenter.go(session)
+
+  return {
+    activeNavBar: 'search',
+    error: validationResult,
```
Cruikshanks, line 45:
```suggestion
    error,
```

Cruikshanks (changes requested):
I've got some NITS to update this to the _very latest_ template standard.

But unfortunately, this is not the validation they were after.

It is not that the submitted return has a total of 0. It's because the lines were _blank_.

So, it is perfectly acceptable to submit a return where you enter a zero on just one of the lines. The reason is that it might be the only applicable line for the period, and that is what the user told us they did. B&D need to capture the values as provided by a licensee.

Another example we previously made changes to support, is a licensee with a new meter. Their meter readings might be, 0, 0, 0, 0 etc. We have to be able to capture the entries, even if the result is a 0cm return.

What they want the system to stop, is allowing users to submit a return where _every line_ is blank.

So, the validation is if _at least_ one line has a value, the return can be submitted. Otherwise, we should block it.

Ping me if you need to discuss this further.

app/services/return-logs/setup/submit-check.service.js
Jozzey: New convention used

app/services/return-logs/setup/submit-check.service.js
Jozzey: New convention used

PR #2446, opened 2025-09-25
Add migration to add a template id column to notifications

jonathangoulding:

The Notify template ids where not previously stored.

This migration adds the column to record the Notify template id.

https://docs.notifications.service.gov.uk/node.html#templateid-required

The water abstraction service was updated here - https://github.com/DEFRA/water-abstraction-service/pull/2725


jonathangoulding:
Relies on https://github.com/DEFRA/water-abstraction-service/pull/2725

Cruikshanks (approved):

PR #2454, opened 2025-09-26
Add fetch notifications service

jonathangoulding:

We currently create notifications after sending them to Notify. This allowed us to save the notification and the Notify response in one query.

We are changing this approach after seeing some delays in showing a notification on the notice page.

We want to create the notification and mark them as 'pending', and then update the notification with the Notify response.

Once we save the notifications, we need to pull them out with their notification id. 

This change adds the fetch notifications service to support this work.

### DIFF
--- a/test/services/notices/setup/create-notifications.service.test.js
+++ b/test/services/notices/setup/create-notifications.service.test.js
@@ -26,27 +26,30 @@ describe('Notices - Setup - Create notification service', () => {
     })
 
     eventId = event.id
+
+    notifications = [
+      { eventId, createdAt: timestampForPostgres() },
+      { eventId, createdAt: timestampForPostgres() }
+    ]
   })
 
-  describe('when inserting a single notification', () => {
-    beforeEach(async () => {
-      notifications = [{ eventId, createdAt: timestampForPostgres() }]
+  it('should return the saved notifications', async () => {
+    const result = await CreateNotificationsService.go(notifications)
+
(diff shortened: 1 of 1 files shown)


app/services/notices/setup/fetch-notifications.service.js
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * The legacy code used a table called 'events' which groups all the notifications together. This is why we pass in the
+ * eventId as this relates to the 'event_id' column.
+ *
+ * @param {string} eventId - the unique id used to group all the notifications together
+ *
+ * @returns {object} - the notifications for the event
+ */
+async function go(eventId) {
+  return NotificationModel.query().where('event_id', eventId)
```
Cruikshanks, line 21: If using Objection.js you should be able to just use camelcase for properties.


```suggestion
  return NotificationModel.query().where('eventId', eventId)
```

However, my main request is that we specify the fields we want, even if it is all of them (though I'd be surprised if it was), at this time.

Therefore, we need to add `.select()` to the query.

Cruikshanks (changes requested):

app/services/notices/setup/fetch-notifications.service.js
jonathangoulding: fair point well made. 

I will look into what we need and update.

Cruikshanks (approved):

PR #2458, opened 2025-09-29
Update return forms to generate before sending

jonathangoulding:

We are seeing some performance issues before and after updating the notification flow to first save and then fetch the notifications. This was to mark all notification as pending so when the user sees the report page all the notifications are present and marked as pending.

The issue was generating all the PDF for the return forms journey. We here creating the notifications and the PDF before the user could continue to the confirmation page.

When we updated the logic to first save the notification and then fetch, it was simpler to save the PDF as part of that change, knowing we would be generating the PDF just before sending to Notify. So this was a temporary change.

This change implements generating the PDF just before sending to Notify.

This means we no longer generate and save the content before the batch process is started, but we save it as part of the Notify response update.

To achieve this as simply as possible, we have to save the PDF page data into the notification (most of this is already set in the personalisation, but some extra has been added). As this is now in the expected notification format, we need to map the notification data into PDF page data.

This is relevant to mention as we have updated the preview to format the page data through the return forms notifications presenter.

### DIFF
--- a/app/presenters/return-versions/setup/check/returns-requirements.presenter.js
+++ b/app/presenters/return-versions/setup/check/returns-requirements.presenter.js
@@ -34,14 +34,10 @@ function go(requirements, points, journey) {
 }
 
 function _abstractionPeriod(abstractionPeriod) {
-  const {
-    'abstraction-period-start-day': startDay,
-    'abstraction-period-start-month': startMonth,
-    'abstraction-period-end-day': endDay,
-    'abstraction-period-end-month': endMonth
-  } = abstractionPeriod
-  const startDate = formatAbstractionDate(startDay, startMonth)
-  const endDate = formatAbstractionDate(endDay, endMonth)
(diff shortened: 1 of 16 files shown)


app/services/notices/setup/determine-return-forms.service.js
```diff
@@ -16,7 +14,7 @@ const ReturnFormsNotificationPresenter = require('../../../presenters/notices/se
  * @param {object[]} recipients - List of recipient objects, each containing recipient details like email / address.
  * @param {string} eventId - the event id to link all the notifications to an event
  *
- * @returns {Promise<object[]>} - Resolves an array of return forms notifications
+ * @returns {object[]} - Resolves an array of return forms notifications
```
Cruikshanks, line 17: If this is no longer an async service then we should remove `async` from `go()`. Else this change needs to go back in.

Cruikshanks (changes requested):
Just one async thing to check.

PR #2465, opened 2025-10-01
Add returned at date to scheduled notifications view

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-5296

As part of the work to send out letters to licence holders we need a way to be able to know when a letter has been returned. GovNotify allows you to set up a callback url: https://docs.notifications.service.gov.uk/node.html#returned-letters

When this happens we want to be able to register when we were notified that the letter was returned so this PR adds a column that we will use to register when the most recent letter for a notification was returned.

### DIFF
--- a/db/migrations/legacy/20221108007033_water-scheduled-notification.js
+++ b/db/migrations/legacy/20221108007033_water-scheduled-notification.js
@@ -32,6 +32,7 @@ exports.up = function (knex) {
     table.jsonb('return_log_ids')
     table.uuid('template_id')
     table.binary('pdf')
+    table.date('returned_at')
 
     // Legacy timestamps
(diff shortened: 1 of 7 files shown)


Cruikshanks (changes requested):
Really sorry @robertparkinson 🤦 😬 

The public view `scheduled_notifications` is dead, along with the model for it.

When we properly got working on migrating the 'notice' journeys, we realised that 'events' and 'scheduled_notifications' are confusing names for these things.

@jonathangoulding has therefore done the work to bring in a `notifications` view and model, and all changes should be applied to it going forward (you'll see he has done a few recently).

Just as soon as I let up on ad-hoc stuff, Jon will be removing `scheduled_notifications` from this repo.

PR #2466, opened 2025-10-01
Add preview return form to the communications tab

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5295

The legacy code did not save a copy of the return form (paper form) sent to the user. We have updated our notification process to save PDF's sent to Notify, this includes the return form.

We only have return forms as PDF's currently.

This change updates the communications tab to link to the notification page when the notification is a return form.

The notification page has been updated to show a download preview link when the notification PDF has content.

A new service has been added to download the PDF file.

### DIFF
--- a/app/services/notifications/download-notification.service.js
+++ b/app/services/notifications/download-notification.service.js
@@ -1,0 +1,25 @@
+'use strict'
+
+/**
+ * Orchestrates fetching and presenting the data needed for the view notification page
+ * @module DownloadNotificationService
+ */
+
+const FetchDownloadNotificationService = require('./fetch-notification-download.service.js')
+
+/**
+ * Orchestrates fetching and presenting the data needed for the view notification page
+ *
+ * @param {string} notificationId - The UUID of the notification
+ *
+ * @returns {Promise<object>} an object representing the `pageData` needed by the view notification template.
+ */
+async function go(notificationId) {
+  const notificationData = await FetchDownloadNotificationService.go(notificationId)
+
+  return notificationData.pdf
+}
+
+module.exports = {
+  go
+}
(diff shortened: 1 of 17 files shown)


app/services/notifications/download-notification.service.js
```diff
@@ -0,0 +1,25 @@
+'use strict'
+
+/**
+ * Orchestrates fetching and presenting the data needed for the view notification page
+ * @module DownloadNotificationService
+ */
+
+const FetchDownloadNotificationService = require('./fetch-notification-download.service.js')
```
jonathangoulding, line 8: Needs a test

app/services/notifications/fetch-notification-download.service.js
```diff
@@ -0,0 +1,24 @@
+'use strict'
+
+/**
+ * Fetches the matching notification and licence data needed for the view
+ * @module FetchDownloadNotificationService
+ */
+
+const NotificationModel = require('../../models/notification.model.js')
+
+/**
```
jonathangoulding, line 10: Needs a test

test/services/notifications/download-notification.service.test.js
Cruikshanks, line 1: Rather than add a record, the preference is to stub the FetchService. That is the main driver for why we have the two separate rather than in one service: it allows us to stub the DB interaction in these kinds of service tests.

Cruikshanks (changes requested):
I believe we could be stubbing rather than creating an actual record in one of the tests.

Also, as we're touching this page, we should be 'boy scouting' it to the template standard. But that I leave to you about whether you do it as part of this PR, or in a follow-up.

PR #2471, opened 2025-10-02
Set returned at date on returned letters

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-5296

As part of the work to send out letters to licence holders we need a way to be able to know when a letter has been returned. GovNotify allows you to set up a callback url: https://docs.notifications.service.gov.uk/node.html#returned-letters

When this happens we want to be able to register when we were notified that the letter was returned so this PR sets the returned at date when we receive a request from GovNotify on the callback url.

### DIFF
--- a/app/services/notifications/submit-returned-letter.service.js
+++ b/app/services/notifications/submit-returned-letter.service.js
@@ -1,0 +1,24 @@
+'use strict'
+
+/**
+ * Orchestrates updating the returnedAt date for a letter
+ *
+ * @module SubmitReturnedLetterService
+ */
+
+const NotificationModel = require('../../models/notification.model.js')
+
+/**
+ * Orchestrates updating the returnedAt date for a letter
+ *
+ * @param {string} notifyId - The id of the notification
+ *
+ * @returns {Promise<object>} - The returned database object
+ */
+async function go(notifyId) {
+  await NotificationModel.query().patch({ returnedAt: new Date() }).where('notifyId', notifyId)
+}
+
+module.exports = {
+  go
+}
--- a/test/services/notifications/submit-returned-letter.service.test.js
+++ b/test/services/notifications/submit-returned-letter.service.test.js
@@ -1,0 +1,46 @@
+'use strict'
+
+// Test framework dependencies
+const Lab = require('@hapi/lab')
+const Code = require('@hapi/code')
+
+const { describe, it, beforeEach } = (exports.lab = Lab.script())
+const { expect } = Code
+
+const { generateUUID, timestampForPostgres, today } = require('../../../app/lib/general.lib.js')
+
+// Test helpers
+const EventHelper = require('../../support/helpers/event.helper.js')
+const NotificationHelper = require('../../support/helpers/notification.helper.js')
+const NotificationModel = require('../../../app/models/notification.model.js')
+
+// Thing under test
+const SubmitReturnedLetterService = require('../../../app/services/notifications/submit-returned-letter.service.js')
+
+describe('Submit Returned Letter Service', () => {
+  let eventId
+  let notifyId
+  const todaysDate = today()
+
+  beforeEach(async () => {
+    const event = await EventHelper.add({
+      type: 'notification',
+      subtype: 'returnsInvitation'
+    })
+
+    eventId = event.id
+    notifyId = generateUUID()
+
+    await NotificationHelper.add({ eventId, notifyId, createdAt: timestampForPostgres() })
+  })
+
+  describe('when called with a valid notifyId', () => {
+    it('updates the returnedAt date for the provided notification id', async () => {
+      await SubmitReturnedLetterService.go(notifyId)
+
+      const updatedResult = await NotificationModel.query().where('notifyId', notifyId)
+
+      expect(updatedResult[0].returnedAt).to.equal(todaysDate)
+    })
+  })
+})
--- a/app/controllers/notifications.controller.js
+++ b/app/controllers/notifications.controller.js
@@ -5,6 +5,7 @@
  * @module NotificationsController
  */
 
+const SubmitReturnedLetterService = require('../services/notifications/submit-returned-letter.service.js')
 const ViewNotificationService = require('../services/notifications/view-notification.service.js')
 
 const NO_CONTENT_STATUS_CODE = 204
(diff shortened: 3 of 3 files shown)


test/services/notifications/submit-returned-letter.service.test.js
```diff
@@ -0,0 +1,46 @@
...
+const NotificationHelper = require('../../support/helpers/notification.helper.js')
+const NotificationModel = require('../../../app/models/notification.model.js')
+
+// Thing under test
+const SubmitReturnedLetterService = require('../../../app/services/notifications/submit-returned-letter.service.js')
+
+describe('Submit Returned Letter Service', () => {
+  let eventId
+  let notifyId
+  const todaysDate = today()
```
Jozzey, line 23: Our convention is to put consts above lets. Generally we leave a gap between them as well

```suggestion
  const todaysDate = today()

  let eventId
  let notifyId
```

app/services/notifications/submit-returned-letter.service.js
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Orchestrates updating the returnedAt date for a letter
+ *
+ * @param {string} notifyId - The id of the notification
+ *
+ * @returns {Promise<object>} - The returned database object
+ */
+async function go(notifyId) {
+  await NotificationModel.query().patch({ returnedAt: new Date() }).where('notifyId', notifyId)
```
Cruikshanks, line 19: We need to know if we receive a call back but cannot find the matching letter. We just need to log it, but it means we need to wrap this in something that checks we did update a record.

I'd suggest using `.returning('id')` (see [returning tricks](https://vincit.github.io/objection.js/recipes/returning-tricks.html#postgresql-returning-tricks)) and then checking if you actually have an ID.

If not, log (`omg` vs `omfg` - no need for an error) the fact we failed to find a matching notification for the callback.

Cruikshanks (changes requested):
Some feedback and request for changes. I've also pinged a question via DM.

Shout if any questions or issues.

app/services/notifications/process-returned-letter.service.js
```diff
@@ -0,0 +1,36 @@
...
+ * Processes a returned letter callback from GOV.UK Notify
+ *
+ * When a letter sent by Notify is 'returned to sender', they can let us know via a callback request.
+ *
+ * The request from Notify will contain the Notify ID of the notification. We can use that to identify the relevant
+ * notification that has been returned, and update it accordingly.
+ *
+ * @param {string} notifyId - Notify's ID for the notification
+ *
+ * @returns {Promise<object>} - The returned database object
```
Cruikshanks, line 21: We can drop this because we don't return anything.


```suggestion
```

app/services/notifications/process-returned-letter.service.js
```diff
@@ -0,0 +1,36 @@
...
+ */
+async function go(notifyId) {
+  const notification = await NotificationModel.query()
+    .patch({ returnedAt: new Date(), status: 'returned' })
+    .where('notifyId', notifyId)
+    .returning('id')
+
+  if (notification.length === 0) {
+    global.GlobalNotifier.omg('No matching notice found for returned letter request', { notifyId })
```
Cruikshanks, line 30: It's a notification, rather than a notice, that we're updating.

```suggestion
    global.GlobalNotifier.omg('No matching notification found for returned letter request', { notifyId })
```

app/services/notifications/process-returned-letter.service.js
```diff
@@ -0,0 +1,36 @@
...
+ * The request from Notify will contain the Notify ID of the notification. We can use that to identify the relevant
+ * notification that has been returned, and update it accordingly.
+ *
+ * @param {string} notifyId - Notify's ID for the notification
+ *
+ * @returns {Promise<object>} - The returned database object
+ */
+async function go(notifyId) {
+  const notification = await NotificationModel.query()
+    .patch({ returnedAt: new Date(), status: 'returned' })
```
Cruikshanks, line 25: We should use the specific helper we have for time stamps in PostgreSQL.


```suggestion
    .patch({ returnedAt: timestampForPostgres(), status: 'returned' })
```

test/services/notifications/process-returned-letter.service.test.js
```diff
@@ -0,0 +1,47 @@
...
+
+// Test helpers
+const EventHelper = require('../../support/helpers/event.helper.js')
+const NotificationHelper = require('../../support/helpers/notification.helper.js')
+const NotificationModel = require('../../../app/models/notification.model.js')
+
+// Thing under test
+const ProcessReturnedLetterService = require('../../../app/services/notifications/process-returned-letter.service.js')
+
+describe('Submit Returned Letter Service', () => {
```
Cruikshanks, line 20: Named after previous service name.


```suggestion
describe('Notifications - Process Returned Letter service', () => {
```

test/services/notifications/process-returned-letter.service.test.js
```diff
@@ -0,0 +1,47 @@
...
+
+  describe('when called with a valid notifyId', () => {
+    it('updates the returnedAt date for the provided notification id', async () => {
+      await ProcessReturnedLetterService.go(notifyId)
+
+      const updatedResult = await NotificationModel.query().where('notifyId', notifyId)
+
+      expect(updatedResult[0].returnedAt).to.equal(todaysDate)
+    })
+  })
```
Cruikshanks, line 46: We could do with a test that covers what happens when called with an unknown Notify ID.

Cruikshanks (changes requested):
Just some NITS, and we could do with a test that covers what we do should we receive a callback for an unknown Notify ID.

test/services/notifications/process-returned-letter.service.test.js
```diff
@@ -0,0 +1,65 @@
...
+
+      const updatedResult = await NotificationModel.query().where('notifyId', notifyId)
+
+      expect(updatedResult[0].returnedAt).to.equal(todaysDate)
+      expect(notifierStub.omg.called).to.be.false()
+    })
+  })
+
+  describe('when called with a valid notifyId', () => {
+    it('updates the returnedAt date for the provided notification id', async () => {
```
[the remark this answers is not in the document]
robertparkinson: 🤦‍♂️

Jozzey (commented):
👍🏼 I'm happy when @Cruikshanks is happy 😁

Cruikshanks (approved):

PR #2476, opened 2025-10-03
Remove any reference to scheduled notifications

jonathangoulding:
We have created a new table 'water.notification' to hold all the notifications.

This is in use in some places but not all.

This change removes any reference to 'scheduled notifications' and replaces it with 'notifications'.

The view will be dropped.


jonathangoulding:
> This is great, but I was expecting to see a migration to drop the old scheduled_notifications view as part of the final cleanup.
> 
> Shout if that's an issue, or if it doesn't make sense.

I was going to do this in another change?

PR #2477, opened 2025-10-03
Migrate search to system

ben-sagar:
https://eaflood.atlassian.net/browse/WATER-5304

We are migrating the internal search page (the root page of the internal service) from the legacy apps to our new `water-abstraction-system app`.

The existing page allows users to search using different types of values, for example, licence references, licence names, and return references.

We are implementing these different types of search into the new `system` page one at a time to manage the complexity.

This change is to implement searching for licences.

The display of the search results is currently very basic - the actual display format is being worked on by our service designer and will be adjusted to suit whatever design they come up with in future changes.

I haven't added a feature flag for this feature yet as we can't really use it as a replacement for the existing search until we've got other things we can search on. For the moment, this search page can just be accessed at `/system/search`.


Cruikshanks (approved):
As discussed, this is a large PR I'm sure to have feedback on 😬 . However, we're trying to get something else over the line which is taking all my time so I don't have the resource to give it the attention it needs.

Rather than blocking you, I'm just going to go ahead and approve.

The page is essentially hidden, and even if you did find the URL, you couldn't accidentally or maliciously harm the service! (or I hope not!)

I'll look at where we are at when I should have more time next week.

PR #2484, opened 2025-10-06
Update notice type journey entry point

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5308

We are updating the 'standard' journey to allow the user to select the notice type. This is replacing the existing two links on the manage page.

This change adds two buttons to the notice page to allow the user to select a standard or adhoc journey.

Navigation has been updated to make sure the 'flow' between pages is as expected.

We no longer provide the notice type to the initiate session, as the user now selects this on select a notice page. (Abstraction alerts do still need this set, but we base our selection on the journey.)

The previous links on the legacy manage page will be hidden behind a feature flag, and this will be the only entry into the notice journey.


jonathangoulding:
Updated to ticket expectations

Cruikshanks (changes requested):
Just a couple of NITs.

However, I've completely overlooked permissions! 🤦 

If a user has the `bulk_return_notifications` scope then we display the **Create a standard notice** button.

If a user has the `returns` scope then we display the **Create an ad-hoc notice** button.

If they have neither we don't display the buttons or the text above them.

The Jira ticket has been updated to reflect this.

If it helps, if a user has `bulk_return_notifications` they'll also have `returns`.

PR #2490, opened 2025-10-08
Refactor 'return-forms' to 'paper-return' page

jonathangoulding:
We have implemented a 'return-forms' journey for notices on the adhoc journey. This needs to be updated to 'paper-forms'.

This change updates the the 'return-forms' page and the navigation to and from this page.

This is part of a series of changes to refactor any use of 'return-forms' to 'paper-return'


app/controllers/notices-setup.controller.js
```diff
@@ -43,9 +43,9 @@ const SubmitLicenceService = require('../services/notices/setup/submit-licence.s
 const SubmitNoticeTypeService = require('../services/notices/setup/submit-notice-type.service.js')
 const SubmitRecipientNameService = require('../services/notices/setup/submit-recipient-name.service.js')
 const SubmitRemoveLicencesService = require('../services/notices/setup/submit-remove-licences.service.js')
-const SubmitReturnFormsService = require('../services/notices/setup/submit-return-forms.service.js')
 const SubmitReturnsPeriodService = require('../services/notices/setup/returns-period/submit-returns-period.service.js')
 const SubmitSelectRecipientsService = require('../services/notices/setup/submit-select-recipients.service.js')
+const submitPaperReturnService = require('../services/notices/setup/submit-paper-return.service.js')
```
Cruikshanks, line 48: NIT. I also think it needs to be bumped up in the order now.


```suggestion
const SubmitPaperReturnService = require('../services/notices/setup/submit-paper-return.service.js')
```

Cruikshanks (changes requested):
Just the one NIT!

Cruikshanks (approved):

PR #2493, opened 2025-10-08
Refactor the Notice journey to use an enum for the 'journey' and 'noticeType'

jonathangoulding:
We have littered the code with 'hard coded' journey and notice type values, these are fixed values. 

We can have these shared by a lib service, we are adding the journey and notice type as enums. At least the closest we can get this JS using `Obejct.freeze`.

We are changing the 'returnForms' to 'paperReturn', this has highlighted the complexity / impact of making this change in multiple places.

We have implemented all the legacy journeys, and we are now 'cleaning' up.


Cruikshanks (changes requested):
Marked as request changes but as I said, come back to me if I have missed something.

Cruikshanks (approved):

PR #2494, opened 2025-10-08
Refactor 'return-forms' to 'paper-return' notice type page

jonathangoulding:
We have implemented a 'return-forms' journey for notices on the adhoc journey. This needs to be updated to 'paper-forms'.

This change updates the notice type page.

This is part of a series of changes to refactor any use of 'return-forms' to 'paper-return'. 

Updated to template standard.

### DIFF
--- a/app/presenters/notices/setup/notice-type.presenter.js
+++ b/app/presenters/notices/setup/notice-type.presenter.js
@@ -5,6 +5,8 @@
  * @module NoticeTypePresenter
  */
 
+const { NoticeType, NoticeJourney } = require('../../../lib/static-lookups.lib.js')
+
 /**
  * Formats data for the `/notices/setup/{sessionId}/notice-type` page
  *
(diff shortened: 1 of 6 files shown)


app/presenters/notices/setup/notice-type.presenter.js
```diff
@@ -23,43 +25,52 @@ function go(session) {
...
+      value: NoticeType.INVITATIONS,
       text: 'Standard returns invitation'
     },
     {
-      checked: noticeType === 'returnForms',
-      value: 'returnForms',
-      text: 'Submit using a paper form invitation'
+      checked: noticeType === NoticeType.PAPER_RETURN,
+      value: NoticeType.PAPER_RETURN,
+      text: 'Submit using a paper return invitation'
```
Cruikshanks, line 73: Just needs to be 'Paper return'.

And although not part of this, 'Standard returns invitation' needs to be updated to 'Returns invitation'.

I'm happy for that change to be included as part of this.

Cruikshanks (changes requested):

PR #3144, opened 2026-03-20
Edit internal user

ben-sagar:
https://eaflood.atlassian.net/browse/WATER-5514

We have implemented pages for viewing users and now are adding the corresponding pages for editing users.

This change is to implement the page for editing an internal user's permissions.

There are some complications in that only a super user can apply or remove super user permissions, but otherwise it's just a case of a simple edit page to select the required permission level.

The main issue is how the permissions are stored in the database - as a set of roles and groups that are linked to the user record by its old numeric ID rather than the new UUID.

Depending on what happens about all this permissions stuff in the future, we may amend the way this is stored altogether.

### DIFF
--- a/app/controllers/users.controller.js
+++ b/app/controllers/users.controller.js
@@ -5,10 +5,14 @@
  * @module UsersController
  */
 
+const Boom = require('@hapi/boom')
+
 const FetchLegacyIdService = require('../services/users/fetch-legacy-id.service.js')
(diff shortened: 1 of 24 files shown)


Jozzey:
This PR has been superseeded by the work done in PR https://github.com/DEFRA/water-abstraction-system/pull/3412 and can now be closed.

PR #3226, opened 2026-04-10
Update submit cancels to use 'DeleteSessionDal'

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5573

We recently refactored how we use fetch and delete session data, we now have a single way of doing each.

We are moving away from using the database in service test.

This change updates the remaining two places where we delete the session using the session model, they have been updated to use the 'DeleteSessionDal'.

The tests have been updated to stub the 'DeleteSessionDal' and check it has been called. We understand this can be seen as a futile test, but we do this to maintain consistency with the rest of the test suite, where we have more complicated logic surrounding deleting the session.

### DIFF
--- a/test/services/return-logs/setup/submit-cancel.service.test.js
+++ b/test/services/return-logs/setup/submit-cancel.service.test.js
@@ -23,7 +23,7 @@ describe('Return Logs Setup - Submit Cancel service', () => {
   beforeEach(async () => {
     session = SessionModelStub.build(Sinon, {})
 
-    Sinon.stub(DeleteSessionDal, 'go').resolves(session)
+    Sinon.stub(DeleteSessionDal, 'go').resolves()
   })
 
   afterEach(() => {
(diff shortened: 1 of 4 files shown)


test/services/return-logs/setup/submit-cancel.service.test.js
```diff
@@ -17,31 +21,20 @@ describe('Return Logs Setup - Submit Cancel service', () => {
...
-        purposes: 'Evaporative Cooling',
-        receivedDate: '2025-01-31T00:00:00.000Z',
-        returnLogId: '610ac214-03c2-4b25-a89a-65bb81d1624f',
-        returnReference: '1234',
-        siteDescription: 'POINT A, TEST SITE DESCRIPTION',
-        startDate: '2004-04-01T00:00:00.000Z',
-        twoPartTariff: false
-      }
-    })
+    session = SessionModelStub.build(Sinon, {})
```
jonathangoulding, line 24: Not comment

Cruikshanks (approved):

PR #3234, opened 2026-04-13
Update tests to stub 'FetchSessionDal'

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5573

We recently refactored how we use fetch and delete session data, we now have a single way of doing each.

We are moving away from using the database in service test.

This change updates some of the tests still using the session helper to stub the 'FetchSessionDal'

### DIFF
--- a/test/services/return-versions/setup/submit-remove.service.test.js
+++ b/test/services/return-versions/setup/submit-remove.service.test.js
@@ -24,7 +24,7 @@ describe('Return Versions Setup - Submit Remove service', () => {
   let sessionData
   let yarStub
 
-  beforeEach(async () => {
+  beforeEach(() => {
     sessionData = {
       checkPageVisited: false,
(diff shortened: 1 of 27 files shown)


test/services/return-versions/setup/submit-remove.service.test.js
```diff
@@ -19,64 +21,69 @@ describe('Return Versions Setup - Submit Remove service', () => {
   const requirementIndex = 0
 
   let session
+  let sessionData
   let yarStub
 
   beforeEach(async () => {
```
Cruikshanks: We missed an `async`.


```suggestion
  beforeEach(() => {
```

Cruikshanks (changes requested):
Just one missed `async` to be dropped.

Cruikshanks (approved):

PR #3254, opened 2026-04-20
Add the process renewal invitations service job skeleton

jonathangoulding:

This change adds the process renewal invitations service job skeleton. It adds the route, controller and basic tests.


app/controllers/jobs.controller.js
```diff
@@ -56,6 +57,12 @@ async function notificationStatus(_request, h) {
   return h.response().code(HTTP_STATUS_NO_CONTENT)
 }
 
+async function renewalReminders(request, h) {
```
jonathangoulding:
```suggestion
async function renewalReminders(_request, h) {
```

Cruikshanks (changes requested):
Sorry, should have emphasised it more. Thought I'd catch you before you started.

Think of this job as the 'renewal invitations' job, which means some things are going to need a rename in your PR.

Why? The manual process already consists of two steps: an 'invitation to renew' followed by a 'reminder to renew'. We've agreed to just do the first part for now, but we're 99% certain we'll be asked to send the second automatically as well. So, if we use 'reminder' now, we'll have to come up with _another_ name for the second job.

Also, 'invitation' followed by 'reminder' matches exactly what we do with returns, so it would be consistent with the service's existing terminology.

If you want to grab me to confirm what things should be named, give me a ping.

Cruikshanks (approved):

PR #3261, opened 2026-04-21
Add an email alert when jobs fail

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-5600

We are about to create a job that scans the db and sends an email/letter if certain conditions are met. However we have no way in the service to know if this job ran successfully or not as it could be a day when no one is notified. We'd have to run db queries which is burdensome but we still need to know if the job failed so we can make sure people are contacted. 

This PR adds an email alert to the team as part of the notifier lib. We then call this in the catches for the jobs.

### DIFF
--- a/app/lib/base-notifier.lib.js
+++ b/app/lib/base-notifier.lib.js
@@ -9,6 +9,9 @@ const { Notifier } = require('@airbrake/node')
 const Pino = require('pino')
 
 const AirbrakeConfig = require('../../config/airbrake.config.js')
+const CreateEmailRequest = require('../requests/notify/create-email.request.js')
+const NotifyConfig = require('../../config/notify.config.js')
+const { NOTIFY_TEMPLATES } = require('./notify-templates.lib.js')
 
 /**
  * Based class for combined logging and Airbrake (Errbit) notification managers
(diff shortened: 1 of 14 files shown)


test/services/jobs/customer-files/process-customer-files.service.test.js
Cruikshanks, line 1: Picking on this one as the first of the job tests.

When it comes to the `describe('when X fails', () => {`, the `it` wasn't ideally worded.

When we word the `describe` and `it` blocks, we're aiming for a 'readable' sentence.

The previous version did that, but didn't indicate what we were testing. 

Unfortunately, the updated wording fails the 'readablity' metric: "it imfg handles the error".

So, can I suggest the following (and this applies to all the job tests).

```javascript
    it('records the error by calling "omfg()"', async () => {
      await ProcessCustomerFilesService.go()

      const args = notifierStub.omfg.firstCall.args

      expect(args[0]).to.equal('Customer files job failed')
      expect(args[1]).to.be.null()
      expect(args[2]).to.be.an.error()
    })

   it('notifies the team by calling "redAlert()"', async () => {
      await ProcessCustomerFilesService.go()

      const args = notifierStub.redAlert.firstCall.args

      expect(args[0]).to.equal('Customer files job failed')
    })

   it('does not throw an error', async () => {
      await expect(ProcessCustomerFilesService.go()).not.to.reject()
   })
```

app/lib/base-notifier.lib.js
```diff
@@ -109,6 +112,33 @@ class BaseNotifierLib {
...
+    const options = {
+      personalisation: {
+        content: error ? content + ` with: ${error}` : content
+      }
+    }
+
+    const emails = NotifyConfig.alertEmailAddresses.split(',')
+
+    for (const email of emails) {
+      CreateEmailRequest.send(NOTIFY_TEMPLATES.system.statusAlert, email, options)
```
Cruikshanks: We're not awaiting this async call. We also haven't declared `redAlert()` as async, and we're not awaiting the `redAlert()` calls in the jobs themselves.

Would you mind running an experiment for me?

With the logs visible, could you alter `CreateEmailRequest.send()` to purposefully throw an error?

I'm interested in whether our Hapi error plugin would catch the unhandled exception or cause Node to terminate due to an unhandled promise rejection.

Cruikshanks (changes requested):
I'm being picky about the wording in the unit tests, so some changes are needed there.

More importantly, I think we may have exposed ourselves to an unhandled promise rejection, hence I have an experiment for you!

Cruikshanks (approved):

PR #3269, opened 2026-04-23
Create a notice for the renewal invitations job

jonathangoulding:

When sending a renewal invitations, we need to create a notice and send the notifications.

This change creates a notice, this will be an invitation notice type.

We also want to display the 'count' in the log we display when the job completes, this has also been added.

Sending notifications will be implemented in later changes.

### DIFF
--- a/app/lib/static-lookups.lib.js
+++ b/app/lib/static-lookups.lib.js
@@ -324,7 +324,8 @@ const NoticeType = Object.freeze({
   ALTERNATE_INVITATION: 'alternateInvitations',
   INVITATIONS: 'invitations',
   PAPER_RETURN: 'paperReturn',
-  REMINDERS: 'reminders'
+  REMINDERS: 'reminders',
+  RENEWAL_INVITATIONS: 'renewalInvitations'
 })
 
 /**
@@ -355,6 +356,12 @@ const NoticeTypes = Object.freeze({
     prefix: 'RREM-',
     subType: 'returnReminder',
     notificationType: 'Returns reminder'
+  },
+  [NoticeType.RENEWAL_INVITATIONS]: {
+    name: 'Invitations: renewal',
+    prefix: 'REIN-',
+    subType: 'renewalInvitations',
+    notificationType: 'Renewal invitations'
   }
 })
 
--- a/app/lib/static-lookups.lib.js
+++ b/app/lib/static-lookups.lib.js
@@ -358,7 +358,7 @@ const NoticeTypes = Object.freeze({
     notificationType: 'Returns reminder'
   },
   [NoticeType.RENEWAL_INVITATIONS]: {
-    name: 'Invitations: renewal',
+    name: 'Renewals: invitation',
     prefix: 'REIN-',
     subType: 'renewalInvitations',
     notificationType: 'Renewal invitations'
--- a/app/services/jobs/renewal-invitations/process-renewal-invitations.service.js
+++ b/app/services/jobs/renewal-invitations/process-renewal-invitations.service.js
@@ -17,9 +17,9 @@ async function go(days) {
   try {
     const startTime = currentTimeInNanoseconds()
 
-    const invitations = await SendRenewalInvitations.go(days)
+    const recipients = await SendRenewalInvitations.go(days)
 
-    calculateAndLogTimeTaken(startTime, 'Renewal invitations status job complete', { count: invitations.length })
+    calculateAndLogTimeTaken(startTime, 'Renewal invitations status job complete', { count: recipients.length })
   } catch (error) {
     global.GlobalNotifier.omfg('Notification status job failed', null, error)
   }
--- a/app/services/jobs/renewal-invitations/process-renewal-invitations.service.js
+++ b/app/services/jobs/renewal-invitations/process-renewal-invitations.service.js
@@ -17,9 +17,9 @@ async function go(days) {
   try {
     const startTime = currentTimeInNanoseconds()
 
-    await SendRenewalInvitations.go(days)
+    const invitations = await SendRenewalInvitations.go(days)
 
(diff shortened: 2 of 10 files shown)


test/services/jobs/renewal-invitations/process-renewal-invitations.service.test.js
```diff
@@ -20,7 +20,7 @@ describe('Jobs - Renewal Invitations - Process Renewal Invitations service', ()
   let notifierStub
 
   beforeEach(() => {
-    Sinon.stub(SendRenewalInvitations, 'go').resolves([])
+    Sinon.stub(SendRenewalInvitations, 'go').resolves(['mock invitation'])
```
jonathangoulding, line 23: Not too happy with this, but we wanted to count but not provide a complex object.

app/services/jobs/renewal-invitations/send-renewal-invitations.service.js
```diff
@@ -37,6 +42,18 @@ function _expiryDate(futureExpiredDate) {
   return targetDate
 }
 
+async function _notice(recipients) {
+  const { name, prefix, subType } = NoticeTypes[NoticeType.INVITATIONS]
+
+  const noticeType = {
```
Cruikshanks: Forgive, but when a read "type", I think of a single value, not an object with multiple values.

Can I suggest


```suggestion
  const noticeData = {
```

app/services/notices/setup/create-notice.service.js
```diff
@@ -15,7 +15,7 @@ const { timestampForPostgres } = require('../../../lib/general.lib.js')
  * > Notices are event records with a type as `notification`. In the future we intend to move them to their own
  * > `water.notices` table. But for now this explains why the `EventModel` suddenly makes an appearance!
  *
- * @param {SessionModel} session - The session instance
+ * @param {object} session - The session instance, or a session-like object with the expected notice data
```
Cruikshanks: If the thing we are passing isn't a session, let's drop the name to avoid confusion in the future. Can I suggest `noticeData` again? 😁

app/services/jobs/renewal-invitations/process-renewal-invitations.service.js
```diff
@@ -17,9 +17,9 @@ async function go(days) {
   try {
     const startTime = currentTimeInNanoseconds()
 
-    await SendRenewalInvitations.go(days)
+    const invitations = await SendRenewalInvitations.go(days)
```
Cruikshanks: Is this an intent, i.e. `SendRenewalInvitations` will eventually return the invitations sent?

Otherwise, as it is actually returning recipients, should we rename it `recipients'?

test/services/jobs/renewal-invitations/send-renewal-invitations.service.test.js
```diff
@@ -8,21 +8,27 @@ const Sinon = require('sinon')
...
+  const recipients = [{ licenceRefs: generateLicenceRef() }]
 
   let clock
   let expiredDate
   let todayDate
 
   beforeEach(() => {
-    Sinon.stub(FetchRenewalRecipients, 'go').resolves([])
+    Sinon.stub(FetchRenewalRecipients, 'go').resolves(recipients)
+    Sinon.stub(CreateNoticeService, 'go').resolves()
```
Cruikshanks: If we are going to refer to the stubbed service later, we normally assign it to a variable.


```suggestion
    createNoticeStub = Sinon.stub(CreateNoticeService, 'go').resolves()
```

test/services/jobs/renewal-invitations/send-renewal-invitations.service.test.js
```diff
@@ -41,7 +47,25 @@ describe('Jobs - Renewal Invitations - Send Renewal Invitations service', () =>
...
+      const config = args[0]
+      expect(config.name).to.equal('Returns: invitation')
+      expect(config.subType).to.equal('returnInvitation')
+      expect(config.referenceCode).to.startWith('RINV-')
+
+      // Argument 2: The Recipients List
+      expect(args[1]).to.equal(recipients)
+
+      // Argument 3: The issuer email
+      expect(args[2]).to.equal('water_abstractiondigital@environment-agency.gov.uk')
```
Cruikshanks: Does this still work?


```suggestion
      // Argument 1: Notice type
      expect(createNoticeStub.firstCall.args[0]).to.equal({
        name: 'Returns: invitation',
        referenceCode: 'RINV-',
        subType: 'returnInvitation'
      })

      // Argument 2: The Recipients List
      expect(createNoticeStub.firstCall.args[1]).to.equal(recipients)

      // Argument 3: The issuer email
      expect(createNoticeStub.firstCall.args[2]).to.equal('water_abstractiondigital@environment-agency.gov.uk')
```

Cruikshanks (changes requested):
I've assumed we're using the returns invitation notice type as a stand-in until we add a new one for renewal invitations.

If that's the case, just some NITs and suggestions.

app/services/jobs/renewal-invitations/process-renewal-invitations.service.js
jonathangoulding: i think for some reason i was thinking it should match the serivce name. will change

app/lib/static-lookups.lib.js
```diff
@@ -355,6 +356,12 @@ const NoticeTypes = Object.freeze({
     prefix: 'RREM-',
     subType: 'returnReminder',
     notificationType: 'Returns reminder'
+  },
+  [NoticeType.RENEWAL_INVITATIONS]: {
+    name: 'Invitations: renewal',
```
Cruikshanks:
```suggestion
    name: 'Renewals: invitation',
```

Cruikshanks (changes requested):
One last NIT.

Cruikshanks (approved):

PR #3270, opened 2026-04-24
Add renewal invitations to notice page

jonathangoulding:

Renewal invitations are a new type of notice. A renewal invitation is added automatically by a job we have recently implemented.

This change adds the renewal invitations to the notice page.


app/lib/static-lookups.lib.js
```diff
@@ -316,7 +316,8 @@ const noticeMappings = {
   renewal: 'Renewal',
   returnInvitation: 'Returns invitation',
   returnReminder: 'Returns reminder',
-  waterAbstractionAlerts: 'alert'
+  waterAbstractionAlerts: 'alert',
+  renewalInvitations: 'Renewal invitations'
```
jonathangoulding: I want to check if this should be plural ? 

The others are not.

jonathangoulding:
Do we need to feature toggle this ? 

The filter will be shown if we don't. Unless we sit on the PR until the other job has run. 

If the filter is selected it will just show no notices found. 

The job might not pick up any renewal reminders for days when deployed.

Cruikshanks:
I've removed myself as a reviewer. Put me back on when this comes out of draft @jonathangoulding

Cruikshanks (approved):

PR #3278, opened 2026-04-27
Show impacted licences on check page

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-5466

As part of our ongoing work to decommission our legacy UI we identify journeys that are small enough to be ported over in discreet chunks of work. This is part of the change billing address journey.

This PR adds a list of impacted licences to the bottom of the check page.


app/services/billing-accounts/setup/view-check.service.js
```diff
@@ -48,13 +59,17 @@ async function _fetchExistingAddress(session) {
...
+    .where('billingAccountId', session.billingAccount.id)
+    .where('status', 'current')
 
-  const companyContacts = await FetchCompanyContactsService.go(companyId)
+  const impactedLicences = licenceRefs.map((licence) => {
+    return licence.licenceRef
+  })
 
-  return companyContacts
+  return impactedLicences
```
Cruikshanks: This should be moved to `app/dal/billing-accounts/fetch-impacted-licences.dal.js`

Cruikshanks (commented):
As part of our new ways of working, queries like this, rather than being run directly from the service, are instead moved to a DAL module, which is then called from the service.

If this is a problem, let me know.

app/presenters/billing-accounts/setup/check.presenter.js
```diff
@@ -12,10 +12,11 @@
  * @param {object} companyContacts - The company and its contacts
  * @param {object} existingAddress - The existing address of company
  * @param {object} companysHouseResult - The companys house details
+ * @param {Array<string>} impactedLicences - The list of impacted licence references
```
Cruikshanks:
```suggestion
 * @param {string[]} impactedLicences - The list of impacted licence references
```

app/presenters/billing-accounts/setup/check.presenter.js
```diff
@@ -12,10 +12,11 @@
...
  *
  * @returns {object} The data formatted for the view template
  */
-function go(session, companyContacts, existingAddress, companysHouseResult) {
+function go(session, companyContacts, existingAddress, companysHouseResult, impactedLicences = []) {
```
Cruikshanks: The result of `FetchImpactedLicences` is always being passed in, which means we don't have a situation where `impactedLicences` is not set as an arg.

```suggestion
function go(session, companyContacts, existingAddress, companysHouseResult, impactedLicences) {
```

app/dal/billing-accounts/fetch-impacted-licences.dal.js
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Fetches licence references for the provided billing account id.
+ *
+ * When changing the addess for a billing account we show the user a list of the licences impacted by the change. This
+ * services returns all the linked licence references for the provided billing account id.
+ *
+ * @param {string} billingAccountId - The UUID of the billing account.
+ *
+ * @returns {Promise<object[]>} An array of charge version objects with their licenceRefs.
```
Cruikshanks:
```suggestion
 * @returns {Promise<string[]>} The `licenceRef` of the impacted licences.
```

app/dal/billing-accounts/fetch-impacted-licences.dal.js
```diff
@@ -0,0 +1,35 @@
...
+  const licenceRefs = await ChargeVersionModel.query()
+    .select(['licenceRef'])
+    .where('billingAccountId', billingAccountId)
+    .where('status', 'current')
+
+  const impactedLicences = licenceRefs.map((licence) => {
+    return licence.licenceRef
+  })
+
+  return impactedLicences
```
Cruikshanks:
```suggestion
  const chargeVersions = await ChargeVersionModel.query()
    .select(['licenceRef'])
    .where('billingAccountId', billingAccountId)
    .where('status', 'current')

  const impactedLicences = chargeVersions.map((chargeVersion) => {
    return chargeVersion.licenceRef
  })

  return impactedLicences
```

app/dal/billing-accounts/fetch-impacted-licences.dal.js
```diff
@@ -0,0 +1,35 @@
...
+ * Fetches licence references for the provided billing account id.
+ *
+ * When changing the addess for a billing account we show the user a list of the licences impacted by the change. This
+ * services returns all the linked licence references for the provided billing account id.
+ *
+ * @param {string} billingAccountId - The UUID of the billing account.
+ *
+ * @returns {Promise<object[]>} An array of charge version objects with their licenceRefs.
+ */
+async function go(billingAccountId) {
```
Cruikshanks, line 20: As a billing account, I could be linked to multiple charge versions for the same licence. They may have changed for reasons other than a change to the billing account.

This would mean you could get the same licence ref multiple times in the results.

We're not handling this here nor in the presenter, which means you will get a licence ref duplicated on the page.

I'd suggest amending the query to include `.distinct()` to resolve this as simply as possible.

But I'd also ensure the unit tests cover this scenario as well.

test/dal/billing-accounts/fetch-existing-address.dal.test.js
```diff
@@ -0,0 +1,52 @@
...
+  before(async () => {
+    address = await AddressHelper.add()
+  })
+
+  after(async () => {
+    await address.$query().delete()
+  })
+
+  describe('when there is a matching address found', () => {
+    it('returns an array of address lines', async () => {
```
Cruikshanks:
```suggestion
    it('returns the matching address', async () => {
```

test/dal/billing-accounts/fetch-existing-address.dal.test.js
```diff
@@ -0,0 +1,52 @@
...
+        address5: null,
+        address6: null,
+        id: address.id,
+        postcode: address.postcode
+      })
+    })
+  })
+
+  describe('when there are no matching addresses found', () => {
+    it('returns an empty array', async () => {
```
Cruikshanks:
```suggestion
    it('returns "undefined"', async () => {
```

Cruikshanks (changes requested):
Shout if something doesn't make sense, or I've got the wrong end of the stick.

app/dal/billing-accounts/fetch-impacted-licences.dal.js
Cruikshanks, line 1: Sorry, also just spotted that we have an `ORDER BY` on the query. This means we could get inconsistent results every time we run it, which is a pain in the 🍑 for unit and acceptance tests.

test/dal/billing-accounts/fetch-impacted-licences.dal.test.js
Cruikshanks, line 1: I spotted a typo I'd missed first time around: `fristChargeVersion`. 🤦 

So I figured I might as well suggest a pattern I use when I have multiple records of the same type: just use an array.

Also, the benefit of DAL modules is that we can be more specific when checking the results. I figured this was a case we could assert `equals` directly on the result.

But to do that, we need to ensure the result is deterministic, which is when I realised we were not ordering the results.

So, assuming the `.orderBy` has been applied. This is my take on the unit tests.

```javascript
'use strict'

const { describe, it, after, before } = (exports.lab = Lab.script())
const { expect } = Code

// Test helpers
const ChargeVersionHelper = require('../../support/helpers/charge-version.helper.js')
const { generateUUID } = require('../../../app/lib/general.lib.js')
const { generateLicenceRef } = require('../../support/helpers/licence.helper.js')

// Thing under test
const FetchImpactedLicencesDal = require('../../../app/dal/billing-accounts/fetch-impacted-licences.dal.js')

describe('DAL - Fetch Impacted Licences dal', () => {
  const billingAccountId = generateUUID()
  const multiUseLicenceRef = generateLicenceRef()
  const singleUseLicenceRef = generateLicenceRef()

  let chargeVersions
  let multiUseLicenceRef
  let singleUseLicenceRef

  before(async () => {
    // NOTE: We want to confirm the results are sorted. So, of the two references generated, we'll use the 'higher'
    // one to create our first record, to confirm the order isn't a fluke of the order in which the records were created
    const licenceRefs = [generateLicenceRef(), generateLicenceRef()].sort()

    let multiUseLicenceRef = licenceRefs[0]
    let singleUseLicenceRef = licenceRefs[1]

    chargeVersions = [
      await ChargeVersionHelper.add({ billingAccountId, licenceRef: singleUseLicenceRef }),
      await ChargeVersionHelper.add({ billingAccountId, licenceRef: multiUseLicenceRef }),
      await ChargeVersionHelper.add({ billingAccountId, licenceRef: multiUseLicenceRef }),
      await ChargeVersionHelper.add({ licenceRef: generateLicenceRef() })
    ]
  })

  after(async () => {
    for (const chargeVersion of chargeVersions) {
      await chargeVersion.$query().delete()
    }
  })

  describe('when there are charge versions that match the billing account id', () => {
    it('returns an array of unique licence references', async () => {
      const result = await FetchImpactedLicencesDal.go(billingAccountId)

      expect(result).to.equal([multiUseLicenceRef, singleUseLicenceRef])
    })
  })

  describe('when there are no charge versions that match the billing account id', () => {
    it('returns an empty array', async () => {
      const result = await FetchImpactedLicencesDal.go(generateUUID())

      expect(result).to.equal([])
    })
  })
})
```

I haven't been able to run it as I'm in the middle of something else. But hopefully this works.

Cruikshanks (changes requested):
Apologies, this has more to do with stuff I missed the first time around.

Cruikshanks (approved):

PR #3280, opened 2026-04-27
Add sending notifications for renewal invitations

jonathangoulding:

This change creates and sends renewal invitations. This follows the existing patterns in place from the manual notice services. 

The notify templates ids have been added. This behaves different to out existing logic. We have two possible templates to used based on the recipients message type, both the email and letter will send either a 'single licence' or 'multiple licences' template, when the recipients has multiple licence refs we have a template which will list all of the licence refs, and for a single it's just the one licence ref. 

This is important to highlight as the personalisation differed depending on the template to use.


test/services/jobs/renewal-invitations/send-renewal-invitations.service.test.js
```diff
@@ -103,5 +158,19 @@ describe('Jobs - Renewal Invitations - Send Renewal Invitations service', () =>
         })
       })
     })
+
+    describe('the "renewalDate"', () => {
+      describe('when the the day is ', () => {
+        describe('"2026-04-15"', () => {
```
Cruikshanks: There's a typo ('the' is duplicated). But it is more that this reads like you were intending to add other scenarios. If not, I think we can just go with


```suggestion
      describe('when the date is "2026-04-15"', () => {
```

test/services/notices/setup/create-notifications.service.test.js
```diff
@@ -412,4 +412,85 @@ describe('Notices - Setup - Create Notifications service', () => {
...
+        name: 'Renewals: invitation',
+        noticeType: 'renewalInvitations',
+        notificationType: 'Renewals invitation',
+        referenceCode,
+        renewalDate: new Date('2021-11-03'),
+        subType: 'renewalInvitation'
+      }
+    })
+
+    it('returns the created the returns notifications', async () => {
```
Cruikshanks:
```suggestion
    it('returns the created returns notifications', async () => {
```

Cruikshanks (changes requested):
Just NITs

Cruikshanks (approved):

PR #3285, opened 2026-04-29
Update create notice for renewal invitations

jonathangoulding:

This change updates the create notice presenter to handle renewal invitations.

A renewal invitation does not require additional metadata (such as options - excluded licences)


test/services/notices/setup/create-notice.service.test.js
```diff
@@ -38,7 +38,8 @@ describe('Notices - Setup - Create Notice service', () => {
         summer: 'true'
       },
       subType: 'returnInvitation',
-      name: 'Returns: invitation'
+      name: 'Returns: invitation',
+      noticeType: 'reminders'
```
Cruikshanks: Based on the rest of the properties, shouldn't this be


```suggestion
      noticeType: 'invitations'
```

Cruikshanks (changes requested):
I might have it wrong 🤷

Cruikshanks (approved):

PR #3302, opened 2026-05-05
Add implementation for string sort

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-5564

Our SonarCloud report has been marking us down as we don't have implementations for our string sorts. 

This PR adds a compareStrings function that uses String.localeCompare in our string sort functions which should make SonarCloud happy.


app/presenters/licence-versions/view.presenter.js
```diff
@@ -187,11 +188,8 @@ function _removeDuplicatePointsById(points) {
   })
 }
 
-function _sortPointsByDescription(a, b) {
-  return a.description.localeCompare(b.description, 'en', {
-    sensitivity: 'base',
-    ignorePunctuation: true
-  })
+function _sortPointsByDescription(first, second) {
```
Cruikshanks, line 191: We've lost the args `sensitivity: 'base'` and `ignorePunctuation: true` which appear to have affected the results in the unit tests.

It now appears to assert that `ABSTRACTION POINT 'C'` should come before `ABSTRACTION POINT A`, which is not what users would expect.

Should we update the function to accept an optional `options` arg, that we default to `{}` and then spread as args to `localeCompare()`?

Or leave this call alone (I don't SonarQube would have flagged this instance)?

Unless you think the original implementation was wrong and it _should_ be this way.

🤷

Cruikshanks (changes requested):
Just got a concern about what looks to be a change in behaviour after implementing the change.

Cruikshanks (approved):

PR #3308, opened 2026-05-06
Create internal user - Select permissions for the user

Jozzey:
https://eaflood.atlassian.net/browse/WATER-5603

This PR will focus on the part of the journey that allows the user to select the permissions.


app/services/users/internal/setup/view-permissions.service.js
```diff
@@ -0,0 +1,44 @@
...
+    isSuper
+  }
+}
+
+async function _isSuper(auth) {
+  const currentUser = await FetchUserDetailsDal.go(auth.credentials.user.id)
+  const isSuper = currentUser.$permissions().key === 'super'
+
+  return isSuper
+}
```
Cruikshanks: Another 'Alanism' is to avoid prefixing variable and method names with `is` and `has`: you can just use the noun.

However, I know `super` is a protected word, so we cannot do that. But I also think we can rename the variable to be more expressive of its purpose.


```suggestion
  const showSuperPermission = await _showSuperPermission(auth)

  const pageData = PermissionsPresenter.go(session)

  return {
    ...pageData,
    showSuperPermission
  }
}

async function _showSuperPermission(auth) {
  const currentUser = await FetchUserDetailsDal.go(auth.credentials.user.id)
  
  return currentUser.$permissions().key === 'super'
}
```

app/views/users/internal/setup/permissions.njk
```diff
@@ -0,0 +1,100 @@
...
+        },
+        {
+          checked: permissions === 'super',
+          hint: {
+            text: "Full access to all areas of the service."
+          },
+          id: "super",
+          text: "Super user",
+          value: "super"
+        } if isSuper === true,
```
Cruikshanks:
```suggestion
        } if showSuperPermission === true,
```

Cruikshanks (changes requested):
Just a NIT

Cruikshanks (approved):

PR #3314, opened 2026-05-07
  Refactor recipient queries to share fragments

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5559

  The licence holder and primary user SQL logic is essentially the same across the renewal invitations and abstraction alerts recipient queries. The only difference is the limiting conditions appended per use case (e.g. filtering by expiring licences or
  registered licences).

  To avoid duplication, the common SQL is extracted into shared DAL files:
  - `licence-holder-recipient-query.dal.js`
  - `primary-user-recipient-query.dal.js`

  Both `generate-renewal-recipients-query.service.js` and the new `generate-abstraction-alert-recipients-query.service.js` now compose their full queries by interpolating these shared fragments and appending their own `WHERE`/`JOIN` conditions.

  As part of this, the query-generation logic that was embedded in `fetch-abstraction-alert-recipients.service.js` is extracted into its own dedicated service, matching the pattern already established for renewal invitations.


jonathangoulding:
Depends on - https://github.com/DEFRA/water-abstraction-system/pull/3310

test/services/notices/setup/abstraction-alerts/generate-abstraction-alert-recipients-query.service.test.js
```diff
@@ -0,0 +1,257 @@
...
+
+// Test framework dependencies
+const Lab = require('@hapi/lab')
+const Code = require('@hapi/code')
+
+const { describe, it, before, beforeEach, after } = (exports.lab = Lab.script())
+const { expect } = Code
+
+// Test helpers
+const LicenceDocumentHeaderSeeder = require('../../../../support/seeders/licence-document-header.seeder.js')
```
jonathangoulding, line 11: We should be able to remove this seeder.

Cruikshanks (approved):

PR #3333, opened 2026-05-13
Add renewal invitation support to send alternate notice service

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5632

When a renewal invitation email fails to deliver to a primary user, the system now creates and sends an alternate letter to the licence holder as a fallback — matching the existing behaviour for returns invitations.

Changes include:

- `SendNoticeService` now triggers alternate notice handling for `renewalInvitation` subtypes, alongside the existing `returnInvitation`
- `SendAlternateNoticeService` routes to the correct handler based on `notice.subtype` using the `NoticeTypes` static lookup
- `CreateAlternateNoticeService` is extended with a unified `go()` that handles both returns and renewal paths, removing the need for a separate `CreateAlternateRenewalNoticeService`
- `CreateNoticePresenter` now stores `expiryDate` and `renewalDate` in the renewal invitation notice metadata so the alternate notice service can access them
- New `FetchFailedRenewalInvitationsService` fetches failed primary user email notifications for a given renewal notice
- New `FetchAlternateRenewalRecipientsService` fetches licence holder contacts for the licence refs from failed notifications


app/services/notices/setup/create-alternate-notice.service.js
Cruikshanks, line 1: I'm going to disagree with the AI on this one.

> `CreateAlternateNoticeService` is extended with a unified go() that handles both returns and renewal paths, removing the need for a separate `CreateAlternateRenewalNoticeService`

IMO, there is a need, because the service is doing the same thing, but differently for the two notice types.

This is highlighted by the fact that we have methods named after the notice type they deal with.

- `_renewalInvitation()`
- `_renewalNotifications()`
- `_returnsInvitation()`
- `_returnsNotifications()`

What if we were to add a third notice type? To me, this is a 'smell' that the service has split responsibilities.

I think we _do_ need `CreateAlternateRenewalNoticeService`, and that the existing service should be renamed `CreateAlternateReturnsNoticeService`.

I accept that will lead to some duplication when their `_notice()` methods are compared, but not enough to warrant trying to handle both notice types in a single service.

Cruikshanks (changes requested):
I disagree with 'Alan©'!

Cruikshanks (approved):

PR #3336, opened 2026-05-14
Fix issue changing units during return submission

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-5634

An issue has been spotted with return submissions, specifically when submitting abstraction volumes

For example, a user submits a return using abstraction volumes and opting for cubic metres as the unit of measure (UOM).

They meant to submit the return as gallons.

The mistake is spotted, so we 'Edit the return'. Cubic metres is the standard UOM we record all volumes in. So we see what they entered in the 'Cubic metres' column.

|Month   |Cubic metres|
|--------|------------|
|November|245,000     |
|December|275,000     |
|January |120,000     |

We then change the UOM to gallons and come back to the `/check` page.

**What happens**

When the UOM was changed to no longer be cubic metres, we added an extra column titled 'Gallons' next to the existing 'Cubic metres' column.

It shows the result of converting the quantity from cubic metres to gallons.

|Month   |Gallons          |Cubic metres|
|--------|-----------------|------------|
|November|53,892,465.833255|245,000     |
|December|60,491,543.282225|275,000     |
|January |26,396,309.79588 |120,000     |

The conversion is correct. But this is not what was expected.

**What should happen**

The 'line' in the session records the quantity entered separately from the UOM. So, when the UOM changed, the quantity should have remained unchanged. It should now be treated as gallons and converted to cubic metres.

This is what should be shown to the user and recorded in the return when saved.

|Month   |Gallons|Cubic metres|
|--------|-------|------------|
|November|245,000|1,113.79205 |
|December|275,000|1,250.17475 |
|January |120,000|545.5308    |

**The issue**

When the submission was abstraction volumes `ApplyQuantitiesService` was converting what was stored in `line.quantityCubicMetres` to the UOM selected and storing that in `line.quantity`.

It needs to be the other way around, which is how `_readingsUpdate()` does it; hence, submissions by readings are not affected by this bug.


app/services/return-logs/setup/apply-quantities.service.js
Jozzey: Could do with amending this comment to reflect the change in how volumes are updated

Jozzey (changes requested):

Jozzey (approved):

PR #3339, opened 2026-05-15
Add Jira skill for fetching WATER project ticket context

jonathangoulding:
## Summary

- Adds a new Jira skill under `.agents/skills/tools/jira/` with scripts to fetch read-only ticket context from the WATER Jira project
- Adds a setup section to `SKILL.md` explaining the three required env vars (`JIRA_USER`, `JIRA_TOKEN`, `JIRA_BASE_URL`) and how to add them to `.claude/settings.json`
- Registers the skill in `AGENTS.md` with an invocation example

## Test plan

- [x] Set `JIRA_USER`, `JIRA_TOKEN`, and `JIRA_BASE_URL` in `.claude/settings.json`
- [x] Run `bash .agents/skills/tools/jira/auth.sh` and confirm it prints `OK - Authenticated as <name>`
- [x] Run `bash .agents/skills/tools/jira/ticket.sh 4904` and confirm ticket details are printed
- [x] Verify read-only constraint: confirm the skill refuses requests to create or update tickets

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.agents/skills/tools/jira/SKILL.md
+++ b/.agents/skills/tools/jira/SKILL.md
@@ -11,7 +11,7 @@ The scripts require three environment variables:
 
 - **`JIRA_USER`** — the email address you use to log in to Jira
 - **`JIRA_TOKEN`** — a personal API token ([create one here](https://id.atlassian.com/manage-profile/security/api-tokens))
-- **`JIRA_BASE_URL`** — your organisation's Jira base URL, e.g. `https://eaflood.atlassian.net` (no trailing slash)
+- **`JIRA_BASE_URL`** — the organisation's Jira base URL, e.g. `https://eaflood.atlassian.net` (no trailing slash)
 
 Export them in your shell before starting your agent session:
 
--- a/.agents/skills/tools/jira/SKILL.md
+++ b/.agents/skills/tools/jira/SKILL.md
@@ -21,7 +21,7 @@ export JIRA_TOKEN="your-api-token"
 export JIRA_BASE_URL="https://your-org.atlassian.net"
 ```
 
-Or add them to a `.env` file and source it (`source .env`). Most AI agent tools also support configuring environment variables in their settings — consult your agent's documentation for the preferred approach.
+Or add them to a `.env` file and source it (`source .env`).
 
 Run `bash .agents/skills/tools/jira/auth.sh` to verify the credentials are working before using the skill.
 
--- a/.agents/skills/tools/jira/SKILL.md
+++ b/.agents/skills/tools/jira/SKILL.md
@@ -1,0 +1,57 @@
+---
+name: jira
+description: Read-only access to WATER project Jira tickets for task context
+---
+
+# Jira skill
+
+## Setup
+
+The scripts require three environment variables. Add them to `.claude/settings.json` in the project root (create the file if it does not exist):
+
+```json
+{
+  "env": {
+    "JIRA_USER": "your-email@example.com",
+    "JIRA_TOKEN": "your-api-token",
+    "JIRA_BASE_URL": "https://your-org.atlassian.net"
+  }
+}
+```
+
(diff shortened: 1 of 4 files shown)


.agents/skills/tools/jira/SKILL.md
```diff
@@ -0,0 +1,57 @@
+---
+name: jira
+description: Read-only access to WATER project Jira tickets for task context
+---
+
+# Jira skill
+
+## Setup
+
+The scripts require three environment variables. Add them to `.claude/settings.json` in the project root (create the file if it does not exist):
```
Cruikshanks: What if I'm not using claude? 🤷 

Is there a way this can be picked up from env vars? Do I need to give it a try locally with a different agent first, to confirm this is in keeping with our 'generic agent' ethos?

Cruikshanks (changes requested):
This appears to be Claude-specific. Any ideas if it is possible to make this work with, at least, Copilot as well?

.agents/skills/tools/jira/SKILL.md
jonathangoulding: Good spot will look into it.

.agents/skills/tools/jira/SKILL.md
```diff
@@ -0,0 +1,57 @@
...
+
+# Jira skill
+
+## Setup
+
+The scripts require three environment variables:
+
+- **`JIRA_USER`** — the email address you use to log in to Jira
+- **`JIRA_TOKEN`** — a personal API token ([create one here](https://id.atlassian.com/manage-profile/security/api-tokens))
+- **`JIRA_BASE_URL`** — your organisation's Jira base URL, e.g. `https://eaflood.atlassian.net` (no trailing slash)
```
Cruikshanks:
```suggestion
- **`JIRA_BASE_URL`** — the organisation's Jira base URL, e.g. `https://eaflood.atlassian.net` (no trailing slash)
```

.agents/skills/tools/jira/SKILL.md
```diff
@@ -0,0 +1,57 @@
...
+
+Export them in your shell before starting your agent session:
+
+```bash
+export JIRA_USER="your-email@example.com"
+export JIRA_TOKEN="your-api-token"
+export JIRA_BASE_URL="https://your-org.atlassian.net"
+```
+
+Or add them to a `.env` file and source it (`source .env`).
```
Cruikshanks: IMHO, this is content specifically for me, the dev and not for the AGENT. I think we should try to ensure what goes in these files is _only_ for the agent to avoid any confusion (though whether that's important, I have no clue!)

So, this would be something we'd put in a new `Agents` section of the README, along with some waffle about how we're currently using agents, that we opt for open standards where they exist, and we avoid locking ourselves into a specific agent.

I'm happy to have a crack at doing that part (in a follow up change), if you agree this was intended for the 'humie' rather than the agent?

Cruikshanks (changes requested):
A request for change, only if I have understood the intent correctly!

.agents/skills/tools/jira/SKILL.md
jonathangoulding: Yes for the human more than the agent as the script to authenticate has guard clauses

Cruikshanks: Then we should drop this section from the change. I've made a note; it won't get lost!

Cruikshanks (changes requested):
As we've agreed that section is for us 'humies', we should drop it from the change before merging.

.agents/skills/tools/jira/SKILL.md
jonathangoulding: removed

Cruikshanks (approved):

PR #3349, opened 2026-05-19
Create internal user - Check page - Scaffolding

Jozzey:
https://eaflood.atlassian.net/browse/WATER-5604

This PR will focus on scaffolding the files for the Check page part of the journey.


app/presenters/users/internal/setup/check.presenter.js
```diff
@@ -0,0 +1,27 @@
+'use strict'
+
+/**
+ * Formats data for the '' page
```
jonathangoulding: Missing the route here, needs to be added. 

I think we normally do this as part of this change.

jonathangoulding (approved):

PR #3351, opened 2026-05-19
Create internal user - Check page

Jozzey:
https://eaflood.atlassian.net/browse/WATER-5604

This PR will focus on creating the content for the Check page following the scaffolding done in PR https://github.com/DEFRA/water-abstraction-system/pull/3349


test/dal/users/internal/create-user.dal.test.js
```diff
@@ -0,0 +1,127 @@
...
+      await user.$query().delete()
+    }
+  })
+
+  describe('when the permission has no groups or roles', () => {
+    beforeEach(() => {
+      session = { email: generateUserName(), permissions: 'basic' }
+    })
+
+    it('returns a reset GUID', async () => {
```
jonathangoulding: Why do this ?

Jozzey: I've updated the test so that now rather than just checking that a GUID is returned, it checks that the correct users `resetGuid` is returned

app/dal/users/internal/create-user.dal.js
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Creates a new user
+ *
+ * @param {object} session - The session instance
+ *
+ * @returns {Promise<string>} resetGuid - The reset GUID for the created user which is used in the email sent to the
+ * user to complete their account set up
+ */
+async function go(session) {
+  const { application, groups, roles } = userPermissions[session.permissions]
```
jonathangoulding: destructure the session here like. 

const {email, permissions} = session

app/presenters/users/internal/setup/email.presenter.js
```diff
@@ -16,8 +16,9 @@ const { formatEmail } = require('../../../base.presenter.js')
  */
 function go(session) {
   return {
+    activeNavBar: 'users',
     backLink: {
-      href: '/system/users',
+      href: _checkUrl(session),
```
jonathangoulding: i don't think we prefix like this ? 

this would be _href or the _backLink would be in the function.

app/services/users/index-users.service.js
```diff
@@ -30,9 +31,12 @@ async function go(yar, auth, page) {
 
   const pageData = IndexUsersPresenter.go(users, auth)
 
+  const notification = readFlashNotification(yar)
+
   return {
```
jonathangoulding, line 36: order of the object. 

either spread at the start or end - the alanism alpha check should have picked this up. ?

test/dal/users/internal/create-user.dal.test.js
```diff
@@ -0,0 +1,129 @@
...
+
+      expect(result).to.equal(resetGuid)
+    })
+
+    it('creates the user with the correct attributes', async () => {
+      await CreateUserDal.go(session)
+
+      const result = await UserModel.query().where('username', session.email).limit(1).first()
+
+      expect(result.application).to.equal('water_admin')
```
jonathangoulding: Why are all attributes not tested? 

what if the default of a column is updated by a migration.

test/dal/users/internal/create-user.dal.test.js
```diff
@@ -0,0 +1,129 @@
...
+    })
+
+    it('creates the user groups', async () => {
+      await CreateUserDal.go(session)
+
+      const user = await UserModel.query().where('username', session.email).limit(1).first()
+      const result = await UserGroupModel.query().where({ userId: user.userId })
+
+      expect(result).to.have.length(1)
```
jonathangoulding: Is there no other value to test than the length ?

test/services/users/internal/setup/submit-check.service.test.js
```diff
@@ -12,69 +12,79 @@ const { expect } = Code
...
   let session
   let sessionData
+  let yarStub
 
   beforeEach(() => {
-    payload = { placeholder: 'change me' }
-    sessionData = {}
+    sessionData = {
+      email: 'bob.bobbles@environment-agency.gov.uk',
+      permissions: 'billing_and_data'
```
jonathangoulding: this would be permission not permissions. If there is only one permission.

test/dal/users/internal/create-user.dal.test.js
Jozzey: Yes, I can do a better test than that. I got the AI to write the tests for the dal and missed that this one was a bit weak. I'll tidy up some of the other tests in this file as well, I can see a couple of other things that could be done better.

app/presenters/users/internal/setup/email.presenter.js
Jozzey: That's because originally I think I was using the function `checkUrl` which didn't work properly with the '/system/users' route so I had to write a private function for it. I'll rename it to `_href`.

app/services/users/index-users.service.js
Jozzey: The alanism seem to didn't pick it up. I'll move it to the end.

test/dal/users/internal/create-user.dal.test.js
Jozzey: Added additional attributes

app/dal/users/internal/create-user.dal.js
```diff
@@ -0,0 +1,77 @@
...
+ * @param {object} session - The session instance
+ *
+ * @returns {Promise<string>} resetGuid - The reset GUID for the created user which is used in the email sent to the
+ * user to complete their account set up
+ */
+async function go(session) {
+  const { email, permission } = session
+  const { application, groups, roles } = userPermissions[permission]
+
+  // We do this because a "Basic access" user can be both internal and external, so we set it to be internal here
```
jonathangoulding, line 27: I don't think we comment anywhere else. Than helpers / seeders. 

This would be lifted into a private function with the comment as a js doc.

app/dal/users/internal/create-user.dal.js
```diff
@@ -0,0 +1,77 @@
...
+const UserModel = require('../../../models/user.model.js')
+const { generateUUID, timestampForPostgres } = require('../../../lib/general.lib.js')
+const { userPermissions } = require('../../../lib/static-lookups.lib.js')
+
+/**
+ * Creates a new user
+ *
+ * @param {object} session - The session instance
+ *
+ * @returns {Promise<string>} resetGuid - The reset GUID for the created user which is used in the email sent to the
```
jonathangoulding: I think this needs to be changed to 

@returns {Promise<string>} resetGuid - The reset GUID for the created user

This function is only concerned with what it creates and returns and not what the calling function does with it.

app/presenters/users/internal/setup/check.presenter.js
```diff
@@ -14,11 +16,15 @@
  */
 function go(session) {
   return {
-    backLink: {
-      href: '',
-      text: 'Back'
+    activeNavBar: 'users',
+    email: session.email,
```
jonathangoulding: you can destructure the session here. 

Optional resolve if you disagree. 

But this pattern/ approach keeps the session close to the top and shows clearly whats used from the session.

app/presenters/users/internal/setup/email.presenter.js
```diff
@@ -26,6 +27,14 @@ function go(session) {
   }
 }
 
+function _href(session) {
```
jonathangoulding: same as the previous comment. 

Passing in the 'things' as args show clearly what the functions want / need, keeps the functions 'lean' and follows the Principle of Least Knowledge.

app/validators/users/internal/setup/permissions.validator.js
jonathangoulding, line 4: Does this need to become permission ? 

Can catch up about it.

app/views/users/internal/setup/permissions.njk
```diff
@@ -21,7 +21,7 @@
           value: "basic"
```
jonathangoulding, line 21: Not relevant to this PR.

But these values surely exist in the static lib  ?

test/dal/users/internal/create-user.dal.test.js
```diff
@@ -0,0 +1,132 @@
...
+
+    if (user) {
+      await UserGroupModel.query().delete().where({ userId: user.userId })
+      await UserRoleModel.query().delete().where({ userId: user.userId })
+      await user.$query().delete()
+    }
+  })
+
+  describe('when the permission has no groups or roles', () => {
+    beforeEach(() => {
```
jonathangoulding: This should be top level 

and have the nested describes add the permission like: 

session.permission =  'basic'

Reduces duplication and reads better from top down to see the bfe sets up the data for the afe.

test/dal/users/internal/create-user.dal.test.js
```diff
@@ -0,0 +1,132 @@
...
+      session = { email: generateUserName(), permission: 'basic' }
+    })
+
+    it('creates the user with the correct attributes', async () => {
+      await CreateUserDal.go(session)
+
+      const user = await UserModel.query().where('username', session.email).limit(1).first()
+
+      expect(user.application).to.equal('water_admin')
```
jonathangoulding, line 42: Why doesn't this check the whole object response ? 

The discussion before was about what happens if a migration adds a new field, how would this function handle it. 

In the rare case we don't set a default and it blows up as we have seen with legacy code.

app/dal/users/internal/create-user.dal.js
Jozzey: As discussed. We do occasionally add comments like this to explain why we are doing something if it's not immediately obvious.

A couple of examples of this can be found in `app/services/bill-runs/setup/submit-region.service.js`, `app/presenters/licence-monitoring-station/setup/full-condition.presenter.js` and `app/services/bill-runs/setup/determine-blocking-supplementary.service.js`

app/dal/users/internal/create-user.dal.js
Jozzey: Writing it as `@returns {Promise<string>}` would appear to be our standard, see `app/services/bill-runs/check-busy-bill-runs.service.js` & `app/services/return-logs/create-return-logs.service.js` as examples.

Agreed about the actual comment, I've amended it to `@returns {Promise<string>} The reset GUID for the created user` which is more inline with other similar JSDocs.

app/validators/users/internal/setup/permissions.validator.js
Jozzey: Following a discussion with @jonathangoulding about this. A lot of the confusion around what to call this comes from what the page is called "Select permissions for the user", which is why the page is called permissions rather than permission. As it's only possible to select a single "permission" it may be better to call the page "Select **permission** for the user" or words to that effect. The selected permission is then also referred to on the check page as a "permissions", see screenshot:

Screenshot 2026-05-27 at 12 20 43

We agreed to leave as is for this PR and discuss with @Cruikshanks when he is back.

app/views/users/internal/setup/permissions.njk
Jozzey: Originally the radios were built from the values in the static lib but they do not contain all the required data  and not all of them are used. It was @Cruikshanks that advised me to hardcode the radio options rather than try and do something clever with the static lib.

test/dal/users/internal/create-user.dal.test.js
Jozzey: Further to our chat this will not be actioned as it's not really practical to test the output as a whole. For items like the password, resetGuid and resetGuidCreatedAt. I'm not going to know what the values are but I do need to check that they are populated, which isn't possible to do when checking the output as a whole. Currently all fields which are relevant to the creation of a user are tested so any breaking change to the data would be picked up.

jonathangoulding (approved):

PR #3352, opened 2026-05-19
Add renewal invitations to the adhoc notice journey

jonathangoulding:
This change adds the renewal invitations to the adhoc notice journey; the notice type is added as an option to select.

This logic to check for a valid licence will be updated in a later change.

### DIFF
--- a/app/presenters/notices/setup/notice-type.presenter.js
+++ b/app/presenters/notices/setup/notice-type.presenter.js
@@ -71,19 +71,19 @@ function _options(noticeType, journey, auth) {
   }
 
   if (journey === NoticeJourney.ADHOC) {
-    options = [
-      ...options,
-      {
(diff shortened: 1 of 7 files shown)


app/presenters/notices/setup/notice-type.presenter.js
```diff
@@ -77,6 +77,11 @@ function _options(noticeType, journey, auth) {
         checked: noticeType === NoticeType.PAPER_RETURN,
         value: NoticeType.PAPER_RETURN,
         text: 'Paper return'
+      },
+      {
```
jonathangoulding: Can anyone do renewal invitations ?

Cruikshanks: Responded to you on Slack

Cruikshanks (approved):

PR #3358, opened 2026-05-20
Refactor notice setup submit licence

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5646

Extract licence validation for the returns notice types into a dedicated SubmitReturnsLicenceService splitting the Joi schema into a reusable licenceRefSchema helper and a LicenceDueReturnsValidator that extends it.

Move the licence existence check into its own DAL under licences/ and have the session save driven by additionalSessionData returned from the type-specific service rather than mutated directly.


app/dal/licences/fetch-licence-exists.dal.js
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Checks whether a licence exists for the given licence reference.
+ *
+ * @param {string} licenceRef - The licence reference to check
+ *
+ * @returns {Promise<boolean>} Whether a matching licence exists in the database
+ */
+async function go(licenceRef) {
+  const licence = await LicenceModel.query().where('licenceRef', licenceRef).select('licenceRef').first()
```
Cruikshanks: With Objection, if you only want one result, we should also be explicit in the SQL used. `first()` is just a helper method that returns the first result from the result set.



Adding the `limit(1)` can help PostgreSQL optimise the query.
```suggestion
  const licence = await LicenceModel.query().where('licenceRef', licenceRef).select('licenceRef').limit(1).first()
```

app/services/notices/setup/returns-notice/submit-returns-licence.service.js
```diff
@@ -0,0 +1,61 @@
+'use strict'
+
+/**
+ * Orchestrates the returns notice types for the `/notices/setup/{sessionId}/licence` page
```
Cruikshanks:
```suggestion
 * Orchestrates validating returns notice types for the `/notices/setup/{sessionId}/licence` page
```

app/services/notices/setup/returns-notice/submit-returns-licence.service.js
```diff
@@ -0,0 +1,61 @@
...
+ * @module SubmitReturnsLicenceService
+ */
+
+const FetchDueReturnsForLicenceService = require('../returns-notice/fetch-due-returns-for-licence.service.js')
+const FetchLicenceExistsDal = require('../../../../dal/licences/fetch-licence-exists.dal.js')
+const LicenceDueReturnsValidator = require('../../../../validators/notices/setup/licence-due-returns.validator.js')
+const { formatValidationResult } = require('../../../../presenters/base.presenter.js')
+
+/**
+ * Orchestrates the returns notice types for the `/notices/setup/{sessionId}/licence` page
```
Cruikshanks:
```suggestion
 * Orchestrates validating the returns notice types for the `/notices/setup/{sessionId}/licence` page
```

app/dal/notices/setup/check-licence-exists.dal.js
Cruikshanks, line 1: This should go into a folder that mirrors where it's used, not what it's fetching: `app/dal/notices/setup/returns-notice/fetch-licence-exists.dal.js`.

When we reuse it for renewal invitations, it would then move to `app/dal/notices/setup/fetch-licence-exists.dal.js`, so we can go straight to putting it there if you prefer.

But I would also change the module's name. 'Fetch' normally implies we're fetching one or more instances of a record, even if just a few fields of it.

Where the query is being used to figure something else out, for example, the next reference number or whether something exists, we use a different verb.

As we already have `app/dal/users/check-email-exists.dal.js`, I would suggest renaming it `app/dal/notices/setup/check-licence-exists.dal.js`.

test/dal/licences/fetch-licence-exists.dal.test.js
```diff
@@ -0,0 +1,42 @@
...
+const { describe, it, after, before } = (exports.lab = Lab.script())
+const { expect } = Code
+
+// Test helpers
+const LicenceHelper = require('../../support/helpers/licence.helper.js')
+
+// Thing under test
+const FetchLicenceExistsDal = require('../../../app/dal/licences/fetch-licence-exists.dal.js')
+
+describe('DAL - Fetch licence exists dal', () => {
```
Cruikshanks: > I'm aware having checked, a _lot_ of the existing DAL test files are using this pattern, rather than what we do elsewhere. I've missed that on previous PR's. 🤦

This doesn't match our normal pattern. As it is, I would expect

- `Licences - Fetch Licence Exists DAL`

But based on my previous comment, it would be

- `Notices - Setup - Check Licence Exists DAL`

app/validators/notices/setup/licence.validator.js
Cruikshanks, line 1: I'm probably being slow, but I can't see anything currently using this validator. Do we still need it?

Cruikshanks (changes requested):
Initial pass.

Most of these changes are due to our lack of consistency with the DAL. Sorry AI!

Cruikshanks (approved):

PR #3362, opened 2026-05-21
Add licence renewal expiry validation

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5646

When validating a renewal notice we need to check the expired, lapsed and revoked dates. We also need to check the expired date is equal to or greater than 90 days in the future.

This changes updates the 'ProcessRenewalsNoticeLicenceSubmission' to get the licence (if the licence ref is valid) and validate the renewal licence dates.


app/validators/notices/setup/renewal-notice/licence-renewal.validator.js
```diff
@@ -7,25 +7,89 @@
...
+
+function _licenceHasExpiryDate(value, helpers, licenceRenewal) {
+  if (!licenceRenewal || licenceRenewal.expiredDate) {
+    return value
+  }
+
+  return helpers.error('no-expiry-date', { licenceRef: value })
+}
+
+function _licenceExpiryDateInRange(value, helpers, licenceRenewal) {
```
jonathangoulding, line 72: I am still not clear on this one.

Cruikshanks (approved):

app/services/notices/setup/submit-licence.service.js
```diff
@@ -78,14 +79,15 @@ async function _save(session, payload, additionalSessionData) {
...
-  if (noticeType === NoticeType.PAPER_RETURN && !checkPageVisited) {
-    return {
-      redirectUrl: 'paper-return'
-    }
+function _redirect(noticeType, journey, checkPageVisited, licenceChanged) {
+  if (noticeType === NoticeType.PAPER_RETURN) {
+    if (!checkPageVisited || licenceChanged)
+      return {
+        redirectUrl: 'paper-return'
+      }
```
Cruikshanks: SonarCloud is right on this one.


```suggestion
    if (!checkPageVisited || licenceChanged) {
      return {
        redirectUrl: 'paper-return'
      }
    }
```

Cruikshanks (changes requested):
Missed this

Cruikshanks (approved):

PR #3371, opened 2026-05-26
Add common yar stub and replace inline definitions

jonathangoulding:
We have stared to implement some common stubs, we recently added a session and global notifier stub helper to make stubbing consistent and simpler. 

This change follows the previous established pattern to add a stub for 'yar' which is primarily used for notifications. 

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>


app/services/bill-runs/submit-index-bill-runs.service.js
```diff
@@ -88,7 +88,7 @@ function _save(payload, yar) {
 }
 
 function _savedFilters(payload) {
-  const { clear, get, set, ...billRunsFilter } = payload
+  const { clear, flash, get, set, touch, ...billRunsFilter } = payload
```
jonathangoulding: Just need to call out this is done in 5 places. 

I don't fully understand why this is done in the first place.

Jozzey: What it's doing is "give me everything on the yar object except its built-in session methods, and call whatever's left billRunsFilter".

To add some additional confusion here there is a typo. `payload` should be `yar`.

Jozzey: A better way to write that function may be:
```
function _savedFilters(yar) {
  return {
    number: null,
    openFilter: true,
    regions: [],
    runTypes: [],
    statuses: [],
    yearCreated: null,
    ...yar.get('billRunsFilter')
  }
}
```

Jozzey (commented):
I've added some comments. May be worth re-writing those `_savedFilters` functions so they work a bit more like the one in `app/services/notices/submit-view-notice.service.js`, up to you.

Cruikshanks (approved):

PR #3380, opened 2026-05-29
Create internal user - Send verification email

Jozzey:
https://eaflood.atlassian.net/browse/WATER-5658

This PR will focus on the user verification process. In particular, the sending of the email to the user following confirmation of the user details on the check page.


app/dal/users/internal/create-user.dal.js
Cruikshanks: I missed this on the last PR, but I think it is worth sorting now.

This function/DAL is only for creating internal users. So, we should never have to resolve what the value of `application` should be: it will always be `'water_admin'` so we can just hard-code it in the insert query.

app/dal/users/internal/create-user.dal.js
```diff
@@ -41,10 +44,29 @@ async function go(session) {
...
+  const eventData = {
+    createdAt: timestampForPostgres(),
+    entities: [],
+    issuer: username,
+    licences: [],
+    metadata: { user: email, userId },
+    subtype: 'internal',
+    type: 'new-user',
+    updatedAt: timestampForPostgres()
+  }
```
Cruikshanks: When we are responsible for setting the `createdAt` and `updatedAt` values, I like to first instantiate the value so that both are synced.

That way, it is crystal-clear that the record has not been updated since it was created.


```suggestion
  const { username } = await FetchUserDetailsDal.go(auth.credentials.user.id)
  
  const timestamp = timestampForPostgres()

  const eventData = {
    createdAt: timestamp,
    entities: [],
    issuer: username,
    licences: [],
    metadata: { user: email, userId },
    subtype: 'internal',
    type: 'new-user',
    updatedAt: timestamp
  }
```

app/dal/users/internal/create-user.dal.js
```diff
@@ -41,10 +44,29 @@ async function go(session) {
       await _insertUserRoles(roleIds, id, trx)
     }
 
+    await _insertEvent(auth, email, trx, userId)
```
Cruikshanks: We have a convention (I think!) that the `trx` is always the last arg.


```suggestion
    await _insertEvent(auth, email, userId, trx)
```

app/dal/users/internal/create-user.dal.js
```diff
@@ -41,10 +44,29 @@ async function go(session) {
...
+
     return resetGuid
   })
 }
 
+async function _insertEvent(auth, email, trx, userId) {
```
Cruikshanks:
```suggestion
async function _insertEvent(auth, email, userId, trx) {
```

app/dal/users/internal/create-user.dal.js
```diff
@@ -41,10 +44,29 @@ async function go(session) {
...
+  const { username } = await FetchUserDetailsDal.go(auth.credentials.user.id)
```
Cruikshanks: `FetchUseDetailsDal` was written to provide data for the view internal user page.

As such, it is returning _a lot_ of data about the user, all of which is redundant in this scenario. All we want is the `username`.

Fortunately, there already exists a much lighter DAL module which will do this for you. Can I suggest switching to `app/dal/users/fetch-user.dal.js`?

app/dal/users/internal/update-notification.dal.js
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @param {object} notification - The notification model instance to update
+ * @param {object} sendResult - The result from sending the notification
+ *
+ * @returns {Promise<object>} The updated notification record
+ */
+async function go(notification, sendResult) {
+  const { notifyError, notifyId, notifyStatus, plaintext, status } = sendResult
+
+  return notification.$query().patch({ notifyError, notifyId, notifyStatus, plaintext, status })
```
Cruikshanks: Whenever we update, ideally, we should also update the `updatedAt` field.

> You'll need to require `lib/general.lib.js`


```suggestion
  const { notifyError, notifyId, notifyStatus, plaintext, status } = sendResult
  
  const updatedAt = timestampForPostgres()

  return notification.$query().patch({ notifyError, notifyId, notifyStatus, plaintext, status, updatedAt })
```

app/dal/users/internal/create-verification-notification.dal.js
Cruikshanks, line 1: This looks to be specifically creating the verification notification.

So, rather than assume this is a generic DAL for creating notifications, I would suggest renaming it to make this clear: `create-verification-notification.dal.js`.

test/services/users/internal/setup/submit-check.service.test.js
```diff
@@ -12,29 +12,48 @@ const { expect } = Code
...
+  }
+  const resetGuid = '8f5123e3-a982-40c1-843c-b6ef09b1c6bd'
+
+  let auth
   let session
   let sessionData
   let yarStub
 
   beforeEach(() => {
+    auth = { credentials: { user: { id: 1 } } }
```
Cruikshanks: I could be wrong, but I believe `id` is a GUID, and `userId` is a number. So, I'd suggest change this to a GUID to more closely reflect what the service being tested would see.

test/services/users/internal/setup/submit-check.service.test.js
```diff
@@ -44,48 +63,40 @@ describe('Users - Internal - Setup - Submit Check Service', () => {
...
-          text: 'There was a problem adding the user. Please try again.'
-        })
+      expect(flashType).to.equal('notification')
+      expect(notificationData).to.equal({
+        titleText: 'User added',
+        text: `We have emailed ${session.email} instructions to complete their account set up.`
       })
     })
+
+    it('calls the SendVerificationEmailService with the notification', async () => {
```
Cruikshanks:
```suggestion
    it('calls the SendVerificationEmailService with the new notification', async () => {
```

Cruikshanks (changes requested):
First pass feedback. Should be in first thing Monday if you need to ping me about anything.

Cruikshanks (approved):

PR #3391, opened 2026-06-03
Move additional contacts recipients into recipient queries

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5633

Extract the additional contact recipient SQL into a shared query fragment in recipient-queries.dal.js alongside the existing licence holder and primary user queries, and add test coverage for the new query and recipient scenarios.

We are moving from generate functions to using query strings, this makes the logic simpler and easier to test. Where we have 'scenarios'  setting up the recipients data, we no longer need to pass in partial query's to support testing.


test/dal/notices/recipient-queries.dal.test.js
```diff
@@ -0,0 +1,216 @@
...
+        new Date('2023-01-01')
+      )
+    })
+
+    after(async () => {
+      await RecipientScenariosSeeder.clean(scenarios)
+    })
+
+    describe('when called', () => {
+      it('additionalContactRecipientQuery is a valid SQL string', () => {
```
Cruikshanks:
```suggestion
      it('returns the expected query', () => {
```

test/dal/notices/recipient-queries.dal.test.js
```diff
@@ -0,0 +1,216 @@
...
+      // 1) Licence holder present
+      scenarios.withLicenceHolder = await RecipientScenariosSeeder.licenceHolderOnly()
+    })
+
+    after(async () => {
+      await RecipientScenariosSeeder.clean(scenarios)
+    })
+
+    describe('when called', () => {
+      it('licenceHolderRecipientQuery is a valid SQL string', () => {
```
Cruikshanks:
```suggestion
      it('returns the expected query', () => {
```

test/dal/notices/recipient-queries.dal.test.js
```diff
@@ -0,0 +1,216 @@
...
+      // 2) No primary user present. The contact should NOT appear in results.
+      scenarios.noPrimaryUser = await RecipientScenariosSeeder.licenceHolderOnly()
+    })
+
+    after(async () => {
+      await RecipientScenariosSeeder.clean(scenarios)
+    })
+
+    describe('when called', () => {
+      it('primaryUserRecipientQuery is a valid SQL string', () => {
```
Cruikshanks:
```suggestion
      it('returns the expected query', () => {
```

Cruikshanks (changes requested):
Just NITs, and the PR title needs to change to reflect that this is now a refactor ahead of adding support for specific licence references.

Cruikshanks (approved):

PR #3410, opened 2026-06-08
Initiate company contacts session with licences

jonathangoulding:

We need to seed the licences for the company when we intiate the session so we can display (and use) the licences in the company contact journey.

This change adds logic to fetch licences for a company and populates the company contact session with the licences.

---------

Co-authored-by: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/app/dal/company-contacts/fetch-company-licences.dal.js
+++ b/app/dal/company-contacts/fetch-company-licences.dal.js
@@ -1,0 +1,30 @@
+'use strict'
+
+/**
+ * Fetch all licences linked to the given company via their licence version holders
+ * @module FetchCompanyLicencesDal
+ */
+
+const LicenceModel = require('../../models/licence.model.js')
+
+/**
+ * Fetch all licences linked to the given company via their licence version holders
+ *
+ * @param {string} companyId - The UUID of the company to fetch licences for
+ *
+ * @returns {Promise<object[]>} An array of licence objects with `id` and `licenceRef`, sorted by `licenceRef`
+ */
+async function go(companyId) {
+  return LicenceModel.query()
+    .select(['licences.id', 'licences.licenceRef'])
+    .whereExists(
+      LicenceModel.relatedQuery('licenceVersions')
+        .innerJoinRelated('licenceVersionHolder')
+        .where('licenceVersionHolder.companyId', companyId)
+    )
+    .orderBy([{ column: 'licenceRef', order: 'asc' }])
+}
+
+module.exports = {
+  go
+}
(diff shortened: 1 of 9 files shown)


app/dal/company-contacts/fetch-company-licences.dal.js
```diff
@@ -0,0 +1,43 @@
...
+  const query = `
+  SELECT
+    l.id,
+    l.licence_ref as "licenceRef"
+  FROM licences l
+  ${currentLicenceVersionsJoin}
+  WHERE
+    (l.expired_date IS NULL OR l.expired_date >= ?)
+    AND (l.lapsed_date IS NULL OR l.lapsed_date >= ?)
+    AND (l.revoked_date IS NULL OR l.revoked_date >= ?)
```
Cruikshanks: The example I gave, licences that 'end' on the current date are still eligible, hence the `>=`.

But in general, if today is 2026-06-09, and the licence ends on the same date, it is 'ended' and no longer live.

```suggestion
    (l.expired_date IS NULL OR l.expired_date > ?)
    AND (l.lapsed_date IS NULL OR l.lapsed_date > ?)
    AND (l.revoked_date IS NULL OR l.revoked_date > ?)
```

test/dal/company-contacts/fetch-company-licences.dal.test.js
Cruikshanks, line 1: ## Dates

Rather than using fixed dates, have you thought about using the test helpers `test/support/general.js`? We have `yesterday()` and `tomorrow()`, and combined with `app/lib/general.lib.js` `today()` would mean the dates we're using are clear and dynamic. They also test close to the condition in the WHERE clause.

## Scenarios

You came up with a great way for a scenario to know how to clean itself, for example, those in `test/support/seeders/crm-contacts.seeder.js`.

Wouldn't this benefit from the same approach?

Then, taking that a step further, are `_scenario()` and `_scenarioWithMultipleLicences()` doing enough to warrant moving them to a `test/support/seeders/` module of their own?

Cruikshanks (changes requested):
Sorry, the query needs another tweak.

Also, I have some thoughts on the tests for it. Feel free to push back if you disagree.

test/dal/company-contacts/fetch-company-licences.dal.test.js
jonathangoulding: Updated the dates to use yesterday() and tomorrow()

Cruikshanks (approved):

PR #3412, opened 2026-06-09
Create edit internal user

Jozzey:
https://eaflood.atlassian.net/browse/WATER-5514

Following on from the work we did to create an internal user. We are now going to implement a mechanism to edit an internal user, which initially will just allow editing their permission level.


Jozzey:
A seperate PR is going to be raised in relation to this, which will hide the "Edit" button on the "User details" page for an internal user, when the authenticated user is the same as the user being viewed.

This will prevent a user from editing their own permissions.

app/dal/users/internal/update-user.dal.js
```diff
@@ -0,0 +1,113 @@
...
+ * @returns {Promise<string>} The resetGuid for the updated user
+ */
+async function go(auth, session) {
+  const { email, permission, user } = session
+  const { groups, roles } = userPermissions[permission]
+
+  const { currentPermission, id, userId, username } = user
+
+  const groupIds = await GroupModel.query().select('id').whereIn('group', groups)
+  const roleIds = await RoleModel.query().select('id').whereIn('role', roles)
```
Cruikshanks: There is no point fetching these if we are not going to use them.

Perhaps add a private function that returns an object with the IDs if the permission has changed, e.g. `{ groupIds: [], roleIds: [] }`, else NULL.

Then check if this object is null or not inside the transaction, to determine if `await _insertUserGroupsRoles()` should be called.

Alternatively, the function always returns an object with those properties, but they are empty arrays if the permission has not been changed.

Then we split up `_deleteExistingGroupsRoles()` into `_deleteExistingGroups()` and `_deleteExistingRoles()`, and then take advantage of the logic already in `_insertUserGroupsRoles()` as to whether to process the groups and roles.

test/dal/users/internal/update-user.dal.test.js
```diff
@@ -0,0 +1,199 @@
...
+
+        const user = await UserModel.query().where('username', existingUser.username).limit(1).first()
+
+        expect(user.username).to.equal(existingUser.username)
+        expect(result).to.be.undefined()
+      })
+    })
+
+    describe('and the permission has changed', () => {
+      describe('and the new permission has no groups or roles', () => {
```
Cruikshanks, line 140: Unfortunately, the test states it is checking what happens when the permission has changed, but in this case it hasn't.

You might be better off setting up a new user with a permission and then testing

- what happens when changed to basic
- what happens when changed to another permission
- what happens when not changed

I think this is a case of _not_ defaulting to a user with permissions is making the tests harder, not easier.

Cruikshanks (changes requested):
I think just one more tweak to `UpdateUserDal` and then we've got it 😁

app/dal/users/internal/update-user.dal.js
Jozzey: I've updated it so that it works similar to your first suggestion. I don't think the other way would work as if the permissions had changed to a basic user the `groupIds` & `roleIds` arrays would legitimately both be empty.

test/dal/users/internal/update-user.dal.test.js
Jozzey: That will teach me not to try and save a bit of time by getting the AI to write tests. Seem to have spent longer fixing the AIs tests than I would have just writing them myself 🤣

app/dal/users/internal/update-user.dal.js
```diff
@@ -0,0 +1,128 @@
...
+    const [groupIds, roleIds] = await Promise.all([
+      GroupModel.query().select('id').whereIn('group', groups),
+      RoleModel.query().select('id').whereIn('role', roles)
+    ])
+
+    return { groupIds, roleIds }
+  }
+
+  return null
+}
```
Cruikshanks, line 111: I'd flip this. The idea of a guard clause is to bail early if the method is not applicable.


```suggestion
async function _newGroupsRoles(currentPermission, permission) {
  if (permission === currentPermission) {
    return null
  }

  const { groups, roles } = userPermissions[permission]

  const [groupIds, roleIds] = await Promise.all([
    GroupModel.query().select('id').whereIn('group', groups),
    RoleModel.query().select('id').whereIn('role', roles)
  ])

  return { groupIds, roleIds }
}
```

test/dal/users/internal/update-user.dal.test.js
```diff
@@ -0,0 +1,212 @@
...
+  describe('when called', () => {
+    describe('and the email has changed', () => {
+      it('updates the user with the correct attributes', async () => {
+        await UpdateUserDal.go(auth, session)
+
+        const user = await UserModel.query().where('username', session.email).limit(1).first()
+
+        expect(user.resetGuid).to.exist()
+        expect(user.username).to.equal(session.email)
+      })
```
Cruikshanks: Since we're doing a thorough check of resetGuid in the test below, I would skip it here and make the description more specific. But I would also ensure we have _updated_ the existing record rather than create a new one.


```suggestion
      it('updates the username', async () => {
        await UpdateUserDal.go(auth, session)

        const user = await UserModel.query().findById(existingUser.id)

        expect(user.username).to.equal(session.email)
      })
```

test/dal/users/internal/update-user.dal.test.js
```diff
@@ -0,0 +1,212 @@
...
+        )
+
+        expect(event.createdAt).to.be.instanceof(Date)
+        expect(event.updatedAt).to.be.instanceof(Date)
+      })
+
+      it('replaces the existing resetGuid with a new one and returns the updated users resetGuid', async () => {
+        const result = await UpdateUserDal.go(auth, session)
+
+        const { resetGuid } = await UserModel.query().where('username', session.email).limit(1).first()
```
Cruikshanks: Again, we should ensure we are fetching our existing user record.


```suggestion
        const { resetGuid } = await UserModel.query().findById(existingUser.id)
```

test/dal/users/internal/update-user.dal.test.js
```diff
@@ -0,0 +1,212 @@
...
+
+    describe('and the email has not changed', () => {
+      beforeEach(() => {
+        session.email = existingUser.username
+      })
+
+      it('does not update the username and returns undefined', async () => {
+        const result = await UpdateUserDal.go(auth, session)
+
+        const user = await UserModel.query().where('username', existingUser.username).limit(1).first()
```
Cruikshanks:
```suggestion
        const user = await UserModel.query().findById(existingUser.id)
```

Cruikshanks (changes requested):
Just NITs now!

Cruikshanks (approved):

PR #3417, opened 2026-06-10
Add licences to company contact journey

jonathangoulding:

 When a contact is set to receive alerts for 'some' licences, the journey now branches from the abstraction alerts page to a licences page.

  - Add licences checkbox list to the licences page, with items pre-checked from any existing session data
  - Fix redirect in submit-abstraction-alerts: 'some' routes to /licences, 'yes'/'no' route to /check
  - Save selected licences to session.abstractionAlertLicences in submit-licences
  - Pass abstractionAlertLicences through submit-check as a JSON string when 'some', or null otherwise, for both create and update paths
  - Persist abstractionAlertLicences in create-company-contact and update-company-contact services

### DIFF
--- a/test/services/company-contacts/setup/submit-licences.service.test.js
+++ b/test/services/company-contacts/setup/submit-licences.service.test.js
@@ -30,8 +30,8 @@ describe('Company Contacts - Setup - Licences Service', () => {
 
   beforeEach(() => {
     licence = {
-      id: generateLicenceRef(),
-      licenceRef: generateUUID()
+      id: generateUUID(),
+      licenceRef: generateLicenceRef()
     }
 
     company = CustomersFixtures.company()
--- a/test/services/company-contacts/setup/submit-licences.service.test.js
+++ b/test/services/company-contacts/setup/submit-licences.service.test.js
@@ -9,7 +9,10 @@ const { describe, it, beforeEach, afterEach } = (exports.lab = Lab.script())
 const { expect } = Code
 
 // Test helpers
+const CustomersFixtures = require('../../../support/fixtures/customers.fixture.js')
 const SessionModelStub = require('../../../support/stubs/session.stub.js')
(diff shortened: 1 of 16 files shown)


test/services/company-contacts/setup/submit-licences.service.test.js
```diff
@@ -48,6 +62,24 @@ describe('Company Contacts - Setup - Licences Service', () => {
...
+
+        fetchSessionStub.resolves(session)
+      })
+
+      it('saves the selected licences', async () => {
+        await SubmitLicencesService.go(session.id, payload)
+
+        expect(session.abstractionAlertLicences).to.equal([licence.id])
+      })
+    })
```
Cruikshanks: Unless you strongly feel it is necessary, I don't believe we need this scenario. `handleOneOptionSelected()` is a shared helper method with its own testing specifically to avoid having to test this scenario in every page that uses checkboxes.

test/validators/company-contacts/setup/licences.validator.test.js
```diff
@@ -14,7 +14,7 @@ describe('Company Contacts - Setup - Licences Validator', () => {
   let payload
 
   beforeEach(() => {
-    payload = { placeholder: '' }
+    payload = { licences: ['1'] }
```
Cruikshanks: I believe what we get back is the licence ID. Unless there is a strong reason not to, we should use data that represents what the service actually deals with.


```suggestion
    payload = { licences: ['0111902c-1bea-4530-bcae-4f4dcc90a6cc'] }
```

Cruikshanks (changes requested):
Just a couple of NITs

test/validators/company-contacts/setup/licences.validator.test.js
jonathangoulding: Used the generate uuid func.

Cruikshanks (approved):

PR #3422, opened 2026-06-11
Add abstraction alerts some to contact details page

jonathangoulding:

This adds the logic to show the updated text on the company contacts details page.

This will now show the text for an abstraction alert that is 'no', 'yes' or 'some'.

This also change adds the '$abstractionAlertType' instance method to the company contact model.

This gives us a single place to work out if an abstraction alert is 'no', 'yes' or 'some'.

### DIFF
--- a/app/presenters/company-contacts/contact-details.presenter.js
+++ b/app/presenters/company-contacts/contact-details.presenter.js
@@ -5,6 +5,7 @@
  * @module ContactDetailsPresenter
  */
 
+const { abstractionAlertsLabel } = require('../crm.presenter.js')
 const { formatEmail, formatLongDate } = require('../base.presenter.js')
 
 /**
@@ -38,13 +39,8 @@ function go(company, companyContact) {
 
 function _abstractionAlerts(companyContact) {
   const abstractionAlerts = companyContact.$abstractionAlertType()
-  const abstractionAlertsText = {
-    no: 'No',
-    some: 'Yes, for some licences',
-    yes: 'Yes, for all licences'
-  }
 
-  return abstractionAlertsText[abstractionAlerts]
+  return abstractionAlertsLabel(abstractionAlerts)
 }
 
 function _created(companyContact) {
--- a/app/presenters/company-contacts/contact-details.presenter.js
+++ b/app/presenters/company-contacts/contact-details.presenter.js
@@ -23,7 +23,7 @@ function go(company, companyContact) {
       text: 'Go back to licence holder contacts'
     },
     contact: {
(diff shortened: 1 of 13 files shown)


app/presenters/company-contacts/contact-details.presenter.js
```diff
@@ -36,6 +36,17 @@ function go(company, companyContact) {
   }
 }
 
+function _abstractionAlerts(companyContact) {
+  const abstractionAlerts = companyContact.$abstractionAlertType()
+  const abstractionAlertsText = {
+    no: 'No',
+    some: 'Yes, for some licences',
+    yes: 'Yes, for all licences'
+  }
```
Cruikshanks: I thought the idea was that `$abstractionAlertType()` would return an object, for example, `{ label: 'Yes, for some licences', value: 'some' }`, so you wouldn't need to do this anymore?

test/models/company-contact.model.test.js
```diff
@@ -158,4 +159,41 @@ describe('Company Contacts model', () => {
...
+          })
+        })
+
+        it('returns "some"', () => {
+          const result = testRecord.$abstractionAlertType()
+
+          expect(result).to.equal('some')
+        })
+      })
+    })
```
Cruikshanks, line 199:
```suggestion
    describe('when "abstractionAlerts" is enabled', () => {
      before(async () => {
        testRecord = await CompanyContactHelper.add({ abstractionAlerts: true })
      })
      
      describe('and "abstractionAlertLicences" is null', () => {
        it('returns "yes"', () => {
          const result = testRecord.$abstractionAlertType()

          expect(result).to.equal('yes')
        })
      })

      describe('and "abstractionAlertLicences" is populated', () => {
        before(async () => {
          testRecord = await CompanyContactHelper.add({
            abstractionAlertLicences: JSON.stringify([generateUUID()]),
            abstractionAlerts: true
          })
        })

        it('returns "some"', () => {
          const result = testRecord.$abstractionAlertType()

          expect(result).to.equal('some')
        })
      })
    })
```

Cruikshanks (changes requested):
Just a clarification and a NIT.

app/presenters/company-contacts/contact-details.presenter.js
jonathangoulding: We need them for the create where this would not be set.

jonathangoulding: So they have to exist for the create. 

We just need to know what type it is.

Cruikshanks (approved):

PR #3427, opened 2026-06-12
Add licences to company contact check page

jonathangoulding:

This change adds the licences to the company contact check page.

The licences need to be checked against the abstraction alert licences, if a licence has ended since being added to the abstraction alert licences, it will not show, this is implicit based on the licences.

We have introduced a generic macro to handle how we show the details.


app/presenters/company-contacts/setup/check.presenter.js
```diff
@@ -57,6 +58,18 @@ function _emailInUse(sentNotification, companyContact) {
...
+  if (abstractionAlerts !== 'some') {
+    return []
+  }
+
+  return licences.reduce((acc, licence) => {
+    if (abstractionAlertLicences.includes(licence.id)) {
+      acc.push(licence.licenceRef)
+    }
+    return acc
+  }, [])
```
Cruikshanks: This is only because `reduce()` scares me.

Would this work, and would you consider it 'dumber'? 😁 


```suggestion
  return licences.filter((licence) => {
    return abstractionAlertLicences.includes(licence.id)
  })
```

test/presenters/company-contacts/setup/check.presenter.test.js
```diff
@@ -115,6 +128,70 @@ describe('Company Contacts - Setup - Check Presenter', () => {
...
+      })
+
+      describe('when the abstraction alert is not "some"', () => {
+        it('returns an empty array', () => {
+          const result = CheckPresenter.go(session, savedCompanyContacts, sentNotification)
+
+          expect(result.licences).to.be.empty()
+        })
+      })
+    })
```
Cruikshanks, line 190: I think I would structure and word the tests differently. Apologies if I have broken a test (I've not been able to run this)


```suggestion
    describe('the "licences" property', () => {
      describe('when the user is set to receive abstraction alerts for all licences ("yes")', () => {
        it('returns an empty array', () => {
          const result = CheckPresenter.go(session, savedCompanyContacts, sentNotification)

          expect(result.licences).to.be.empty()
        })
      })

      describe('when the user is set not to receive abstraction alerts ("no")', () => {
        it('returns an empty array', () => {
          const result = CheckPresenter.go(session, savedCompanyContacts, sentNotification)

          expect(result.licences).to.be.empty()
        })
      })

      describe('when the user is set to receive abstraction alerts for specific licences ("some")', () => {
        beforeEach(() => {
          session.abstractionAlerts = 'some'
        })

        describe('and the licence holder has "live" licences', () => {
          describe('and the selected licences for the user matches the "live" licences', () => {
            beforeEach(() => {
              session.abstractionAlertLicences = [licences[0].id]
            })

            it('returns the matching licence references', () => {
              const result = CheckPresenter.go(session, savedCompanyContacts, sentNotification)

              expect(result.licences).to.equal([licences[0].licenceRef])
            })
          })

          describe('but none of the selected licences for the user matches the "live" licences', () => {
            it('returns an empty array', () => {
              const result = CheckPresenter.go(session, savedCompanyContacts, sentNotification)

              expect(result.licences).to.be.empty()
            })
          })
        })

        describe('but the licence holder has no "live" licences', () => {
          beforeEach(() => {
            session.licences = []
            session.abstractionAlertLicences = [licences[0].id]
          })

          describe('even though the user has selected licences to receive abstraction alerts for', () => {
            it('returns an empty array', () => {
              const result = CheckPresenter.go(session, savedCompanyContacts, sentNotification)

              expect(result.licences).to.be.empty()
            })
          })
        })
      })
    })
```

Cruikshanks (changes requested):

app/presenters/company-contacts/setup/check.presenter.js
jonathangoulding: I originally did a filter then map. 

We need the licence refs for the view.

Cruikshanks (approved):

PR #3434, opened 2026-06-15
Add warning for ended licences company details page

jonathangoulding:

When a licence has ended, if could still be 'saved' in the abstraction alert licences. When this is the case we want to show a warnign to the user.

This change adds a warning to the user when any licence from the abstraction alert licences has ended.

This change also introduces a '$ended' instance method for the lenience model.

### DIFF
--- a/test/models/licence.model.test.js
+++ b/test/models/licence.model.test.js
@@ -45,7 +45,7 @@ const UserModel = require('../../app/models/user.model.js')
 const WorkflowHelper = require('../support/helpers/workflow.helper.js')
 const WorkflowModel = require('../../app/models/workflow.model.js')
 const { generateUUID, today } = require('../../app/lib/general.lib.js')
-const { tomorrow } = require('../support/general.js')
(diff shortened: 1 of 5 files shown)


test/models/licence.model.test.js
```diff
@@ -670,6 +670,109 @@ describe('Licence model', () => {
...
+
+          endedRecord = LicenceModel.fromJson({ expiredDate, lapsedDate, revokedDate })
+        })
+
+        it('returns true', () => {
+          const result = endedRecord.$ended()
+
+          expect(result).to.be.true()
+        })
+      })
```
Cruikshanks: This needs to be in the section above because right now it reads as

- when a licence has not ended
  - because an end date is today
    - returns true

This could lead someone to think licences that end 'today' are not ended, when in fact the test confirms they are.

Cruikshanks (changes requested):
Just a NIT

Cruikshanks (approved):

PR #3435, opened 2026-06-15
Handle no licences company contact page

jonathangoulding:

It is possible for a company to have no 'active' licences. When this happens, we need to hide the 'show all licences' option so we don't show an empty page for licences.

This change checks if there are licences, if there is we show the 'Show all licences' option.

### DIFF
--- a/app/views/company-contacts/setup/abstraction-alerts.njk
+++ b/app/views/company-contacts/setup/abstraction-alerts.njk
@@ -21,7 +21,7 @@
           value: "some",
           text: "Yes, for some licences",
           checked: abstractionAlerts === 'some'
-        },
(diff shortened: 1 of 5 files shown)


app/views/company-contacts/setup/abstraction-alerts.njk
Cruikshanks, line 1: We include a divider on a page that has radio buttons when there are more than 2 options.

If we're hiding this option, then we also need to hide the divider, so I would suggest adding the same ` if showSomeLicences` to the 'divider' option.

Cruikshanks (changes requested):
Just a NIT

Cruikshanks (approved):

PR #3436, opened 2026-06-15
Handle no licences company contact page edit

jonathangoulding:

It is possible for a company to have no 'active' licences. When this happens, we hide the 'show all licences' option so we don't show an empty page for licences.

This change updates the edit session to default the abstraction alert to null when there are no licences, overwiring the users' previous selection.

### DIFF
--- a/test/services/company-contacts/setup/initiate-edit-session.service.test.js
+++ b/test/services/company-contacts/setup/initiate-edit-session.service.test.js
@@ -120,7 +120,7 @@ describe('Company Contacts - Setup - Initiate edit Session service', () => {
             stubFetchCompanyLicencesDal.resolves([])
           })
 
-          it('returns "some"', async () => {
+          it('returns "no"', async () => {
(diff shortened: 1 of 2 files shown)


test/services/company-contacts/setup/initiate-edit-session.service.test.js
```diff
@@ -85,12 +85,48 @@ describe('Company Contacts - Setup - Initiate edit Session service', () => {
...
+            })
+          })
+        })
+
+        describe('and there are no active licences', () => {
+          beforeEach(() => {
+            stubFetchCompanyLicencesDal.resolves([])
+          })
+
+          it('returns "some"', async () => {
```
Cruikshanks: Oh no, it doesn't!


```suggestion
          it('returns "no"', async () => {
```

Cruikshanks (changes requested):
NIT!

Cruikshanks (approved):

Cruikshanks (approved):

PR #3437, opened 2026-06-15
Update existing checkboxes to use bracket notation

jonathangoulding:
When a checkbox group uses a plain name (e.g. name="regions"), the browser's form submission behaviour is inconsistent:

- Nothing checked  → field absent from payload (undefined)
- One checked      → payload value is a string
- Multiple checked → payload value is an array

We previously handled this in each submit service by calling handleOneOptionSelected(), which normalised all three cases to an array. This commit takes the approach agreed on the dev call: appending [] to every checkbox name attribute (e.g. name="regions[]") so Hapi's body parser always coerces a selection into an array, removing the need for the helper entirely.

With bracket notation the three cases become:

- Nothing checked  → field still absent from payload (undefined)
- One checked      → payload value is a single-element array
- Multiple checked → payload value is an array

The undefined case is handled differently depending on context:

- Filter pages (bill-runs index/review, notices index) have no "required" validator on the checkbox fields, so undefined would flow through and be stored in the session. These services default the field to [] with ??= [] before saving.

- submit-alert-thresholds assigns payload.alertThresholds to the session object before calling the validator, so undefined would corrupt the in-memory session used to render the error page. ??= [] prevents that.

- submit-licences calls _payloadSelection() (which calls licences.includes()) before the validator runs. ??= [] prevents a TypeError on undefined.

- submit-purpose calls _combinePurposeDetails() (which iterates payload.purposes with for...of) before the validator runs. ??= [] prevents a TypeError on undefined.

- All other submit services validate first and only access the array when validation passes, so undefined is caught as a "required" error and no ??= [] is needed.

handleOneOptionSelected() is removed from submit-page.lib.js as it is no longer used anywhere.

Co-authored-by: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/app/lib/submit-page.lib.js
+++ b/app/lib/submit-page.lib.js
@@ -30,52 +30,6 @@ function clearFilters(payload, yar, filterKey) {
 }
 
 /**
- * Normalizes checkbox form data to always be an array
- *
- * HTML checkboxes behave inconsistently when submitted:
- * - No selection: property is undefined
- * - Single selection: property is a string
- * - Multiple selections: property is an array
- *
- * This helper ensures the property always exists as an array, making it easier to process checkbox data consistently
- * throughout the application.
- *
- * @param {object} payload - The form payload object to be modified in place
- * @param {string} key - The property name in the payload to normalize
- *
- * @returns {void} The payload object is modified directly
- *
- * @example
- * // No selection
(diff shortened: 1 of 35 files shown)


jonathangoulding:
Having issues with the "??= []" "fix".

PR #3450, opened 2026-06-17
Migrate disable/enable internal users

Jozzey:
https://eaflood.atlassian.net/browse/WATER-5682

Add functionality to allow a user with appropriate permissions to enable/disable an internal user.  This option should only be available when editing a user.


app/dal/users/internal/fetch-user-details.dal.js
```diff
@@ -16,7 +16,7 @@ const UserModel = require('../../../models/user.model.js')
  */
 async function go(id) {
   return UserModel.query()
-    .select(['id', 'userId', 'username'])
+    .select(['id', 'enabled', 'userId', 'username'])
```
Cruikshanks: We don't need this change. The 'status' modifier already ensures this field is returned.


```suggestion
    .select(['id', 'userId', 'username'])
```

app/dal/users/internal/update-user.dal.js
```diff
@@ -110,14 +110,19 @@ async function _newGroupsRoles(currentPermission, permission) {
...
 
-  if (email !== username) {
+  if (email !== currentUsername) {
     userData.resetGuid = generateUUID()
     userData.username = email
   }
 
+  if (enabled !== currentlyEnabled) {
+    userData.enabled = enabled
+  }
```
Cruikshanks: Do we need to worry about this check?

It won't affect the query's performance. It just means if the user hasn't changed the access for the user being edited, we'd just be setting it to the existing value. 🤷

Cruikshanks (changes requested):
Just a couple of NITs

app/dal/users/internal/update-user.dal.js
Jozzey: If you are happy for it to be removed, I shall remove it.

Cruikshanks (approved):

PR #3451, opened 2026-06-17
Add licence ended modifier

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5683

This change adds an 'ended' modifier to the lcience model, this works alongside the 'ended' instance method to establish if a licence has ended.

### DIFF
--- a/test/models/licence.model.test.js
+++ b/test/models/licence.model.test.js
@@ -682,6 +682,22 @@ describe('Licence model', () => {
       revokedDate = null
     })
 
+    describe('when the "ended" modifier has been used', () => {
+      let testLicence
+
+      before(async () => {
+        testLicence = await LicenceModel.query().findById(testRecord.id).modify('ended')
+      })
+
+      it('returns the dates', () => {
+        expect(testLicence).to.equal({
+          expiredDate: null,
+          lapsedDate: null,
+          revokedDate: null
+        })
+      })
+    })
+
     describe('when a licence has ended', () => {
       describe('because is has expired', () => {
         beforeEach(() => {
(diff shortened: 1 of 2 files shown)


test/models/licence.model.test.js
```diff
@@ -682,6 +682,18 @@ describe('Licence model', () => {
       revokedDate = null
     })
 
+    describe('when the "ended" modifier has been used', () => {
+      it('returns the dates', async () => {
+        const result = await LicenceModel.query().findById(testRecord.id).modify('ended')
```
Cruikshanks: If you don't add a `select()` to your Objection query, it'll do the equivalent of a `SELECT *` in SQL.

So, if you really want to confirm your modifier is working, you need to include something like `select(['id'])`. Without the modifier, only `id` would be populated in the instances returned.

When you add your modifier, then we should expect only `id`, and the end dates to be populated.


```suggestion
        const result = await LicenceModel.query().select(['id']).findById(testRecord.id).modify('ended')
```

Cruikshanks (changes requested):
I think the test needs tweaking.

test/models/licence.model.test.js
```diff
@@ -682,6 +682,18 @@ describe('Licence model', () => {
...
 
+    describe('when the "ended" modifier has been used', () => {
+      it('returns the dates', async () => {
+        const result = await LicenceModel.query().select(['id']).findById(testRecord.id).modify('ended')
+
+        expect(result).to.equal({
+          expiredDate: null,
+          lapsedDate: null,
+          revokedDate: null
+        })
```
jonathangoulding, line 694:
```suggestion
        expect(result).to.equal({
          id: testRecord.id,
          expiredDate: null,
          lapsedDate: null,
          revokedDate: null
        })
```

Cruikshanks (approved):

PR #3454, opened 2026-06-17
Only allow tagging an active licences for a monitoring station

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5684

This change stops the user from tagging a licence that has ended.

We have also lifted out the fetch logic within the service and moved this into the dal, this is part of our ongoing effort to lift database logic out of services into the dal.

### DIFF
--- a/app/dal/licence-monitoring-station/fetch-licence.dal.js
+++ b/app/dal/licence-monitoring-station/fetch-licence.dal.js
@@ -1,0 +1,25 @@
+'use strict'
+
+/**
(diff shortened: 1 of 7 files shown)


Cruikshanks (changes requested):
Thank you for all the [Boy Scout work](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html) in this PR!

Cruikshanks (approved):
Wrong chuffing button!

PR #3462, opened 2026-06-18
Add licence ended warning to the monitoring station licence page

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5689

When a licence has ended, it no longer receives abstraction alerts, however, it may still be tagged. So it will still be displayed on the monitoring station page. 

This change adds a warning to monitoring station view licence page to highlight the licence has ended (and will not receive abstraction alerts). 

We resue the warning from the licence summary page and have lifted this into a shared presenter. 

We have lifted the fetches into Dal as part of our ongoing effort to extract the databse logic from the services.


Cruikshanks (dismissed):

Cruikshanks (approved):
Glad one of the Alan's was paying attention.

PR #3469, opened 2026-06-19
Update company contact cancel page to show 'some' licences

jonathangoulding:

We recently implemented a change to allow the user to select some licences to send abstraction alerts for.

We update the check page to show 'some' licences and to display the chosen licences, as the canel page closely mirros the check page we need to also update the canel page to match.

This change adds the 'some' licence logic to the cancel page.

Some logic has been lifted into reusable code for this and future changes.

### DIFF
--- a/app/presenters/crm.presenter.js
+++ b/app/presenters/crm.presenter.js
@@ -36,24 +36,39 @@ function formatContact(contact, billingQueryArgs) {
 }
 
 /**
- * Returns the licence references the user has selected to receive abstraction alerts for
+ * Compares a user's selected licences to 'live' ones for a company, and returns those that are live
  *
- * This is used to display the licence to the user.
+ * This is used in licence holder contact pages, where we need to show what licence a user has been assigned for a given
+ * notice type.
  *
- * @param {object[]} licences - The licences for the company
- * @param {string[]} abstractionAlertLicences - The IDs of the licences selected to receive abstraction alerts
- * @param {string} abstractionAlerts - The abstraction alerts value ('no', 'some', or 'yes')
+ * For example, licence holder contacts can be set to receive abstraction alerts. If the option has been set as 'yes'
+ * or 'no', then the calling page simply needs to show the setting, either 'Yes, all licences' or 'No'.
  *
- * @returns {string[]} The licence references the user has selected, or an empty array if not applicable
+ * If the option has been set to 'some', then the calling page will show 'Yes, some licences' alongside a list of the
+ * selected licence references.
+ *
+ * The complexity is that over time, licences can expire, lapse, or be revoked. When we show the selected licences, the
+ * intention is to make it clear which ones they will receive a notification for. So, we need to exclude those that are
+ * no longer 'live'.
+ *
+ * This function determines what the list of licence references should be, based on the settings of the user and the
+ * current 'live' licences for the licence holder.
+ *
+ * @param {object[]} liveLicences - The licences for the company
+ * @param {string[]} selectedLicences - The IDs of the selected licences the user should receive notifications for
+ * @param {string} noticeSetting - The notice setting ('no', 'some', or 'yes')
+ *
+ * @returns {string[]} The licence references of 'live' licences the user has selected, or an empty array if not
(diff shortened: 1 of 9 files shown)


app/presenters/crm.presenter.js
```diff
@@ -35,6 +35,31 @@ function formatContact(contact, billingQueryArgs) {
...
+  }
+
+  return licences
+    .filter((licence) => {
+      return abstractionAlertLicences.includes(licence.id)
+    })
+    .map((licence) => {
+      return licence.licenceRef
+    })
+}
```
Cruikshanks, line 76: It's a bit naughty because we are only using this for abstraction alerts. But I'm confident, based on what the UCD team are doing, that we will want the same functionality for a number of notice types.

Therefore I'm suggesting we make this more generic, and as long as we implement 'licence selection' for the other notice types in the same way, we have something we can use for all.

```javascript
/**
 * Compares a user's selected licences to 'live' ones for a company, and returns those that are live
 *
 * This is used in licence holder contact pages, where we need to show what licence a user has been assigned for a given
 * notice type.
 *
 * For example, licence holder contacts can be set to receive abstraction alerts. If the option has been set as 'yes'
 * or 'no', then the calling page simply needs to show the setting, either 'Yes, all licences' or 'No'.
 *
 * If the option has been set to 'some', then the calling page will show 'Yes, some licences' alongside a list of the
 * selected licence references.
 *
 * The complexity is that over time, licences can expire, lapse, or be revoked. When we show the selected licences, the
 * intention is to make it clear which ones they will receive a notification for. So, we need to exclude those that are
 * no longer 'live'.
 *
 * This function determines what the list of licence references should be, based on the settings of the user and the
 * current 'live' licences for the licence holder.
 *
 * @param {object[]} liveLicences - The licences for the company
 * @param {string[]} selectedLicences - The IDs of the selected licences the user should receive notifications for
 * @param {string} noticeSetting - The notice setting ('no', 'some', or 'yes')
 *
 * @returns {string[]} The licence references of 'live' licences the user has selected, or an empty array if not
 * applicable
 */
function selectedLiveLicences(liveLicences, selectedLicences, noticeSetting) {
  if (noticeSetting !== 'some' || !selectedLicences?.length) {
    return []
  }

  return liveLicences
    .filter((licence) => {
      return selectedLicences.includes(licence.id)
    })
    .map((licence) => {
      return licence.licenceRef
    })
}
```

test/support/fixtures/licence.fixture.js
```diff
@@ -25,4 +25,25 @@ function licenceEnds(endDate = null) {
...
+/**
+ * An array of licences with a licence ref
+ *
+ * @returns {LicenceModel[]} An array of licences with licence refs
+ */
+function licenceWithLicenceRefs() {
+  return [licenceWithLicenceRef()]
+}
+
+module.exports = { licenceEnds, licenceWithLicenceRefs }
```
Cruikshanks: I'd argue that whatever this fixture returns should include an ID and a licence reference.

So to start with, I would just have a `licence()` fixture that returns that. I'd then update `licenceEnds()` to call this.

I don't think we then also need a separate fixture that returns an array of this, especially an array of one. That is something we can leave the tests to setup.


```suggestion
/**
 * Generates an instance of `LicenceModel` with an ID and a licence reference
 *
 * You can override or add additional properties by passing them in as an object
 *
 * @param {object} [data={}] - an object of properties to override or add to the licence
 *
 * @returns {LicenceModel} A licence with a licence ref
 */
function licence(data = {}) {
  return LicenceModel.fromJson({
    id: generateUUID(),
    licenceRef: generateLicenceRef(),
    ...data
  })
}

module.exports = { licence, licenceEnds }
```

`licenceEnds()` updated as 

```javascript
function licenceEnds(endDate = null) {
  return licence({
    expiredDate: endDate,
    lapsedDate: null,
    revokedDate: null
  })
}
```

Cruikshanks (changes requested):
First pass. Ping me if anything doesn't make sense or you disagree.

Cruikshanks (approved):

PR #3470, opened 2026-06-19
Update company contact remove page to show 'some' licences

jonathangoulding:

We recently implemented a change to allow the user to select some licences to send abstraction alerts for.

We update the contact details page to show 'some' licences and to display the chosen licences.

This change adds the 'some' licence logic to the remove page.

### DIFF
--- a/app/presenters/company-contacts/remove-company-contact.presenter.js
+++ b/app/presenters/company-contacts/remove-company-contact.presenter.js
@@ -6,30 +6,45 @@
  */
 
 const { formatEmail } = require('../base.presenter.js')
+const { abstractionAlertsLabel } = require('../crm.presenter.js')
 
 /**
  * Formats data for the '/company-contacts/{id}/remove' page
  *
  * @param {module:CompanyModel} company - The customer from the companies table
  * @param {module:CompanyContactModel} companyContact - The customer contact from the company contacts table
+ * @param {object[]} licences - The licences linked to the abstraction alert licences
  *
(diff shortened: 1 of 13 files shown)


app/presenters/company-contacts/remove-company-contact.presenter.js
```diff
@@ -6,32 +6,47 @@
...
 
+function _licences(abstractionAlertType, licences) {
+  if (abstractionAlertType !== 'some') {
+    return []
+  }
+
+  return licences.map((licence) => {
+    return licence.licenceRef
+  })
+}
```
Cruikshanks, line 48: I would have thought we wanted to reuse the fantastic bit of refactoring you suggested in #3469 .

Currently, this looks like it will just show _all_ the licences the user selected, which could make it look very different from the page the user just came from.

Cruikshanks (changes requested):
I think it would be better to wait until #3469 is merged, and we can then assess what we can appropriately reuse.

app/presenters/company-contacts/remove-company-contact.presenter.js
jonathangoulding: This is different logic. 

We want to show all the licences the same as the contact details page.

Cruikshanks: Sorry, that's right. On the contact details page, we show them all but display a warning. 🤦

Cruikshanks (approved):

PR #3476, opened 2026-06-23
Refactor companies services and dal

jonathangoulding:
We are moving service which comminnucate directly with the database into the DAL folder. This creates a seperation between services and external dependencies (in this case a database).

This change refactors the database services in the companies services into the dal folder (and tests).

### DIFF
--- a/test/dal/companies/fetch-licences.dal.test.js
+++ b/test/dal/companies/fetch-licences.dal.test.js
@@ -18,9 +18,9 @@ const { generateRandomInteger, generateUUID } = require('../../../app/lib/genera
 const databaseConfig = require('../../../config/database.config.js')
 
 // Thing under test
-const FetchLicencesService = require('../../../app/services/companies/fetch-licences.service.js')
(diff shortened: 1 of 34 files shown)


test/dal/companies/fetch-licences.dal.test.js
```diff
@@ -18,9 +18,9 @@ const { generateRandomInteger, generateUUID } = require('../../../app/lib/genera
 const databaseConfig = require('../../../config/database.config.js')
 
 // Thing under test
-const FetchLicencesService = require('../../../app/services/companies/fetch-licences.service.js')
+const FetchLicencesDal = require('../../../app/dal/companies/fetch-licences.dal.js')
 
-describe('Companies - Fetch Licences service', () => {
+describe('Companies - Fetch Licences dal', () => {
```
jonathangoulding, line 23: These tests could use the updated seeders. 

I think we can update this to no longer use the licence version holder table ?

In another change.

Cruikshanks (approved):

PR #3478, opened 2026-06-23
Order of communications changes on refresh - Licences

Jozzey:
https://eaflood.atlassian.net/browse/WATER-5674

Our QA team have spotted that when viewing communications, their order can sometimes be different. For example, view the table, select a notification to view, then return to the table, and the order is different.

Users expect the order to be consistent every time they view the page.

We are going to start with some housekeeping. Where we have a `fetch-notifications.service.js` (or something similar), move it into the DAL. For example, app/services/return-logs/fetch-notifications.service.js.

Then update all fetch-notifications.dal.js modules to order by:
- created at desc
- message type asc
- status asc
- id asc (id will always be unique, so as a method of last resort, it will ensure the order is always consistent).

This PR will focus on the Licences Communications `app/services/licences/fetch-communications.service.js`.

### DIFF
--- a/app/dal/licences/fetch-notifications.dal.js
+++ b/app/dal/licences/fetch-notifications.dal.js
@@ -29,7 +29,12 @@ async function _fetch(licenceRef, page) {
   return NotificationModel.query()
     .select(['createdAt', 'id', 'messageType', 'status'])
     .where('licences', '?', licenceRef)
-    .orderBy('createdAt', 'DESC')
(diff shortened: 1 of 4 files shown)


Jozzey:
SonarCloud is complaining about duplicate code as these notification Dals are quite similar. Since they were like this before I started making changes I assume this is fine.

Cruikshanks (approved):

PR #3481, opened 2026-06-23
Remove licence version holder from queries

jonathangoulding:
We no longer require the 'licence version holder' view. The licence version now has a company and address id. We are moving away from the licence version holder (we intend to remove it).

This change removes the use of licence version holder from any remaining queries.

We will remove the model and relations in another change, as we want to show clearly the business logic changes.

### DIFF
--- a/test/dal/companies/fetch-licences.dal.test.js
+++ b/test/dal/companies/fetch-licences.dal.test.js
@@ -41,8 +41,6 @@ describe('Companies - Fetch Licences dal', () => {
 
     companyId = licenceHolder.company.id
 
-    //startDate: new Date('2022-01-01'),
-
     // This is an older licence version linked to a different company. This confirms we are getting the right current
     // licence holder details
     olderLicenceVersion = await LicenceVersionHelper.add({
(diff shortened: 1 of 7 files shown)


test/dal/companies/fetch-licences.dal.test.js
```diff
@@ -23,66 +23,48 @@ const FetchLicencesDal = require('../../../app/dal/companies/fetch-licences.dal.
...
-      companyId,
-      derivedName: 'Kanemitsu Corporation',
-      licenceVersionId: licenceVersion.id
-    })
+
+    licenceHolder = await CRMContactsSeeder.licenceHolder({ licence }, 'Kanemitsu Corporation')
+
+    companyId = licenceHolder.company.id
+
+    //startDate: new Date('2022-01-01'),
```
Cruikshanks: Is this a commented-out line that needs removing?

Cruikshanks (changes requested):
I think, just one NIT

Cruikshanks (approved):

PR #3490, opened 2026-06-29
Update adhoc returns invitation templates and personalisation

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5664

We have updated adhoc returns invitations templates which now include the licence ref.

This change adds the licence ref to the personalisation.

### DIFF
--- a/app/presenters/notices/setup/returns-notice-notifications.presenter.js
+++ b/app/presenters/notices/setup/returns-notice-notifications.presenter.js
@@ -68,7 +68,7 @@ function _email(recipient, noticeId, session) {
     messageType,
     messageRef: MESSAGE_REFS[noticeType][journey],
     personalisation: {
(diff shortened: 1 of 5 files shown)


app/presenters/notices/setup/returns-notice-notifications.presenter.js
```diff
@@ -68,6 +68,7 @@ function _email(recipient, noticeId, session) {
     messageType,
     messageRef: MESSAGE_REFS[noticeType][journey],
     personalisation: {
+      licencenumber: recipient.licence_refs[0],
```
jonathangoulding: I think we need to get this updated in the template to camel case before merging.

Cruikshanks: Done!

Cruikshanks (approved):

PR #3523, opened 2026-07-13
Fix remaining test failures from the Hapi Lab to Vitest migration

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-5710

**Summary**

- Brings the suite from 1007/1052 to 1052/1052 passing test files
- Converts leftover Sinon-era patterns (`getCall`, `calledWith`, `onFirstCall`/`onSecondCall`, `returnsThis`, `throws`, `withArgs`) to their Vitest equivalents
- Adds missing `vi.spyOn` mocks for services that were silently falling through to real (DB-hitting) implementations
- Fixes notifier assertions that only checked 2 of the 3 arguments passed to `omfg`
- Fixes two production bugs left over from the CJS to ESM conversion: `check-all-licence-end-dates.service.js` and `hapi-pino.plugin.js` were calling `.go` on services that are now default function exports, so the calls were silently failing
- Fixes `base.request.test.js` calling `defaultOptions.retry` on the function reference instead of `defaultOptions().retry`, which meant PATCH/POST retry tests were quietly using Got's built-in defaults instead of our configured retry methods
- Rewrites `markdown.filter.test.js`, which tested a `globalThis.GlobalMarked` mechanism the source no longer uses since it now imports `marked` directly
- Deduplicates repeated mock setup in `match-and-allocate.service.test.js`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/plugins/hapi-pino.plugin.js
+++ b/app/plugins/hapi-pino.plugin.js
@@ -35,7 +35,7 @@ export default {
     // We want our logs to focus on the main requests and not become full of 'noise' from requests for /assets or
     // pings from the AWS load balancer to /status. We pass this function to hapi-pino to control what gets filtered
     // https://github.com/pinojs/hapi-pino#optionsignorefunc-options-request--boolean
-    ignoreFunc: HapiPinoIgnoreRequestService.go,
+    ignoreFunc: HapiPinoIgnoreRequestService,
     // Add the request params as pathParams to the response event log. This, along with `logPathParams` and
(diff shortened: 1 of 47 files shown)


Cruikshanks (approved):
Happy to merge in on the basis this gets us to 100% passing tests.

It's done some stuff, like moving `const` declarations for reasons. But I don't care right now as they can resolved on the next lint/format fix pass.

PR #3533, opened 2026-07-15
Use vi.mock over spyOn for default export in index-notices test

jonathangoulding:
## Summary
- Replaced `vi.spyOn(FetchNoticesService, 'default')` with `vi.mock()` on a default-imported `FetchNoticesService`, matching the pattern already used in `view-bill.service.test.js`
- Removed a stray `describe.only` that would have skipped the rest of the suite

## Why vi.mock() over spyOn on the namespace default

**The problem with `vi.spyOn(FetchNoticesService, 'default')`**

When you `import * as FetchNoticesService`, you get the module's namespace object. ESM namespace objects are supposed to be immutable/read-only bindings — `spyOn` only works on it because Vitest's transform layer makes module exports configurable under the hood. That's an implementation detail you're relying on, not a guarantee. It also produces a type error (`Argument type string is not assignable to parameter type Classes<...> | Methods<...>`) — tooling doesn't expect you to spy on a property called `'default'` on a namespace object, because that's not really what `default` is semantically; it's a special binding, not a plain object property.

**Why vi.mock() + default import is the correct shape**

- It replaces the module at the resolution level (hoisted before any imports run), not by mutating a property on an already-resolved namespace. That's how ESM mocking is meant to work.
- The test file's import (`import FetchNoticesService from '.../fetch-notices.service.js'`) now matches how the real code imports it — so the mock and the production import shape are symmetric, rather than the test using `import *` just to get a spyable handle.
- `FetchNoticesService.mockResolvedValue(...)` reads as "this function returns X" — more direct than `vi.spyOn(Namespace, 'default').mockResolvedValue(...)`.
- It's the pattern already established elsewhere in the codebase (`view-bill.service.test.js`), so it's consistent rather than a one-off.

`vi.spyOn` is the right tool when you only want to override one export of a module while leaving the rest real (e.g. spying on a named export among several). For a module whose only export is the default function you're stubbing wholesale, `vi.mock()` is more idiomatic and avoids leaning on the namespace-object mutability quirk.

### DIFF
--- a/test/services/notices/index-notices.service.test.js
+++ b/test/services/notices/index-notices.service.test.js
@@ -3,10 +3,12 @@ import * as NoticesFixture from '../../support/fixtures/notices.fixture.js'
 import YarStub from '../../support/stubs/yar.stub.js'
 
 // Things to stub
-import * as FetchNoticesService from '../../../app/services/notices/fetch-notices.service.js'
+import FetchNoticesService from '../../../app/services/notices/fetch-notices.service.js'
 
 // Thing under test
 import IndexNoticesService from '../../../app/services/notices/index-notices.service.js'
+
+vi.mock('../../../app/services/notices/fetch-notices.service.js')
 
 describe('Notices - Index Notices service', () => {
   let auth
@@ -33,7 +35,7 @@ describe('Notices - Index Notices service', () => {
       const results = NoticesFixture.mapToFetchNoticesResult([NoticesFixture.alertReduce()])
 
       fetchResults = { results, total: 1 }
-      vi.spyOn(FetchNoticesService, 'default').mockResolvedValue(fetchResults)
+      FetchNoticesService.mockResolvedValue(fetchResults)
     })
 
     it('returns page data for the view', async () => {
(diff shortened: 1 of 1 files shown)


test/services/notices/index-notices.service.test.js
```diff
@@ -3,11 +3,13 @@ import * as NoticesFixture from '../../support/fixtures/notices.fixture.js'
 import YarStub from '../../support/stubs/yar.stub.js'
 
 // Things to stub
-import * as FetchNoticesService from '../../../app/services/notices/fetch-notices.service.js'
+import FetchNoticesService from '../../../app/services/notices/fetch-notices.service.js'
 
 // Thing under test
 import IndexNoticesService from '../../../app/services/notices/index-notices.service.js'
 
+vi.mock('../../../app/services/notices/fetch-notices.service.js')
```
jonathangoulding, line 11: If this becomes the new established pattern maybe we tweak it 

```
// Things to mock
import FetchNoticesService from '../../../app/services/notices/fetch-notices.service.js'
vi.mock('../../../app/services/notices/fetch-notices.service.js')
```

